PR #16, opened 2020-01-02
Backend ADR

### LINKED ISSUE #5: Write ADR for choice of backend language/framework
Which primary language do we want to use on the backend, and which API library/framework will we use?

Based on the research done in #7 we plan to make extensive use of the `topojson` library, which runs on Node.js.

So we're fairly certain we'll want to use either JS or TypeScript as our language for the backend, but we're still uncertain about an API library/framework.

Based on what what framework we choose, this ADR may encompass #14 as well, or we may tackle that separately (depending on whether we choose a lightweight framework like Flask or a more all-encompassing one like Django, to name some Python web frameworks I'm familiar with).

Considerations:
 - Does the web framework support TypeScript, either out of the box or through a well-maintained third-party @types library?
 - Does the web framework have support built-in for an ORM & database migrations, or a recommended set of libraries that it can interoperate with?

### LINKED ISSUE #14: Write ADR for choice of backend database migrations / ORM library
This may end up being superseded by #5 if we choose a web framework that includes it's own ORM & migrations system (as in Django), but we are leaning towards Node.js and most Node frameworks I am familiar with are more minimal and require pairing them with a separate DB management library.

maurizi:
## Overview

Combined ADRs for backend framework & DB management.

Closes #5  
Closes #14

### DIFF
--- a/doc/arch/adr-01-backend-framework.md
+++ b/doc/arch/adr-01-backend-framework.md
@@ -14,13 +14,13 @@ In addition to being able to perform TopoJSON operations, the back-end API will 
 
 I evaluated [11 different Node.js web frameworks](https://docs.google.com/spreadsheets/d/1zLixcL1Xt53iYPkvzDHidg48uyKrevO7i7zj6EiBMBA/edit?usp=sharing), which generally fell into 3 types:
 
-- Minimalistic, generally with only routing and minimal HTTP support. Other features are potentially available as middleware.
+- Minimalist, generally with only routing and minimal HTTP support. Other features are potentially available as middleware.
    - Express
    - koa
    - hapi
    - Fastify
    - Next
- - Fully featured, with support for user authorization & registration, an ORM, database migrations, etc. Typically built *using* on of the more minimalistic frameworks under the hood.
+ - Fully featured, with support for user authorization & registration, an ORM, database migrations, etc. Typically built *using* one of the more minimalist frameworks under the hood.
    - Adonis
    - Feathers
    - Nest
@@ -31,11 +31,11 @@ I evaluated [11 different Node.js web frameworks](https://docs.google.com/spread
 
 
 From these frameworks, I immediately identified a few that won't work well for our purposes:
- - MeteorJS is popular and well-documented framework, but it only supports MongoDB for persisting data
+ - MeteorJS is a popular and well-documented framework, but it only supports MongoDB for persisting data
  - Next is primarily focused on server-side rendering of React applications, and does not provide many of the features we would need, nor does it have a large third-party plugin ecosystem to make up for the lack of first-party support.
 
 
-For the more minimalistic frameworks, there are third-party tools we could leverage to make up for missing features.
+For the more minimalist frameworks, there are third-party tools we could leverage to make up for missing features.
 
 User registration:
  - Auth0 can provide this as an external service, and is a service we are familiar with
@@ -47,10 +47,10 @@ Database querying & migrations management
  - For a more low-level approach, `node-postgres` or KnexJS combined with `node-db-migrate`
 
 
-The more "batteries-included" frameworks tend to use the above libraries (in some cases even allowing swapping in one with another), rather than including their own implementations. This means that when choosing between one of these frameworks or a more minimalistic one like Express, what we are really choosing between for the most part is whether to put the pieces together ourselves or whether to use the combination of libraries brought together by a framework like Nest or Feathers.
+The more "batteries-included" frameworks tend to use the above libraries (in some cases even allowing swapping in one with another), rather than including their own implementations. This means that when choosing between one of these frameworks or a more minimalist one like Express, what we are really choosing between for the most part is whether to put the pieces together ourselves or whether to use the combination of libraries brought together by a framework like Nest or Feathers.
 
 
-We could also instead consider using a minimalist framework in a hybrid approach, where we build most of the back-end API in a language/framework we are familiar with, such as Python & Django, and encapsulate the TopoJSON-specific operations in a micro-service. This would allow us to not need to learn a new way of managing database migrations, user authentication or management, etc. but with a cost of greater architectural complexity and a potential performance loss due to communication overheads. For such an approach I think it would make sense to go with Express, as we have familiarity with using it for other microservices on previous projects.
+We could also instead consider using a minimalist framework in a hybrid approach, where we build most of the back-end API in a language/framework we are familiar with, such as Python & Django, and encapsulate the TopoJSON-specific operations in a microservice. This would allow us to not need to learn a new way of managing database migrations, user authentication or management, etc., but with a cost of greater architectural complexity and a potential performance loss due to communication overheads. For such an approach I think it would make sense to go with Express, as we have familiarity with using it for other microservices on previous projects.
 
 
 Decision
--- a/doc/arch/adr-01-backend-framework.md
+++ b/doc/arch/adr-01-backend-framework.md
@@ -10,6 +10,8 @@ The main performance critical operation we know the back-end API will need to su
 
 TopoJSON is written in JavaScript (there are TopoJSON libraries in other languages, but none support the critical merge operations that make TopoJSON useful for our situation), and in order to be able to use it in a performant manner we must create a Node.js service that can keep the TopoJSON data structure in-memory. Due to familiarity and enthusiasm from the team, we would prefer to use TypeScript for this service over plain JavaScript.
 
+As an additional point in favor of Node.js / TypeScript, using TypeScript will enable us to share code with the frontend, which will be particularly helpful as we expect to need to mirror some client-side operations on the backend to support contest submissions validation.
+
 In addition to being able to perform TopoJSON operations, the back-end API will need to be able to provide user authentication (and potentially user registration and management as well) and communicate with a PostgreSQL database to persist data.
 
 I evaluated [11 different Node.js web frameworks](https://docs.google.com/spreadsheets/d/1zLixcL1Xt53iYPkvzDHidg48uyKrevO7i7zj6EiBMBA/edit?usp=sharing), which generally fell into 3 types:
@@ -61,6 +63,8 @@ We will proceed by using TypeScript with Nest as the framework for the DistrictB
 The documentation for Nest was among the best of the frameworks I investigated, and the [purposeful similarity to Angular](https://codesandbox.io/s/jjo90y00xw?from-embed) will hopefully provide an easier learning curve given the familiarity within the development team with Angular.
 
 By using Nest we won't have to choose a user authentication strategy (it provides an integration with Passport) or database management library (it uses TypeORM by default). Because it is built on-top of Express we can still use Express-middleware if necessary, but we may not have to as there are a number of plugins (https://github.com/juliandavidmr/awesome-nestjs#components--libraries). Nest also provides a dependency injection framework, which is an added bonus.
+
+---
 
 Feathers was another major contender, and it provides many of the basics we need. Ultimately the choice between them mainly came down to Nest appearing to have better TypeScript support, and the familiarity of the Angular-esque application structure used by Nest.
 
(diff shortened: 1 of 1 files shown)


doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+I evaluated [11 different Node.js web frameworks](https://docs.google.com/spreadsheets/d/1zLixcL1Xt53iYPkvzDHidg48uyKrevO7i7zj6EiBMBA/edit?usp=sharing), which generally fell into 3 types:
+
+- Minimalistic, generally with only routing and minimal HTTP support. Other features are potentially available as middleware.
+   - Express
+   - koa
+   - hapi
+   - Fastify
+   - Next
+ - Fully featured, with support for user authorization & registration, an ORM, database migrations, etc. Typically built *using* on of the more minimalistic frameworks under the hood.
```
kshepard:
```suggestion
 - Fully featured, with support for user authorization & registration, an ORM, database migrations, etc. Typically built *using* one of the more minimalist frameworks under the hood.
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+   - Feathers
+   - Nest
+   - Meteor
+ - Somewhere in-between, with some extra features built-in, and others only as third-party plugins.
+   - Sails
+   - Loopback
+
+
+From these frameworks, I immediately identified a few that won't work well for our purposes:
+ - MeteorJS is popular and well-documented framework, but it only supports MongoDB for persisting data
```
kshepard:
```suggestion
 - MeteorJS is a popular and well-documented framework, but it only supports MongoDB for persisting data
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+ - Next is primarily focused on server-side rendering of React applications, and does not provide many of the features we would need, nor does it have a large third-party plugin ecosystem to make up for the lack of first-party support.
+
+
+For the more minimalistic frameworks, there are third-party tools we could leverage to make up for missing features.
```
kshepard:
```suggestion
For the more minimalist frameworks, there are third-party tools we could leverage to make up for missing features.
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+ - Auth0 can provide this as an external service, and is a service we are familiar with
+ - [Passport](http://www.passportjs.org/docs/)
+
+Database querying & migrations management
+ - [TypeORM](https://typeorm.io/#/), TypeScript first ORM with support for generating migrations
+ - [Sequelize](https://sequelize.org/) more dynamic Javascript-focused ORM, supports migrations but can't generate them
+ - For a more low-level approach, `node-postgres` or KnexJS combined with `node-db-migrate`
+
+
+The more "batteries-included" frameworks tend to use the above libraries (in some cases even allowing swapping in one with another), rather than including their own implementations. This means that when choosing between one of these frameworks or a more minimalistic one like Express, what we are really choosing between for the most part is whether to put the pieces together ourselves or whether to use the combination of libraries brought together by a framework like Nest or Feathers.
```
kshepard:
```suggestion
The more "batteries-included" frameworks tend to use the above libraries (in some cases even allowing swapping in one with another), rather than including their own implementations. This means that when choosing between one of these frameworks or a more minimalist one like Express, what we are really choosing between for the most part is whether to put the pieces together ourselves or whether to use the combination of libraries brought together by a framework like Nest or Feathers.
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+We could also instead consider using a minimalist framework in a hybrid approach, where we build most of the back-end API in a language/framework we are familiar with, such as Python & Django, and encapsulate the TopoJSON-specific operations in a micro-service. This would allow us to not need to learn a new way of managing database migrations, user authentication or management, etc. but with a cost of greater architectural complexity and a potential performance loss due to communication overheads. For such an approach I think it would make sense to go with Express, as we have familiarity with using it for other microservices on previous projects.
```
kshepard:
```suggestion
We could also instead consider using a minimalist framework in a hybrid approach, where we build most of the back-end API in a language/framework we are familiar with, such as Python & Django, and encapsulate the TopoJSON-specific operations in a microservice. This would allow us to not need to learn a new way of managing database migrations, user authentication or management, etc., but with a cost of greater architectural complexity and a potential performance loss due to communication overheads. For such an approach I think it would make sense to go with Express, as we have familiarity with using it for other microservices on previous projects.
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+TopoJSON is written in JavaScript (there are TopoJSON libraries in other languages, but none support the critical merge operations that make TopoJSON useful for our situation), and in order to be able to use it in a performant manner we must create a Node.js service that can keep the TopoJSON data structure in-memory. Due to familiarity and enthusiasm from the team, we would prefer to use TypeScript for this service over plain JavaScript.
+
+In addition to being able to perform TopoJSON operations, the back-end API will need to be able to provide user authentication (and potentially user registration and management as well) and communicate with a PostgreSQL database to persist data.
+
+I evaluated [11 different Node.js web frameworks](https://docs.google.com/spreadsheets/d/1zLixcL1Xt53iYPkvzDHidg48uyKrevO7i7zj6EiBMBA/edit?usp=sharing), which generally fell into 3 types:
+
+- Minimalistic, generally with only routing and minimal HTTP support. Other features are potentially available as middleware.
```
kshepard:
```suggestion
- Minimalist, generally with only routing and minimal HTTP support. Other features are potentially available as middleware.
```

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
+Backend Framework
+==================
+
+Context
+-------
+
+DistrictBuilder 2 is intended to be a primarily front-end application, with a back-end API to persist data and support operations that cannot be feasibly completed in the browser in a performant manner.
+
+The main performance critical operation we know the back-end API will need to support is the merging of geographic areas to form new districts. Through some [research and experimentation](https://docs.google.com/document/d/1znHcDr1wXQ1vr6cNiI5Y9gl5uwIveVPj659Wpug70Is/edit?usp=sharing), we have decided to rely upon [TopoJSON](https://github.com/topojson/topojson) to support this operation.
```
kshepard: > [research and experimentation]

Just want to point out that this is a private document linked in this public ADR, which may be confusing to people that can't access it. I'm unsure of exactly the best solution to this.

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+TopoJSON is written in JavaScript (there are TopoJSON libraries in other languages, but none support the critical merge operations that make TopoJSON useful for our situation), and in order to be able to use it in a performant manner we must create a Node.js service that can keep the TopoJSON data structure in-memory. Due to familiarity and enthusiasm from the team, we would prefer to use TypeScript for this service over plain JavaScript.
+
+In addition to being able to perform TopoJSON operations, the back-end API will need to be able to provide user authentication (and potentially user registration and management as well) and communicate with a PostgreSQL database to persist data.
+
+I evaluated [11 different Node.js web frameworks](https://docs.google.com/spreadsheets/d/1zLixcL1Xt53iYPkvzDHidg48uyKrevO7i7zj6EiBMBA/edit?usp=sharing), which generally fell into 3 types:
```
kshepard: > [11 different Node.js web frameworks]

Another link to a private doc.

doc/arch/adr-01-backend-framework.md
maurizi: I'll make this & the other document set to viewable (but not editable) to anyone with a link. There isn't any information in them that should remain private.

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+Feathers was another major contender, and it provides many of the basics we need. Ultimately the choice between them mainly came down to Nest appearing to have better TypeScript support, and the familiarity of the Angular-esque application structure used by Nest.
+
+Adonis is seemingly less popular and well-known than either Feathers or Nest, though it did seem to offer a similar feature-set to Nest.
+
+The hybrid/microservice approach was compelling, but the added architectural complexity is significant, and I don't think our back-end needs really require a framework as complex and all-encompassing as Django, Additionally, the ability to write the backend in a statically-typed language like TypeScript is attractive, which we would be mainly giving up with a microservice-based approach.
```
hectcastro: It is unclear what this last paragraph is intending to communicate:

- In the first sentence, it seems like a hybrid/microservice approach is being compared to a monolithic web framework (e.g., Django), but both seem to be getting dinged.
- It is unclear how a microservice-based approach is incompatible with using TypeScript.

doc/arch/adr-01-backend-framework.md
```diff
@@ -0,0 +1,79 @@
...
+
+
+The more "batteries-included" frameworks tend to use the above libraries (in some cases even allowing swapping in one with another), rather than including their own implementations. This means that when choosing between one of these frameworks or a more minimalist one like Express, what we are really choosing between for the most part is whether to put the pieces together ourselves or whether to use the combination of libraries brought together by a framework like Nest or Feathers.
+
+
+We could also instead consider using a minimalist framework in a hybrid approach, where we build most of the back-end API in a language/framework we are familiar with, such as Python & Django, and encapsulate the TopoJSON-specific operations in a microservice. This would allow us to not need to learn a new way of managing database migrations, user authentication or management, etc., but with a cost of greater architectural complexity and a potential performance loss due to communication overheads. For such an approach I think it would make sense to go with Express, as we have familiarity with using it for other microservices on previous projects.
+
+
+Decision
```
hectcastro: I haven't dug deep into Nest enough to be able to assess the tradeoffs well, but was there any consideration given to building an application backend in a language ecosystem we're more comfortable in, and only building the TopoJSON manipulation backend in Node.js (kind of like a microservice)?

And as a slight deviation of the question above, was there consideration for an approach that offloads persistence and user management to a third-party service, leaving only a need for a more minimal TopoJSON specific backend service?

doc/arch/adr-01-backend-framework.md
maurizi: This could probably be rephrased.

> In the first sentence, it seems like a hybrid/microservice approach is being compared to a monolithic web framework (e.g., Django), but both seem to be getting dinged.

The intent is to compare the chosen approach (Nest) with a combination of a framework like Django (which has many features that we likely don't need, like templates, an admin site, etc.) paired with a Node.js microservice.

> It is unclear how a microservice-based approach is incompatible with using TypeScript.

By primarily using Python & Django with only specific services in TypeScript, we'd be omitting static typing in most of the codebase.

Additionally, since we're already using TypeScript on the frontend, by using it on the backend as well we have the opportunity to share code directly between the two, which we'll likely need to do for validation of contest submissions. This was touched slightly in the [TopoJSON R&D document](https://docs.google.com/document/d/1znHcDr1wXQ1vr6cNiI5Y9gl5uwIveVPj659Wpug70Is/edit?usp=sharing) but I neglected to mention that here - I can incorporate that context here too.

doc/arch/adr-01-backend-framework.md
maurizi: > ... was there any consideration given to building an application backend in a language ecosystem we're more comfortable in, and only building the TopoJSON manipulation backend in Node.js (kind of like a microservice)?

Yes, this was definitely considered, see the paragraph above at the end of the Context section as well as the paragraph at the end of this section.

While I think Nest looks like it might be able to handle all of the things we'd actually need Django for and think it will work well for our use case, I think using Django & a Node.js microservice would also serve our needs, and might be preferable to some of the _other_ frameworks I investigated besides Nest.

doc/arch/adr-01-backend-framework.md
hectcastro: Ah, I see. I misread or unintentionally missed the last paragraph of the context section, which helps in understanding this paragraph. Also, looks like this response ends up covering one of the Qs I had attached to the decision section.

To make it easier to take any action against my original comment, I would consider separating (possibly with a header, or a horizontal rule) the first 3 paragraphs of the decision section with the last 3. It seems like the first 3 are specific to what the decision is, while the last 3 talk more about what the decision was not.

PR #17, opened 2020-01-03
Add missing scripts/use_version

maurizi:
## Overview

This script appears to have been mistakenly not committed in #13

Script contents were cribbed from GreenKeeper, but perhaps may need to be tweaked for this project, I'm unsure.

## Testing Instructions

- `scripts/setup`
- `scripts/update`


pcaisse:
This was actually modified in Greenkeeper somewhat recently in this PR and it probably makes sense to do the same here (linking to a resolved conversation doesn't seem to open it up automatically, but see the conversation in `scripts/test` which resulted in getting rid of `scripts/use_version` and including that code as a function in `scripts/update`): https://github.com/azavea/greenkeeper-frontend/pull/27

maurizi:
@pcaisse This is ready for another look

kshepard:
Just noticed that `scripts/use_version` is also being called in `scripts/test`. And based on the Greenkeeper discussion, it sound like the thing to do would be to remove this line from `scripts/test`.

PR #18, opened 2020-01-07
Add district data storage ADR

### LINKED ISSUE #8: Write ADR for how district data will be stored
As a continuation of the research task, we should document our findings to demonstrate our chosen approach.

Based on the [research in #7](https://docs.google.com/document/d/1znHcDr1wXQ1vr6cNiI5Y9gl5uwIveVPj659Wpug70Is/edit?usp=sharing) we think TopoJSON is the best approach for merging geometries of the base geo units (blocks, block groups, counties), and the performance is reasonable enough that we think we can store districts as merely a list(s) of IDs of the base geounits, without needing to store the geometries of the districts themselves (we'd calculate them on-the-fly as needed).

Before we settle on this conclusion though, we want to test a more real-world use case that combines geometries of varying levels (e.g. combine Bucks county w/ some block groups & blocks in neighboring counties) and investigate the performance and ease of use of this test case.

kshepard:
## Overview

ADR for storage of district data

Closes #8

### DIFF
--- a/doc/arch/adr-02-district-data-storage.md
+++ b/doc/arch/adr-02-district-data-storage.md
@@ -1,0 +1,66 @@
+# District Data Storage
+
+## Context
+
+[ADR 01](./adr-01-backend-framework.md) briefly discusses the research and experimentation carried out in order to determine the most efficient methods to perform merging of geographic areas to form new districts. It was determined that using [TopoJSON](https://github.com/topojson/topojson) topologies stored in memory was the leading candidate. This ADR assumes we'll be using TopoJSON and lays out an approach for how we'll go about storing district data given that we'll have access to in-memory topologies to make use of.
+
+One of the open tasks going into this research was to gain experience with TopoJSON's capability of storing multiple geolevels within a single topology, and sharing arc data. Storing the static topology data in such a manner allows us to build districts by mixing and matching geounits in different geolevels, and should lead to large performance gains over building up districts from solely blocks. This is particularly the case in the context of DistrictBuilder where the standard user workflow is to work with larger geolevels while building districts, and to only drop down to the block-level when fine tuning.
+
+In order to test this out, I created a rough script that reads a block-level GeoJSON file and constructs a TopoJSON file with multiple geolevels. The block-level GeoJSON file I tested with was Delaware and only had references to tracts and counties. Note: we will likely be using blockgroups in production rather than tracts, but they are similarly sized, so it shouldn't have a noticeable impact on performance measurements. The script is here: [mk-topojson-hierarchy.js](https://gist.github.com/kshepard/8eaca19d06fa88a25279997a3c4160d7#file-mk-topojson-hierarchy-js).
+
+Another script ([benchmark.js](https://gist.github.com/kshepard/8eaca19d06fa88a25279997a3c4160d7#file-benchmark-js)) was produced to consume this TopoJSON output and randomly create districts in several ways:
+ * Using blocks-only to build districts
+ * Using tracts-only to build districts
+ * Using a combination of blocks and tracts to build districts
+
+After running the script multiple times on my X1 Carbon (4th Gen) laptop, here are some results for constructing 10 districts:
+ * Blocks-only: 1500 - 2500 ms
+ * Tracts-only: 100 - 200 ms
+ * Combination (the number of blocks was gradually increased from none to all):
+
+|Blocks| ms |
+|------|----|
+|     0| 157|
+|  3191| 384|
+|  5982| 464|
+|  9053| 610|
+| 11931| 756|
+| 14811|1023|
+| 17166|1032|
+| 18815|1174|
+| 20513|1359|
+| 21915|1654|
+| 23272|1901|
+| 24115|2013|
+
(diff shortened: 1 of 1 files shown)


kshepard:
Thanks for the great questions! I tried to answer them all, but let me know if I missed anything, or if you have follow-ups. Also happy to jump on a call if you want a deeper-dive.

>     * Do the performance numbers benchmarked here involve generating this representation? I didn't see this in the gists, but I might have not realized it was there. If they don't, I'd be interested to see more info specifically about the performance of the serialization and deserialization of this format for those synthetic districts.

No, you're right, this structure was formulated after the creation of the GISTs (which were primarily created for demonstrating the ability and performance of merging elements across multiple geolevels). It would take quite a bit more work to be able to demonstrate using this structure, since it would rely on having static data lookups, which will be worked out in a future [ADR](https://github.com/PublicMapping/district-builder-2/issues/9). I agree that benchmarking the performance of actually using this structure would be a great thing to investigate.

> 
>     * Do you know if there are any performance implications of having mixed types (array or number) in each array? All else equal, I would expect homogeneous arrays to have higher performance, but Javascript may pay a performance price just for supporting heterogeneous arrays, regardless of the contents of any particular array, so it might not make much difference in this case.
> 

The performance of heterogeneous arrays will absolutely be slower than homogeneous arrays, particularly since the alternative homogeneous arrays would be able to be represented with ArrayBuffers. I'd still make an educated guess that the overall performance of this structure in practice will be better, since it will be much smaller in all but the worst-case scenario of needing to go to the block level for all sub-arrays. On the server-side, we'll need to iterate through each item in the array regardless of the structure, and the proposed structure will dramatically cut down on the number of iterations required. The other advantages of this representation are data integrity and transfer speed over the wire.

>     * When a user selects a set of geographies, my recollection is that Mapbox GL provides us with a list of geography IDs that they intend to add to a particular district. Assuming that's correct, what's the process by which that list of geography IDs would get transformed into this representation?
> 
>     * Similar question to the previous, but in reverse: If a user loads up a plan that they've previously created, and the server sends this data structure in a response payload, how does the client-side code use that payload to highlight the appropriate geographies on the map?
> 

I'll answer these two questions together, since they're closely related. First I'd like to point out an important detail for how the district highlighting will work on the front-end. Along with this data structure, we will also be shipping districts in GeoJSON format. The GeoJSON is what will be used for displaying the districts. That is to say, the array representation will not be used for highlighting districts, and will only be minimally used on the front-end. The main reason for this is we will not have access to vector tiles for all geolevels at all zoom levels, so we can't rely on making use of the vector tile polygons for district highlighting purposes.

When a user is highlighting geounits, we'll be making use of Mapbox GL's `setFeatureState` functionality which allows us to efficiently highlight polygons based on their unique ids. In the callback for the mouse interaction event, we will also have full access to all of the vector features (including properties) for all selected polygons. When a user is ready to commit a change to districts, we will have that information, and at that point will update the district array structure in order to send it over to the back-end.

The most efficient way I envision this update to happen is to have the geolevel hierarchy indices present on the properties of each polygon in the vector tile. For example, a block will have the properties: `{countyIndex: 3, blockgroupIndex: 2, blockIndex: 4, ...}`. Updating the structure with this information will be straightforward, since we will be able to traverse down the structure indices to set the district id of the appropriate item. Keep in mind that we'll also have a set of static data loaded on the front-end in ArrayBuffers that will also be able to help out here. i.e. if the block we want to assign doesn't already have the blockgroups expanded for that county, we'll be able to perform a lookup on the static data to assist in performing that expansion before setting the district id for that block.


>     * How is the ordering of geographies within each array and sub-array determined?
> 

When we initially process the block-level GeoJSON, we will be constructing several pieces of data: the TopoJSON (with hierarchy), the vector tiles, and the static data lookups. All of these pieces of constructed data will be relying on the positional indices of the geounits in one way or another, and they will need to remain internally consistent. My initial thought is to just use the ordering of the geounits added to the TopoJSON hierarchy as the source of truth, and build out all data accordingly. It may be a good idea to sort geounits within each level by it's unique id, but I haven't yet found a reason why that would be needed. I think the most important thing is that they're all kept consistent, so lookups are also consistent across all data sources.

>     * What does a real-world district look like when expressed this way? I.e. are real-world districts structured such that they can take advantage of the hierarchical representation, and are there aspects of real-world districts that cause problems for this representation? In other words, where on the proportion-of-blocks scale above do real-world districts tend to fall?

That is a great question, and one that I'm also keenly wanting and answer for. I considered going down that path several times, and only didn't in the interest of time, since it would require doing an extensive amount of up-front work to figure out. I don't think I can realistically perform this investigation at this time as part of this ADR, but agree that it'd be a great thing to work out. From looking at different district configurations, my initial hunch is that the number of times we'd need to go down from the blockgroup level to the block level for a completed set of districts would be somewhere around 25% of the time or less. In the benchmarking results posted in this ADR, that would correspond to the row of 5982 blocks, at 464 milliseconds. But I would really like to be able to have concrete numbers for this across a few real-world sets of districts, and think that would be a great follow-up task.

ddohler:
Awesome, thanks for these responses -- these are all very helpful in understanding what's known and what remains uncertain. This data structure seems like a good idea based on what we know now -- I'm interested to see how everything performs once we can work with real-world data!

PR #28, opened 2020-01-14
Add basic Map component

### LINKED ISSUE #10: Add map with basic data layers
To start off, it would be good to get a basic Mapbox GL map integrated into the application and load in some test vector tiles that contain counties, blockgroups, and blocks for Pennsylvania.

There is code in the prototype repo and experimentation repo that does this. It may not be possible to directly port it over (that would be nice), but it can at least serve as inspiration and guidance.

maurizi:
## Overview

This PR adds a simple map component using `mapbox-gl-js`, showing 1 layer of Census block groups in PA.

### Demo

localhost_3003_

### Notes

I copied the tiles from the [prototype project](https://github.com/azavea/districtbuilder-2/tree/master/public) and uploaded them to the S3 bucket `global-districtbuilder-dev-us-east-1`. I also copied the what styling I used from the prototype.

I only display the block group layer because that was the only one in the prototype project that seemed to be in vector tile format already.

## Testing Instructions

- `scripts/update`
- `scripts/server`

Closes #10

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "@types/react-dom": "16.9.4",
     "@types/react-redux": "7.1.5",
     "axios": "0.19.0",
+    "mapbox-gl": "^1.6.1",
     "react": "16.11.0",
     "react-dom": "16.11.0",
     "react-redux": "7.1.3",
@@ -46,6 +47,7 @@
     ]
   },
   "devDependencies": {
+    "@types/mapbox-gl": "^1.6.1",
     "prettier": "1.18.2",
     "tslint": "5.20.1",
     "tslint-config-prettier": "1.18.0",
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "@types/jest": "24.0.22",
+    "@types/mapbox-gl": "^1.6.1",
     "@types/node": "12.12.6",
     "@types/react": "16.9.11",
     "@types/react-dom": "16.9.4",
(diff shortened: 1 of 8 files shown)


src/components/Map.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { mapboxStyle } from '../constants/map';
+
+const styles = {
+  width: '100%',
+  height: '100%',
+};
+
+const Map = () => {
+  const [map, setMap] = useState<MapboxGL.Map | null>(null);
+  const mapRef = useRef<HTMLDivElement>(null);
```
maurizi: Would be better to disable the lint rule about functional classes and use a class based component for the Mapbox GL interaction here? I would be open to that, it was a little annoying to arrive at this and I still had to add an `eslint-disable-next-line`.

package.json
```diff
@@ -46,6 +47,7 @@
     ]
   },
   "devDependencies": {
+    "@types/mapbox-gl": "^1.6.1",
```
maurizi: Should this be in `dependencies` instead of `devDependencies`? Just now noticing all the other `@types` modules are there.

kshepard: Yeah, probably put it in `dependencies` for consistency. I was just trying to read about when to choose one vs. the other, and I suspect it probably won't end up mattering either way for our use-case.

src/components/Map.tsx
kshepard: I think we can roll with this for now, and refactor it as needed when we learn more. It sounds like there are some competing opinions on how to best use Mapbox GL and React, so I imagine our understanding will evolve as we start adding functionality.

PR #30, opened 2020-02-04
Clean up project docker setup

maurizi:
## Overview

Adds further containerization of project assets, beyond that introduced in #29 

### Demo

N/A

### Notes

Context for this:
 - The existing setup relied on using [Docker's user remapping](https://docs.docker.com/engine/security/userns-remap/), which adds complexity, cutting against our desire for the project to be accessible to outside developers. Without this setup, running `scripts/setup` would succeed on the first run but fail due to file permission issues on subsequent uses.
 - Using `node_modules` on both the frontend and backend can lead to incorrect bindings for any Node libraries that have a compiled native component (e.g. if the host is on Mac, but the container is Linux, or the dependency is linking against a system library).

## Testing Instructions

- `scripts/setup`
- Run `scripts/setup` _again_
- `scripts/server` should start all services
- `scripts/test` should properly execute all linting/tests
- Try out the new `scripts/yarn`
- Verify README changes make sense


README.md
```diff
@@ -71,101 +71,102 @@ Note that there are two `package.json` files. Yarn commands can be run on the fr
...
 Builds the app for production to the `build` folder.
 
 
-### Yarn commands (backend -- run from `src/server`)
+### Yarn commands (server)
 
-### `yarn start|start:dev|start:debug|start:prod`
+### `./scripts/yarn server run start|start:dev|start:debug|start:prod`
 
 Not intended to be used directly. Instead use `scripts/server`, which starts the development server with proper environment variables to allow connecting to the database container.
```
kshepard: I think this note is no longer correct.

PR #42, opened 2020-02-18
Add service and endpoint to retrieve TopoJSON from S3

### LINKED ISSUE #25: Demonstate the ability to load some data into memory upon startup that will be accessible when handling requests
In order to efficiently perform district merge requests, it appears that we'll need to have topology and other static data kept in memory on the server. This task is to demonstrate that we are able to load some mock data on startup that can be accessed across multiple HTTP requests, without duplicating that data or needing to reload on each HTTP request.

tl;dr - How do we do singletons in Nest?

maurizi:
## Overview

Demonstrates how a NestJS singleton service can serve as a cache for data from S3.

The controller endpoint is ultimately likely to be unnecessary, but I added it to demonstrate how to use the TopologyService

### Demo

image

### Notes

This conflicts with changes in @rbreslow's PR. I'm planning to wait until his is merged to address those conflicts.

## Testing Instructions

- `scripts/setup`
- You should be able to download TopoJSON from http://localhost:3005/districts/topology/pa
- After the initial access, the above endpoint should continue to function even after disconnecting your computer's internet connection, because the value is cached.

Closes #25


.env
```diff
@@ -1,3 +1,4 @@
 POSTGRES_USER=district_builder
 POSTGRES_PASSWORD=district_builder
 POSTGRES_DB=district_builder
+AWS_PROFILE=district-builder
```
kshepard: We should probably mention this in the README (failed for me when first testing, since I didn't have the profile set up).

src/server/src/districts/services/topology.service.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+@Injectable()
+export class TopologyService {
+  private readonly layers: { readonly [key: string]: Promise<Topology> };
+  private readonly logger = new Logger(TopologyService.name);
+
+  constructor() {
+    const s3 = new S3();
+    this.layers = mapValues(LAYERS, (url, key) =>
```
kshepard: I just realized that we never turned on `"strict": true` in our `tsconfig.json` compiler options. In this PR, I'm seeing several occurrences of things that will implicitly have an `any` type, which will limit the usefulness of the type checking. We can handle this one of two ways:
 1) Ignore it for now, and create an issue to turn on the strict flag and fix all existing problems
 2) Turn on the flag now and fix the new problems before merging.

I'm fine with either option, and it's more dependent on what your current time constraints are looking like.

.env
rbreslow: When #40 is merged this should be passed through via `docker-compose.yml`.

kshepard:
Nice, I like the updated commit. In order to appease the linter, you'll need to add `"readonly-keyword": false` to tslint.json.

docker-compose.yml
```diff
@@ -20,13 +20,15 @@ services:
       - POSTGRES_USER=districtbuilder
       - POSTGRES_PASSWORD=districtbuilder
       - POSTGRES_DB=districtbuilder
+      - AWS_PROFILE=district-builder
```
rbreslow:
```suggestion
      - AWS_PROFILE=${AWS_PROFILE:-district-builder}
```

PR #53, opened 2020-03-04
Update districts controller to add endpoint to retrieve district GeoJSON

### LINKED ISSUE #32: Create endpoint that takes a district configuration and returns districts GeoJSON
As laid out in #8, district data will be stored in a nested structure of positional geounits. This endpoint will take one of these structures and use TopoJSON stored in memory (as determined in #25) to perform TopoJSON merging, and return GeoJSON districts.

Note: the endpoint will also need a couple other pieces of data to complete this task:
 * A way to determine which TopoJSON file to use
 * A hierarchy of geolevels for the TopoJSON configuration (I don't believe this can be derived from the TopoJSON itself)

maurizi:
## Overview

Uses the topology on S3 as well as a passed in district definition to generate GeoJSON.

### Demo

DE Geojson for districts defined as `[1, 2, [1,1,1,2,2,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]]`
image

## Testing Instructions

- POSTing to the endpoint with a valid district definition should return GeoJSON (e.g. `http --json :3005/districts/de districts:='[1, 2, [1,1,1,2,2,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]]'`)
- POSTing to the endpoint with a _invalid) district definition should return a 400 HTTP status (e.g. `http --json :3005/districts/de districts:='[1, 2, 3, 4]'`)

Closes #32


src/server/src/districts/services/topology.service.ts
```diff
@@ -3,35 +3,66 @@ import S3 from "aws-sdk/clients/s3";
...
+
 const BUCKET = "global-districtbuilder-dev-us-east-1";
-const LAYERS = {
-  pa: "pa/data/bg-lines.topojson"
+const LAYERS: { [key: string]: Layer } = {
+  de: {
+    url: "de/data/DE.topojson",
+    definition: {
+      groups: ["county", "tract", "block"],
+      groupIDs: {
```
kshepard: I think this can be done away with now, since we decided to keep it simple and use the same identifiers for both. We control the topology-generation step, so I don't think there's a benefit to keeping this around. Feel free to get rid of the temporary PA configuration too if that interferes with this change at all.

Issue #52, opened 2020-03-04
User registration pages


tgilcrest:
AC:
* Page where user can enter email address and password to get a email with a registration link
* After submitting, the user sees a page that says "Thanks for signing up for DistrictBuilder! Please click the link in your registration email to continue."
* User can click registration link in email to confirm registration
* Email link takes user to success page that then redirects to home page

maurizi:
We should probably update the copy in the registration email as part of this issue as well.

PR #74, opened 2020-04-02
User registration

### LINKED ISSUE #49: User registration
Endpoints:
* Sign up for account
  * Triggers invitation email
* Uses link in invitation email to complete registration
Models:
* User model(s) 
  * Update existing with fields we need
  * Create new models as necessary (preregistered user?)

AC:
* Send new user payload to registration endpoint
* Email address receives registration email
* Clicking on link in email registers user
* User is now listed on the home page (show that a user has been created)

maurizi:
## Overview

This pull request adds a 2 step user registration system with email verification

### Demo

```
$ http http://localhost:3005/auth/email/register email=mmaurizi@azavea.com password=mike12345 name="Mike M"
HTTP/1.1 201 Created                                                                                                                                                                    
Connection: keep-alive                                                                                                                                                                  
Content-Length: 7
Content-Type: text/html; charset=utf-8                                                                                                                                                  
Date: Thu, 02 Apr 2020 15:39:44 GMT                                                                                                                                                     
ETag: W/"7-gFw9aUPkrb4JBgwx0K+SLEMaCno"                                                                                                                                                 
X-Powered-By: Express                                                                                                                                                                   
                                                                                                                                                                                        
SUCCESS

$ http POST http://localhost:3005/auth/email/resend-verification/mmaurizi@azavea.com
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 7
Content-Type: text/html; charset=utf-8
Date: Thu, 02 Apr 2020 16:13:31 GMT
ETag: W/"7-gFw9aUPkrb4JBgwx0K+SLEMaCno"
X-Powered-By: Express

SUCCESS

$ http POST http://localhost:3005/auth/email/verify/Iku9ORZhSplZzKRSyWoo7vDJn9U
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 114
Content-Type: application/json; charset=utf-8
Date: Thu, 02 Apr 2020 16:13:50 GMT
ETag: W/"72-kDki5zX15WVb9OY36DyWP6gSu3M"
X-Powered-By: Express

{
    "email": "mmaurizi@azavea.com",
    "id": "0614a225-8d2b-4c71-acdc-d66b4869b470",
    "isEmailVerified": true,
    "name": "Mike M"
}

$ http POST http://localhost:3005/auth/email/login email=mmaurizi@azavea.com password=mike12345
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 114
Content-Type: application/json; charset=utf-8
Date: Thu, 02 Apr 2020 16:41:44 GMT
ETag: W/"72-kDki5zX15WVb9OY36DyWP6gSu3M"
X-Powered-By: Express

{
    "email": "mmaurizi@azavea.com",
    "id": "0614a225-8d2b-4c71-acdc-d66b4869b470",
    "isEmailVerified": true,
    "name": "Mike M"
}

$ http POST http://localhost:3005/auth/email/login email=mmaurizi@azavea.com password=mike54321
HTTP/1.1 400 Bad Request
Connection: keep-alive
Content-Length: 74
Content-Type: application/json; charset=utf-8
Date: Thu, 02 Apr 2020 16:41:52 GMT
ETag: W/"4a-J+TcAdnMDmQd2xHUXvqyIXRNQbg"
X-Powered-By: Express

{
    "error": "Invalid password",
    "message": "INVALID_PASSWORD",
    "statusCode": 400
}


```

### Notes

 - I haven't disabled `login` for users whose emails are not verified because I didn't know if we wanted to restrict login entirely or merely restrict access in the application to features that need email.
 - I'm deferring password validation to a future issue, hopefully I can find something that has an equivalent React UI component to validate during registration in #52 
 - We need to add issues Forgot password & change password APIs & frontend pages, we missed those in our initial planning
 - We should add an issue to update the email registration copy / style it.
 - I didn't 1-for-1 copy much, but I used this as a reference work quite a bit: https://github.com/marcomelilli/nestjs-email-authentication/

## Testing Instructions

- `scripts/update`
- Create a user using the `/auth/email/register`endpoint
- Login using `auth/email/login`
- Verify that you see email contents printed to the console running `scripts/server`
- Test resending an email with `/auth/email/resend-verification/:email`
- Using the token in the email printed to the console, verify your user using `auth/email/verify/:token` 

Closes #49


src/shared/constants.ts
```diff
@@ -0,0 +1,28 @@
...
+export enum ResendResponse {
+  SUCCESS,
+  NOT_FOUND,
+  ERROR
+}
+
+export enum VerifyEmailErrors {
+  NOT_FOUND,
+  ERROR
+}
```
maurizi, line 28: @kshepard I wasn't really sure how what the best way to send errors to the client, so I decided to try out having some enums in the `shared` module and sending the string representation of the enum values, which we can then match on in the client.

Happy to switch it up to something else if you have a preferred strategy.

src/server/src/users/entities/user.entity.ts
```diff
@@ -8,4 +11,43 @@ export class User implements IUser {
...
+      this.passwordHash = params.passwordHash;
+    }
+  }
+
+  async comparePassword(password: string): Promise<boolean> {
+    return bcrypt.compare(password, this.passwordHash);
+  }
+
+  async setPassword(password: string): Promise<void> {
+    this.passwordHash = await bcrypt.hash(password, BCRYPT_SALT_ROUNDS);
```
maurizi, line 46: `tslint` is complaining about modification here, but and regrettably this API is inherently mutable.
I originally had password being set in the constructor but `bcrypt` really wants us to use the async version, so I think if we want to eliminate mutation here I have to drop the constructor and instead use an async static `create` method.

kshepard: I think it's pretty expected that we'll run into the occasional case where mutation is a must (particularly on the OOP-heavy Nest backend). I still think having the rule turned on globally is a good thing, and we can be explicit in times where it's needed by using a comment to disable a line/section. So go ahead and do so here.

src/shared/constants.ts
kshepard: No better alternative immediately comes to mind. I think this is great to start with, and we can reevaluate once we start consuming this data on the client.

src/server/src/users/services/users.service.ts
```diff
@@ -2,11 +2,23 @@ import { Injectable } from "@nestjs/common";
 import { InjectRepository } from "@nestjs/typeorm";
 import { TypeOrmCrudService } from "@nestjsx/crud-typeorm";
 import { Repository } from "typeorm";
+
+import { Register } from "../../../../shared/entities";
 import { User } from "../entities/user.entity";
 
 @Injectable()
 export class UsersService extends TypeOrmCrudService<User> {
```
kshepard, line 10: Unsure about this, but should we maybe change this to no longer extend the `TypeOrmCrudService`, or at least restrict it if possible? My worry is that it'll still be possible to use this endpoint to create users, so I want to lock that down.

src/server/src/auth/controllers/auth.controller.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+      // TODO: return JWT instead of user
+      return verifiedUser;
+    } catch (error) {
+      throw new BadRequestException(VerifyEmailErrors[VerifyEmailErrors.ERROR]);
+    }
+  }
+
+  @Post("email/resend-verification/:email")
+  public async sendEmailVerification(@Param("email") email: string): Promise<string> {
```
kshepard, line 98: Errors for `resend-verification` have some strange additional nesting:
```
{
    "error": {
        "message": {
            "error": "User not found for this email",
            "message": "NOT_FOUND",
            "statusCode": 404
        },
        "response": {
            "error": "User not found for this email",
            "message": "NOT_FOUND",
            "statusCode": 404
        },
        "status": 404
    },
    "message": "ERROR",
    "statusCode": 400
}
```

src/server/src/auth/controllers/auth.controller.ts
```diff
@@ -0,0 +1,113 @@
...
+    let userOrError;
+    try {
+      userOrError = await this.authService.validateLogin(login.email, login.password);
+    } catch (error) {
+      throw new BadRequestException(LoginErrors.ERROR);
+    }
+
+    if (userOrError === null || userOrError === LoginErrors.NOT_FOUND) {
+      throw new NotFoundException(
+        LoginErrors[LoginErrors.NOT_FOUND],
```
kshepard, line 50: One discussion point for later (no need to do anything about it now): I think it might be better if we provide fewer details back to the user about why a login failed. Right now it seems too easy to check whether or not a user exists in the system, and that can be exploitable.

src/server/src/auth/entities/login.dto.ts
```diff
@@ -0,0 +1,10 @@
+import { IsEmail, IsNotEmpty } from "class-validator";
+
+import { Login } from "../../../../shared/entities";
+
+export class LoginDto implements Login {
+  @IsEmail()
```
kshepard, line 6: I love these validators, they work great.

src/server/src/auth/controllers/auth.controller.ts
maurizi: This was not an oversight, but intentional. I don't think checking whether a user exists is something we should try and prevent. Given that we are planning to enforce uniqueness on email address on registration, we will intentionally be telling the user during registration whether a user with the requested email exists, so I don't think it is worth it to prevent user enumeration in login.

I think the better option if it is a major concern, would be to prevent against automation on these endpoints by rate limiting or requiring a captcha.

Happy to discuss this further, I know there are multiple opinions on this subject and there are trade-offs between security and user experience.

kshepard: Cool, glad you've given it some thought. It's a good point about being able to get this information during registration. I was thinking about rate limiting too, so maybe that could be a good thing to add in the future.

src/server/src/users/services/users.service.ts
maurizi: Turns out you can restrict it, did so here: https://github.com/PublicMapping/district-builder-2/pull/74/commits/54bcb72d39209457f9150017231aebcd4382de31

src/server/src/auth/controllers/auth.controller.ts
maurizi: Fixed in https://github.com/PublicMapping/district-builder-2/pull/74/commits/88f37ddb32275bdf54abe1700eabec08c5bc0929

PR #79, opened 2020-04-09
Add JWT support for auth on server

### LINKED ISSUE #50: User Authentication
Endpoint:
* Login
  * Use JWT

AC:
* Can retrieve JWT from endpoint with username/password
* Valid JWT will return response for existing endpoints (districts, users), except healthcheck
* Invalid/missing JWT will return appropriate status code

pcaisse:
## Overview

Add JWT support for auth on server.

Verifying email address and logging in now return a JWT which is then required to access any endpoint with the `@UseGuards(JwtAuthGuard)` decorator.

### Demo

```
09:36 $ http http://localhost:3005/auth/email/register email=pc@azavea.com password=test name="Peter C"
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 7
Content-Type: text/html; charset=utf-8
Date: Thu, 09 Apr 2020 13:49:45 GMT
ETag: W/"7-gFw9aUPkrb4JBgwx0K+SLEMaCno"
X-Powered-By: Express

SUCCESS

09:49 $ http POST http://localhost:3005/auth/email/resend-verification/peterc@azavea.com
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 7
Content-Type: text/html; charset=utf-8
Date: Thu, 09 Apr 2020 13:49:51 GMT
ETag: W/"7-gFw9aUPkrb4JBgwx0K+SLEMaCno"
X-Powered-By: Express

SUCCESS

09:49 $ http POST http://localhost:3005/auth/email/verify/bEyUZJ5EvmGwolj_6CfBNl_kjSk
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 337
Content-Type: text/html; charset=utf-8
Date: Thu, 09 Apr 2020 13:50:11 GMT
ETag: W/"151-7lVQ0pAulid4bQvlMsAftSFuR9k"
X-Powered-By: Express

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJiNzZlYTk4OS1jYThjLTQyZWYtYjk4My00MWJmZTg4ZjAwMTIiLCJpZCI6ImI3NmVhOTg5LWNhOGMtNDJlZi1iOTgzLTQxYmZlODhmMDAxMiIsIm5hbWUiOiJQZXRlciBDIiwiZW1haWwiOiJwZXRlcmNAYXphdmVhLmNvbSIsImlzRW1haWxWZXJpZmllZCI6dHJ1ZSwiaWF0IjoxNTg2NDQwMjExLCJleHAiOjE1ODY2MTMwMTF9.S70DAhMZvnK_bTHo4u55cdhshhP5VrhdiIK_juwKoto

09:50 $ http http://localhost:3005/auth/email/login email=peterc@azavea.com password=test
HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 337
Content-Type: text/html; charset=utf-8
Date: Thu, 09 Apr 2020 13:53:18 GMT
ETag: W/"151-HVHdlcX+aTDK7F+3E06F7idk1QE"
X-Powered-By: Express

eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJiNzZlYTk4OS1jYThjLTQyZWYtYjk4My00MWJmZTg4ZjAwMTIiLCJpZCI6ImI3NmVhOTg5LWNhOGMtNDJlZi1iOTgzLTQxYmZlODhmMDAxMiIsIm5hbWUiOiJQZXRlciBDIiwiZW1haWwiOiJwZXRlcmNAYXphdmVhLmNvbSIsImlzRW1haWxWZXJpZmllZCI6dHJ1ZSwiaWF0IjoxNTg2NDQwMzk4LCJleHAiOjE1ODY2MTMxOTh9.iarmOv47Tm4u1bk2E-vRxKuqlAT6JStxnhhypsFLux8

11:49 $ http http://localhost:3005/users 'Authorization:Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJiNzZlYTk4OS1jYThjLTQyZWYtYjk4My00MWJmZTg4ZjAwMTIiLCJpZCI6ImI3NmVhOTg5LWNhOGMtNDJlZi1iOTgzLTQxYmZlODhmMDAxMiIsIm5hbWUiOiJQZXRlciBDIiwiZW1haWwiOiJwZXRlcmNAYXphdmVhLmNvbSIsImlzRW1haWxWZXJpZmllZCI6dHJ1ZSwiaWF0IjoxNTg2NDQ0NDM3LCJleHAiOjE1ODY2MTcyMzd9.EQkoDGMhA35rxm07OVBQtGOTBcdUrrfaw30XEuiWDWQ'
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 85
Content-Type: application/json; charset=utf-8
Date: Fri, 10 Apr 2020 15:49:33 GMT
ETag: W/"55-9Ga/LQZ+0gPa0ku5RMDlnWyuQ50"
X-Powered-By: Express

{
    "id": "b76ea989-ca8c-42ef-b983-41bfe88f0012",
    "isEmailVerified": true,
    "name": "Peter C"
}

11:49 $ http http://localhost:3005/users
HTTP/1.1 401 Unauthorized
Connection: keep-alive
Content-Length: 41
Content-Type: application/json; charset=utf-8
Date: Fri, 10 Apr 2020 16:07:06 GMT
ETag: W/"29-O6gP33ALmcYdbLhsI8ncjapAJ44"
X-Powered-By: Express

{
    "error": "Unauthorized",
    "statusCode": 401
}
```

### Notes

This implementation was largely based on the official docs: https://docs.nestjs.com/techniques/authentication

It includes a few refactors:
1. Remove the generic `ERROR` value from the error enums in `constants.ts`. Those generic errors weren't actually being returned by any function or consumed in any way except in the messages of exceptions at the very end. It seems like we should only have our services return _specific_ errors since if there's an unexpected error, the function will throw, so there doesn't seem to be much of a need to encode generic errors in types.
1. Relatedly, ddcbdf5 tries to bring uniformity to the way we're handling errors in the auth controller (see that commit for details). Essentially, because any service function could throw -- and it seems these 3rd party libraries often do -- the idea is to wrap the happy path in a `try` block and then inspect the error in the `catch`. For example:
```ts
    try {
      // this could throw (unexpected exception)
      const verifiedUser = await this.authService.verifyEmail(token);
      if (verifiedUser === undefined) {
        // here we intentionally throw to return a 404
        throw new NotFoundException(
          VerifyEmailErrors[VerifyEmailErrors.NOT_FOUND],
          "Email or user not found for token"
        );
      }
      // this could also throw
      return this.authService.generateJwt(verifiedUser);
    } catch (error) {
      if (error instanceof HttpException) {
        throw error;
      } else {
        this.logger.error(`Error verifying email token: ${error}`);
        throw new InternalServerErrorException();
      }
    }
```
The nice thing about using classes for exceptions is that we can use `instanceof` to determine what type of exception it is.

The motivation for this was errors cropping up that weren't caught by the type system (eg. if you fail to pass a `sub` into `JwtService.sign`, it will throw unexpectedly) since many of these functions take imprecise types and TS doesn't have any way to account for the fact that a function could throw. That combined with the fact that NestJS has you throw errors to return the appropriate HTTP status code made it hard to work with as it was before.

This seems like a pretty safe approach in that we'll be sure to catch errors, log appropriately, and not bubble up sensitive data, but it is boilerplate-y (perhaps we should discuss more how to better handle it).

## Testing Instructions
I think due to all the changes it's worth going through the full auth flow for testing as shown in the demo:

* Register a new user
* Re-request verification email
* Use the verification code to get a JWT
* Log in with that user (should get the same JWT back)
* Use that JWT to list users
* Make that same request without the JWT and ensure you get a 401

Closes #50


package.json
```diff
@@ -47,6 +51,7 @@
     "@types/jest": "24.0.22",
     "@types/mapbox-gl": "1.6.1",
     "@types/node": "12.12.6",
+    "@types/passport-jwt": "3.0.3",
```
maurizi, line 54: This is the wrong `package.json` / `yarn.lock` to modify - this is the one for the frontend app.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "PORT=3003 react-scripts start",
+    "start": "react-scripts start",
```
maurizi, line 6: :tada:

package.json
pcaisse: Whoops! :facepalm: I'll fix that

src/server/src/users/controllers/users.controller.ts
```diff
@@ -32,4 +34,10 @@ import { UsersService } from "../services/users.service";
 @Controller("users")
 export class UsersController implements CrudController<User> {
   constructor(public service: UsersService) {}
+
+  @UseGuards(JwtAuthGuard)
+  @Get("profile")
+  getProfile(@Request() request: any): User {
+    return request.user;
+  }
```
maurizi, line 46: I _think_ this controller endpoint is superfluous because the auto-generated `getOneBase` endpoint should do the same thing (and the `CrudAuth` decorator should function the same as the `JwtAuthGuard`?)

pcaisse: Hmm... I tried deleting that endpoint and I got this rather cryptic error (which seems to be what happens when you try to access an endpoint that doesn't exist):
```
server_1    | [Nest] 40   - 04/09/2020, 3:13:16 PM   [ExceptionsHandler] ENOENT: no such file or directory, stat '/home/node/app/server/dist/server/src/static/index.html' +1982ms
server_1    | Error: ENOENT: no such file or directory, stat '/home/node/app/server/dist/server/src/static/index.html'
```

pcaisse: Did you want me to try to get it working without the `JwtAuthGuard` decorator? If we have to remember to decorate controller methods then we could run the risk of forgetting and accidentally not requiring authorization to access an endpoint, but I'm hesitant to try to get it working without an explicit decorator and add even more magic.

pcaisse: I pushed a commit (2aa3474) to require auth for all the endpoints in the user controller

PR #80, opened 2020-04-13
Log in to app 

### LINKED ISSUE #51: Add user login page
AC:
* Page where user can enter username/password and submit to authenticate
* User has a set login session

pcaisse:
## Overview
Implement app login.

The approach taken in this PR is after successful login the JWT is saved in local storage and also in `axios`'s headers config so it's automatically included in subsequent requests (the saved JWT is also automatically loaded from local storage when the app starts). Then when private routes are matched, we check the expiration in the JWT payload to make sure the user is logged in and their session hasn't expired yet. If they haven't logged in or their session has expired, we send them to the login page.

### Demo
district_builder_login

### Notes
~As noted in the comments in https://github.com/PublicMapping/district-builder-2/pull/79 the `/api/users` endpoint only returns a single user for reasons unkown, which is why "Users: none" is being displayed on the project page~ This is addressed in c6268bb 

Prior to this PR, proxying was only configured for the `/users` path. This includes a change to put all routes behind `/api` and proxy that (see 853096f). It seemed like the official way to do this is by using [app.setGlobalPrefix](https://docs.nestjs.com/faq/global-prefix). However, when I tried that the routes would match but all validation was broken (no requests would validate) which lead me to address that issue via 6687b59

## Testing Instructions
* Run `./scripts/update`
* Restart app
* Try to visit project without logging in and you should be redirected to the login page
* Ensure you have a registered user (can do it like so):
```
$ http http://localhost:3003/api/auth/email/register email=<email> password=<password> name=<name>
```
* Log in with bad creds and you should see an error message
* Log in successfully and you should be redirected to the main page
* Click the link to the project and you should be able to view it
* The logged in user's email address should be displayed
* Refresh and you should still be able to view it

Closes #51


src/client/screens/LoginScreen.tsx
```diff
@@ -0,0 +1,63 @@
...
+          onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+            setLoginForm({
+              ...loginForm,
+              password: e.currentTarget.value
+            })
+          }
+        />
+      </div>
+      <div>
+        <button onClick={() => store.dispatch(authenticate(loginForm))}>Log in</button>
```
kshepard, line 50: If it's not too difficult, it'd be great to make it so pressing the enter key triggers this action. This probably applies to registration as well.

kshepard:
Also, thanks for making the change to put everything behind `/api`: this was on my TODO list for a while.

PR #81, opened 2020-04-13
Containerize "manage" CLI project

### LINKED ISSUE #66: Containerize "manage" CLI project
## Overview

We've been getting away so far without having it containerized, but it's getting to the point where we now have some additional dependencies that need to be setup (e.g. `tippecanoe`), and we'll soon need database access.

## Acceptance criteria

The "manage" CLI project is containerized and integrated with CI (e.g. code is linted and has the ability to fail a build)

kshepard:
## Overview

Containerizes the "manage" CLI project, since it's got to the point where we now have some more complex dependencies that need to be setup (e.g. `tippecanoe`), and we'll soon need database access.

### Demo

db-containerize-cli

### Notes

 * This does not add database access, since that will be coming in the following task, but some DB/AWS credentials have been added to the environment in the `docker-compose` file in preparation.
 * The `.prettierrc` file in the root wasn't making its way to the containers, which was causing formatting to behave differently inside and outside the container. Because the config is trivially small, I copied the file to server/manage, rather than complicating the Docker setup. If the `prettier` config ever becomes more complicated, we may want to unify the files.
 * I added a `.gitignore`d directory, `data`, as a place that can be used for storing input/output of the processing command.
 * I added `NODE_OPTIONS=--max-old-space-size=14336` to the `docker-compose` environment vars. 14336MB is the maximum we've needed so far (with California), and there doesn't appear to be any harm in having this set even for a smaller processing job, since it doesn't pre-allocate memory, and is more of a maximum that it's allowed to grow to.
 * @hectcastro: Added you as a reviewer, since you requested Ops involvement on this one. Feel free to reassign as appropriate.

## Testing Instructions

-  `./scripts/update`
- `./scripts/manage`
- You should see command output similar to the above screenshot
- That should be good enough for testing. But if you'd also like to see `tippecanoe` in action, you can do the following to process a GeoJSON file:
  - `mkdir src/manage/data/output-de`
  - Copy `DE.geojson` to `src/manage/data/`
  - `./scripts/manage process-geojson data/DE.geojson -o data/output-de -n 12,4,4 -x 12,12,12 -l block,tract,county`
  - Verify the expected files (topojson, static data, and vector tiles) have been created under `output-de`.

Closes #66


kshepard:
Please hold off on reviewing this for a little bit, I'm making some changes.

kshepard:
Alright, it's ready for review.

docker-compose.yml
```diff
@@ -48,3 +48,28 @@ services:
...
+      - ./src/server:/home/node/app/server
+      - ./src/shared:/home/node/app/shared
+      - ./src/manage:/home/node/app/manage
+      - $HOME/.cache/district-builder:/usr/local/share/.cache/yarn
+      - $HOME/.aws:/root/.aws:ro
+    working_dir: /home/node/app/manage
+    depends_on:
+      database:
+        condition: service_healthy
+    command: bash
```
maurizi, line 75: Consider breaking this out into a separate `docker-compose.manage.yml` file so that `scripts/server` doesn't bring up the `manage` container unnecessarily. Not necessary as it doesn't really hurt anything or take up significant resources, but it is a little odd for that to happen.

Alternatively, we could change `scripts/server` to explicitly only bring up the other 3 services.

hectcastro:
Planning to review this afternoon!

docker-compose.yml
kshepard: Good call. I went with the `scripts/server` modification route.

src/manage/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+    procps=2:3.3* \
+    " \
+    && apt-get update && apt-get install -yq $deps --no-install-recommends
+
+COPY package.json /home/node/app/manage
+COPY yarn.lock /home/node/app/manage
+RUN set -ex \
+    && yarn install
+
+COPY --from=tippecanoe /usr/local/bin/tippecanoe /usr/local/bin/tippecanoe
```
hectcastro, line 39: This makes use of Docker's [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/) support to isolate the Tippecanoe build in the stage above and just copy the binary into this stage.

I used the testing instructions to convert `DE.geojson` and it appeared to work correctly.

src/manage/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+
+CMD ["/home/node/app/manage/bin/run"]
```
hectcastro, line 41: Should this made the container image `ENTRYPOINT` vs. `CMD`? If switched to `ENTRYPOINT`, future invocations would only need to pass `process-geojson` as an argument to something like `docker-compose run`.

src/manage/Dockerfile
kshepard: This looks awesome, thanks for making that change.

src/manage/Dockerfile
kshepard: I also had that thought and originally used `ENTRYPOINT` in order to forego supplying `./bin/run` in the `./scripts/manage` invocation. Unfortunately, there are other scripts, such as `./scripts/yarn` which take client/server/manage as an argument and assume a common interface among all of those containers. Therefore, I stuck with `CMD` as it was the path of least resistance.

PR #82, opened 2020-04-14
Register new user account

### LINKED ISSUE #52: User registration pages

pcaisse:
## Overview

Implement user registration

### Demo

district_builder_register

### Notes

I opted to just use React component state instead of using Redux, instead storing the request state in the component itself. Redux seemed like overkill since the state we want to track is dead simple, we don't care about the success response (just that the request succeeded/failed), and it's hard to imagine any other page needing to know about that state. The advantage of this is we save a bunch of boilerplate, with the potential disadvantage that it's inconsistent with how we do it other places and violates separation of concerns. I'm fine with changing it to do it the Redux/redux-loop way if anyone feels strongly about it.

## Testing Instructions
* Restart server
* Load app in browser and click "Register" link on main page
* Fill out the form with an invalid email and ensure an error message is shown
* Fill out the form with all required fields and a new, valid email address
* Should see the success page telling you to check your email
* Check the console and follow the link that shows up in the email
* You should see another success page and be logged in
* Click link back to main page and then click the project link
* Should see the email that you registered with displayed

Closes #52

### DIFF
--- a/src/client/screens/RegistrationScreen.tsx
+++ b/src/client/screens/RegistrationScreen.tsx
@@ -1,0 +1,75 @@
+import React, { useState } from "react";
+
+import { Register } from "../../shared/entities";
+import { registerUser } from "../api";
+import { Resource } from "../types";
+
+const isFormValid = (form: RegistrationForm): boolean =>
+  Object.values(form).some(value => value.trim() === "") || form.password !== form.confirmPassword;
+
+interface RegistrationForm extends Register {
+  readonly confirmPassword: string;
+}
+
+const RegistrationScreen = () => {
+  const [registrationForm, setRegistrationForm] = useState<RegistrationForm>({
+    email: "",
+    password: "",
+    confirmPassword: "",
+    name: ""
+  });
+  const [registrationResource, setRegistrationResource] = useState<Resource<void>>({
+    isPending: false
+  });
+  const setForm = (field: keyof RegistrationForm) => (e: React.ChangeEvent<HTMLInputElement>) =>
+    setRegistrationForm({
+      ...registrationForm,
+      [field]: e.currentTarget.value
+    });
+  return "resource" in registrationResource ? (
+    <div>
+      Thanks for signing up for DistrictBuilder! Please click the link in your registration email to
+      continue.
+    </div>
+  ) : (
+    <>
+      <div>
+        <input type="text" placeholder="Name" onChange={setForm("name")} />
+      </div>
+      <div>
+        <input type="text" placeholder="Email" onChange={setForm("email")} />
+      </div>
+      <div>
+        <input type="password" placeholder="Password" onChange={setForm("password")} />
+      </div>
+      <div>
+        <input
+          type="password"
+          placeholder="Confirm password"
+          onChange={setForm("confirmPassword")}
+        />
+      </div>
+      <div>
+        <button
+          onClick={() => {
+            setRegistrationResource({ isPending: true });
+            registerUser(registrationForm.name, registrationForm.email, registrationForm.password)
+              .then(() => setRegistrationResource({ resource: void 0 }))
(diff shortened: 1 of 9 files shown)


jfrankl:
This looks like a good start. What would it take to change this registration workflow so the user doesn't need to check their email to confirm before they can start using the tool? That is how Groundwork handles registrations and I think we should do the same.

pcaisse:
@jfrankl It already works that way. The user can still authenticate and view the project page without having verified their email. The only difference is a `isEmailVerified` flag being set to `false` on the user object.

pcaisse:
Here's an example of a user that hasn't been verified yet (can still view project):
district_builder_user_not_verified

jfrankl:
@pcaisse Great, sounds like this general structure will be compatible with what I have in mind. I'm working on some updated DB designs today and will include login and registration.

src/client/screens/RegistrationScreen.tsx
```diff
@@ -0,0 +1,75 @@
...
+            registerUser(registrationForm.name, registrationForm.email, registrationForm.password)
+              .then(() => setRegistrationResource({ resource: void 0 }))
+              .catch(errorMessage => setRegistrationResource({ errorMessage }));
+          }}
+          disabled={
+            ("isPending" in registrationResource && registrationResource.isPending) ||
+            isFormValid(registrationForm)
+          }
+        >
+          Log in
```
kshepard, line 65: This should instead be something along the lines of: `Register`.

src/client/screens/RegistrationScreen.tsx
```diff
@@ -0,0 +1,75 @@
...
+        </button>
+      </div>
+      {"errorMessage" in registrationResource ? (
+        <div style={{ color: "red" }}>{registrationResource.errorMessage}</div>
```
kshepard, line 69: This may be a more involved change (e.g. needs another issue), but these error messages aren't displaying all the information we'd like. For example, if I attempt to register with a user and provide an invalid email address, I see the error `Request failed with status code 400`, but there is information on the response stating that it wasn't a valid email. Was curious if you've given any thought to good ways to handle this.

src/client/screens/RegistrationScreen.tsx
pcaisse: Woops! I'll fix that

src/client/screens/RegistrationScreen.tsx
pcaisse: Yeah, I wasn't sure how exactly how far to go with it given that this was intended to be POC (i.e. if I should go ahead and parse the response and try to display the error nicely). We're currently getting back pretty granular, field level validation information that we could certainly use. Here's the response for the invalid email address example:
```
{
    "statusCode":400,
    "error":"Bad Request",
    "message":[
        {
            "target":{
                "name":";lskjdf",
                "email":"lkdjf",
                "password":"test"
            },
            "value":"lkdjf",
            "property":"email",
            "children":[

            ],
            "constraints":{
                "isEmail":"email must be an email"
            }
        }
    ]
}
```
I think the next step would be to parse the error response to get out the specific validation errors, though I wasn't sure how exactly it should be displayed (perhaps an error should show up next to the problematic field instead of all grouped together if there are multiple errors). Also, the parsing of the error response may require some decoder library or a utility libraries like lodash, so I wasn't sure I wanted to go down that road.

Do you think I should spend time on it as part of this?

kshepard: Thanks for that. No, I think it'd be overkill to do right now, but wanted to hear your thoughts before I wrote up an issue.

PR #89, opened 2020-04-23
Create and list projects

### LINKED ISSUE #69: Create project page
## Overview

We need a page and related backend components (db model, endpoint) that will allow a user to create a new project (a.k.a. a map).

Wireframes: https://app.abstract.com/projects/9f988240-1dad-11e8-9a40-cba930331af7/branches/45b88b25-78cd-45f9-91f1-3214894201bb/commits/5bf9bf36f176209a90fa18108b90c527d159ad15/files/DCA38482-FC7F-4F48-BECD-44E07D9F7443/layers/315E6576-47E5-41F8-A9FE-8331CB417962

Let's start simple with just a few fields:
 * Project name (string text box)
 * State dropdown (populated via hitting an endpoint)
 * Number of districts (number text box)

## Acceptance criteria
 * A new page exists with the three fields mentioned above, and a `Create project` button
 * Clicking the button will save the project to the database and route to the project page with it's generated id

### LINKED ISSUE #70: List a user's projects on the home page
## Overview

After a user is able to create projects, we need a way to show them which projects they have available. There are no wireframes for this, but a simple list of links should be fine for the time being.

## Acceptance criteria

When a user is logged in, their homepage shows a list of links for the projects that they've created.

pcaisse:
## Overview
Create and list projects.

### Demo
district_builder_create_project

## Notes
I added listing projects to this PR to make it demoable without realizing there was a separate issue for it (#70 )

This PR includes a couple of very small unrelated changes (0317b2e and be25bed)

## Testing Instructions
As per the demo above:
* Ensure you have at least one region config (there's no UI for this yet but there are examples of how to do this in https://github.com/PublicMapping/district-builder-2/pull/87)
* Go to the main page and click 'Create project'
* Fill out the form and should be able to create a project successfully
* Should be redirected to home page and see your new project listed

Then log in as a different user and make sure don't see any projects.

Closes #69 
Closes #70


pcaisse:
@maurizi Sorry to do this but hold off on reviewing this for now. I realized I need to make some more changes to ensure that the FKs are saved correctly and are used for filtering the projects.

pcaisse:
@maurizi Okay, this should be good for review now :sweat_smile: 

I'm not super happy with how it turned out but I did succeed in 1) making user id and region config id required to create a project, 2) making those be set correctly when creating a project, and 3) filtering by the currently logged in user when listing projects _without_ having the frontend pass any extra user info to the backend.

It seems to me that there should be an easier way to access the authenticated user's data that could be applied globally and wouldn't require overriding any methods as in `ProjectsController.createOne`. For regular non-`CrudController` controllers, it seems that having a route requiring JWT auth which uses the JWT strategy as described in the [authentication docs](https://docs.nestjs.com/techniques/authentication#jwt-functionality) -- exactly what we're doing -- should make it so that the user is on the request (should be accessible via `req.user`). Indeed, when doing _filtering_ in the `@CrudAuth` decorator in [users.controller.ts](https://github.com/PublicMapping/district-builder-2/blob/develop/src/server/src/users/controllers/users.controller.ts#L26-L33), it seems that telling it what property to use for the user and then setting `filter` appropriately is sufficient and this is working fine for us. In terms of adding the user to the request, answers like [this one](https://stackoverflow.com/a/53264162) and [the Controller docs](https://github.com/nestjsx/crud/wiki/Controllers#request-authentication) lead me to believe that it should be a simple matter of setting `property` to have the user data be merged into the request object after successful authentication. However, that doesn't seem to work for `CrudController`s and due to the vagaries of NestJS and the magic-ness of how all this works I'm unsure of why it's not added to the request. Another wonky piece to this is that the types don't quite line up properly... the `ProjectController` gets a `ProjectDto` passed into its `createOne` but the service seems to want a `Project` so I had to do some finagling to get the types to type check. I'm also not even really sure where the `createOne` function I'm overriding is even coming from since I don't see it in the `Controller` or `CrudController` types.

In any case, I did find a way to access the user data I needed after successful authentication via `@CrudAuth`'s `persist` which allows you add user data to the DTO (this is covered in the Controller docs linked to above). Unfortunately, it doesn't seem like you can [configure this globally](https://github.com/nestjsx/crud/wiki/Controllers#global-options) as you can with `property` (I tried configuring that locally and globally and it didn't seem to work in either case).

Anyway, that's my story. Hope it makes sense!

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -0,0 +1,56 @@
...
+  @Override()
+  async createOne(
+    @ParsedRequest() req: CrudRequest,
+    @ParsedBody() dto: ProjectDto
+  ): Promise<Project> {
+    return await this.service.createOne(req, {
+      ...dto,
+      regionConfig: (dto.regionConfigId as unknown) as RegionConfig,
+      user: req.parsed.authPersist.userId
+    });
```
maurizi, line 54: The following changes would fix the type issues (though I think you'll need to create a new `region-config-id.dto.ts` to avoid lint):
```diff
diff --git a/src/client/api.ts b/src/client/api.ts
index a10d71e..e22ac4f 100644
--- a/src/client/api.ts
+++ b/src/client/api.ts
@@ -69,7 +69,7 @@ export async function createProject(
 ): Promise<IProject> {
   return new Promise((resolve, reject) => {
     axios
-      .post(`/api/projects`, { name, numberOfDistricts, regionConfigId })
+      .post(`/api/projects`, { name, numberOfDistricts, regionConfig: { id: regionConfigId } })
       .then(response => resolve(response.data))
       .catch(error => reject(error.message));
   });
diff --git a/src/server/src/projects/controllers/projects.controller.ts b/src/server/src/projects/controllers/projects.controller.ts
index 9808456..e7a2661 100644
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -47,10 +47,6 @@ export class ProjectsController implements CrudController<Project> {
     @ParsedRequest() req: CrudRequest,
     @ParsedBody() dto: ProjectDto
   ): Promise<Project> {
-    return await this.service.createOne(req, {
-      ...dto,
-      regionConfig: (dto.regionConfigId as unknown) as RegionConfig,
-      user: req.parsed.authPersist.userId
-    });
+    return await this.service.createOne(req, { ...dto, user: req.parsed.authPersist.userId });
   }
 }
diff --git a/src/server/src/projects/entities/project.dto.ts b/src/server/src/projects/entities/project.dto.ts
index f3edef2..ac0f9f0 100644
--- a/src/server/src/projects/entities/project.dto.ts
+++ b/src/server/src/projects/entities/project.dto.ts
@@ -1,12 +1,17 @@
 import { IsInt, IsNotEmpty, IsUUID } from "class-validator";
 
-import { CreateProjectData, RegionConfigId } from "../../../../shared/entities";
+import { CreateProjectData, ExistingObject, RegionConfigId } from "../../../../shared/entities";
+
+class RegionConfigIdDto implements ExistingObject<RegionConfigId> {
+  @IsUUID()
+  readonly id: RegionConfigId;
+}
 
 export class ProjectDto implements CreateProjectData {
   @IsNotEmpty()
   readonly name: string;
   @IsInt()
   readonly numberOfDistricts: number;
-  @IsUUID()
-  readonly regionConfigId: RegionConfigId;
+
+  readonly regionConfig: RegionConfigIdDto;
 }
diff --git a/src/server/src/projects/entities/project.entity.ts b/src/server/src/projects/entities/project.entity.ts
index d331969..f6003b6 100644
--- a/src/server/src/projects/entities/project.entity.ts
+++ b/src/server/src/projects/entities/project.entity.ts
@@ -1,4 +1,4 @@
-import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from "typeorm";
+import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn, RelationId } from "typeorm";
 import { IProject } from "../../../../shared/entities";
 import { RegionConfig } from "../../region-configs/entities/region-config.entity";
 import { User } from "../../users/entities/user.entity";
diff --git a/src/shared/entities.d.ts b/src/shared/entities.d.ts
index c3b40d7..d25cd8c 100644
--- a/src/shared/entities.d.ts
+++ b/src/shared/entities.d.ts
@@ -1,3 +1,7 @@
+export interface ExistingObject<T> {
+  readonly id: T;
+}
+
 export type UserId = string;
 
 export interface IUser {
@@ -61,5 +65,5 @@ export interface IProject {
 export interface CreateProjectData {
   readonly name: string;
   readonly numberOfDistricts: number;
-  readonly regionConfigId: RegionConfigId;
+  readonly regionConfig: ExistingObject<RegionConfigId>;
 }
```

src/client/actions/projectForm.ts
```diff
@@ -0,0 +1,50 @@
+import { IProject, RegionConfigId } from "../../shared/entities";
+
+export interface ProjectForm {
+  readonly name: string;
+  readonly numberOfDistricts: number;
+  readonly regionConfigId: RegionConfigId | null;
+}
```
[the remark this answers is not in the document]
pcaisse: Yeah, I deliberated on this at one point as well. I think it's worth keeping the form as a separate interface because the form should encompass possible invalid states as well as valid ones (eg. `regionConfigId` is initially null in `ProjectForm` since no default value is selected). I wouldn't want to make the `regionConfigId` in `CreateProjectData` nullable for the sake of reuse since it shouldn't ever be null.

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
       - POSTGRES_DB=districtbuilder
       - AWS_PROFILE=${AWS_PROFILE:-district-builder}
       - JWT_SECRET=insecure
-      - JWT_EXPIRATION_IN_MS=172800000
+      - JWT_EXPIRATION_IN_MS=2629800000  # 1 month
```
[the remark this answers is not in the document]
pcaisse: I wanted to increase it because I got tired of having to set it :laughing: 

Would one week be a happy medium?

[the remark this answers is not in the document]
maurizi: Yeah that sounds low enough that we'll regularly exercise expiration in dev without being overly annoying. We can always later set a different value for production if we want to.

src/server/src/projects/entities/project.dto.ts
```diff
@@ -1,12 +1,12 @@
...
+import { RegionConfig } from "../../region-configs/entities/region-config.entity";
 
 export class ProjectDto implements CreateProjectData {
   @IsNotEmpty()
   readonly name: string;
   @IsInt()
   readonly numberOfDistricts: number;
-  @IsUUID()
-  readonly regionConfigId: RegionConfigId;
+  readonly regionConfig: RegionConfig;
```
maurizi, line 11: Should this be `Pick<RegionConfigDto, "id">` or perhaps just `RegionConfigDto`? I don't think we should be referencing an Entity class from a DTO

src/shared/entities.d.ts
```diff
@@ -61,5 +61,5 @@ export interface IProject {
 export interface CreateProjectData {
   readonly name: string;
   readonly numberOfDistricts: number;
-  readonly regionConfigId: RegionConfigId;
+  readonly regionConfig: Pick<IRegionConfig, "id">;
```
maurizi, line 64: :eye: Didn't know about `Pick` before, nifty!

src/server/src/projects/entities/project.dto.ts
pcaisse: Yeah, I think you're right. I went back to the way you had it originally.

PR #100, opened 2020-04-30
Show validation errors in registration & login screens

### LINKED ISSUE #85: Display better errors in registration/login
## Overview

There is detailed error information returned from the registration/login endpoints (such as an email being invalid, an email not existing in the database, and a password being invalid), however only a generic error message is being displayed to the user. For additional context, see this thread: https://github.com/PublicMapping/district-builder-2/pull/82#discussion_r408439038.

We should consume the information on the response to give a better error reason to the user. The wireframes for these pages are currently in-progress, so this will affect how these errors are to be displayed (e.g. potentially displaying them in line with the relevant input boxes).

One related change worth looking into here is improving our error enums. Potential improvements would be to use string literals (union type) or string enums.

## Acceptance criteria

Registration and login has more detailed error messages displayed to the user, in accordance with the wireframes.

maurizi:
## Overview

Previously, we only showed generic error messages generated for us by `axios` on these screens (error with status code 400, or similar).

Since our backend does extensive error checking, we instead want to actually show those errors to users on the frontend. This required both 1) reworking how we represent errors on the backend to be in a consistent format and 2) refactoring how errors are represented on the frontend to support more than just a single error message sting.

### Demo

localhost_3003_login (1)
localhost_3003_login
localhost_3003_register (2)
localhost_3003_register (1)
![localhost_3003_register](https://user-images.githubusercontent.com/4432106/80764831-85931280-8b0f-11ea-90c9-2c71f6eb2fd5.png

## Testing Instructions

- `scripts/server`

Closes #85


src/client/screens/RegistrationScreen.tsx
```diff
@@ -2,29 +2,34 @@ import React, { useState } from "react";
...
-  const [registrationResource, setRegistrationResource] = useState<Resource<void>>({
-    isPending: false
+  const [registrationResource, setRegistrationResource] = useState<
+    WriteResource<RegistrationForm, void>
+  >({
+    data: {
+      email: "",
+      password: "",
+      confirmPassword: "",
+      name: ""
```
maurizi, line 23: `tslint` is failing this due to the following: `The key 'email' is not in the same order as it is in 'RegistrationForm'.`

Any idea how I can find out what order it actually wants here? I tried a few variations but none worked.

kshepard: I just tried my hand at this and wasn't able to work it out either. I found an issue reported that's related to this `tslint` error when working with inheritance (as we are here). The issue was closed, because `tslint` has been deprecated. Perhaps it's time to switch to `eslint`. In the meantime, I think it may be best to just disable this rule project-wide to prevent frustration.

pcaisse: So assuming that a bug in `tslint` with `extends` was behind this, I changed a few types slightly so that we avoid using `extends` entirely (extending an interface `interface <x> extends <y> { ... }` and using an intersection of types `type <x> = y & { ... }` work similarly, though TIL [there are some differences](https://stackoverflow.com/a/52681859)). As such, these changes seemed to placate the linter (NOTE: I'm unsure why changing this required a change to `isFormInvalid`, though I still think that's a decent change as it's more explicit):
```
diff --git a/src/client/screens/RegistrationScreen.tsx b/src/client/screens/RegistrationScreen.tsx
index 3a83aee..5635410 100644
--- a/src/client/screens/RegistrationScreen.tsx
+++ b/src/client/screens/RegistrationScreen.tsx
@@ -6,11 +6,12 @@ import { WriteResource } from "../resource";
 import { buildErrors, getErrorMessage, getFieldErrors } from "../utility";

 const isFormInvalid = (form: RegistrationForm): boolean =>
-  Object.values(form).some(value => value.trim() === "") || form.password !== form.confirmPassword;
+  [form.name, form.email, form.password].some(value => value.trim() === "") ||
+  form.password !== form.confirmPassword;

-interface RegistrationForm extends Register {
+type RegistrationForm = Register & {
   readonly confirmPassword: string;
-}
+};

 const RegistrationScreen = () => {
   const [registrationResource, setRegistrationResource] = useState<
index 421c4b5..28e3c87 100644
@@ -26,9 +26,9 @@ export interface Login {
   readonly password: string;
 }

-export interface Register extends Login {
+export type Register = Login & {
   readonly name: string;
-}
+};

 export type JWT = string;
```
I don't know if we want to avoid extending interfaces just because of this linter issue but I think we _could_ to no ill effect :shrug:

pcaisse: All of that said, I think I'd prefer to just turn off that lint rule so as not to introduce confusing gotchas for future contributors to the project

src/server/src/auth/controllers/auth.controller.ts
```diff
@@ -46,15 +47,15 @@ export class AuthController {
     try {
       const userOrError = await this.authService.validateLogin(login.email, login.password);
       if (userOrError === LoginErrors.NOT_FOUND) {
-        throw new NotFoundException(
-          LoginErrors[LoginErrors.NOT_FOUND],
-          `Email ${login.email} not found`
-        );
+        throw new NotFoundException({
```
pcaisse, line 50: :+1:

src/server/src/common/bad-request-exception.filter.ts
```diff
@@ -0,0 +1,38 @@
...
+    const message =
+      "message" in errorResponse &&
+      Array.isArray(errorResponse.message) &&
+      errorResponse.message.every((item: any) => item instanceof ValidationError)
+        ? convertValidationErrors(errorResponse.message)
+        : errorResponse.message;
+
+    response.status(status).json({ error, message });
+  }
+}
```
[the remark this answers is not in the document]
maurizi: I don't think this quite works - for starters, it doesn't support errors that are already in an `ErrorMap` format.

tgilcrest:
Could we say "Invalid Username/Password" for both the login errors instead? I think that's more consistent with how other apps work and is a little more secure since you're not leaking info about the user base.

maurizi:
@tgilcrest I strongly believe it is not worthwhile to attempt to prevent email enumeration by only showing "username or password invalid" messages.

Unless we plan to allow duplicate email signups (complicated & requires adding a username in addition to email), or require going into your email to determine if that email is already taken (terrible UX), we've already left the barn door wide open and changing our error messages is just closing a window. See this blog post for further argumentation on this subject https://kevin.burke.dev/kevin/invalid-username-or-password-useless/

To properly account for email enumeration attacks, the appropriate counter measure is to place rate limiting on login / registration routes (I can write up an issue for it).

tgilcrest:
Ah ok, that definitely makes sense, thanks for explaining!

src/client/utility.tsx
```diff
@@ -0,0 +1,27 @@
...
+    : undefined;
+}
+
+export function getFieldErrors<D, T>(resource: WriteResource<D, T>): ErrorMap<D> {
+  return "errors" in resource && typeof resource.errors.message !== "string"
+    ? resource.errors.message
+    : {};
+}
+
+export function buildErrors<D, F extends keyof D>(
```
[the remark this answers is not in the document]
maurizi: I had done it this way because I didn't want to have to pass a generic type argument in TSX (I find the extra angle brackets in `<span<string> foo="bar" />` to be pretty confusing), but it turns out TypeScript can infer type parameters to functional components just as well as it can with functions!

Resolved in d30bad8

src/client/screens/RegistrationScreen.tsx
maurizi: Rule disabled in https://github.com/PublicMapping/district-builder-2/pull/100/commits/91297b5f81359401489074de887031fb1866b272

src/server/src/auth/controllers/auth.controller.ts
```diff
@@ -63,7 +64,7 @@ export class AuthController {
       } else {
         // Intentionally not logging errors as they may contain passwords
         this.logger.error(`Error logging user in`);
-        throw new InternalServerErrorException();
+        throw new InternalServerErrorException("Unable to log in use");
```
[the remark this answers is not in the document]
maurizi:
```suggestion
        throw new InternalServerErrorException("Unable to log in user");
```

maurizi:
@pcaisse This is ready for another look.

src/client/screens/LoginScreen.tsx
```diff
@@ -1,68 +1,67 @@
...
-        store.dispatch(authenticate(loginForm));
+        setLoginResource({ data, isPending: true });
+        authenticateUser(data.email, data.password)
+          .then(jwt => setLoginResource({ data, resource: jwt }))
+          .catch(errors => {
+            setLoginResource({ data, errors });
+          });
       }}
     >
+      {errorMessage ? <div style={{ color: "red" }}>{errorMessage}</div> : null}
```
pcaisse, line 35: I think it would likely make sense for this to be a component as well for when we get around to styling the non-field-specific error message

PR #131, opened 2020-05-06
Fix convertValidationErrors typing issue

maurizi:
## Overview

PR #100 was unintentionally broken after merging, due to a problem uncovered by adding a stricter type to the `convertValidationErrors` function.

According to the typings `constraints` could be undefined here.
I didn't see that happen when testing but we should guard against it regardless.

## Notes

It's a bit alarming that `scripts/update` & `scripts/server` are broken on `develop` but CI passed for #100.

## Testing Instructions

- `scripts/server` should work

### DIFF
--- a/src/server/src/common/bad-request-exception.filter.ts
+++ b/src/server/src/common/bad-request-exception.filter.ts
@@ -7,11 +7,12 @@ function convertValidationErrors(
 ): {
   [field: string]: readonly string[];
 } {
-  return Object.fromEntries(
-    errors.map((error: ValidationError) => {
-      return [error.property, Object.values(error.constraints)];
+  const fieldErrors = errors
+    .map((error: ValidationError) => {
+      return error.constraints ? [error.property, Object.values(error.constraints)] : undefined;
     })
-  );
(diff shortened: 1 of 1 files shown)


pcaisse:
There's something really weird going on here... I saw the error originally but after switching back and forth between branches a few times I don't see it anymore (no compilation errors). Also, judging by the types it doesn't seem like `constraints` should ever be `undefined`:
```
    /**
     * Constraints that failed validation with error messages.
     */
    constraints: {
        [type: string]: string;
    };
```

maurizi:
Ah. This is a difference between `class-validator` versions [`0.11`](https://github.com/typestack/class-validator/blob/v0.11.1/src/validation/ValidationError.ts#L25-L30) and [`0.12`](https://github.com/typestack/class-validator/blob/v0.12.2/src/validation/ValidationError.ts#L25-L30) - `constraints` was made optional in `0.12`.

So really this should just be a part of #132 where we upgrade the library (which probably explains why we both ran into this - that would happen even in `develop` after forgetting to rerun `scripts/update` once you reprovision with newer dependencies).

PR #134, opened 2020-05-08
Add theme-ui and componentize login screen

### LINKED ISSUE #93: Integrate a component library
## Overview

The app currently has no styling. We would like to start working towards being able to approximate the wireframes when creating new features. In order to do so, we need to select a component library and have a standard framework for adding styling. The design team should be consulted on this, since they may have an informed opinion. 

## Acceptance criteria
 * A component library is selected and added to the application
 * One feature of your choosing has been adapted to make use of the component library. This could be something simple, such as the login form. Just something to prove that we are able to make use of the library.

maurizi:
## Overview

Adds [`theme-ui`](https://theme-ui.com/) and uses it to componentize the login screen

### Demo

localhost_3003_login (4)
localhost_3003_login (5)

## Testing Instructions

- `scripts/update`
- `scripts/server`
- Go to the login screen
  - attempt to login with a missing user or incorrect password to see errors
  - verify login still works correctly

Closes #93

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,6 +27,7 @@
     "react-scripts": "3.2.0",
     "redux": "4.0.4",
     "redux-loop": "5.0.1",
+    "theme-ui": "0.4.0-alpha.0",
     "ts.data.json": "0.3.0",
     "typesafe-actions": "5.1.0",
     "typescript": "3.7.5"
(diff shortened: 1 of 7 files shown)


maurizi:
> Just curious: what exactly are the jsx pragmas doing? I tested out removing them, along with the imports, and didn't notice any immediately visible changes.

I have absolutely no idea :man_shrugging: They're called out in the docs as required.

kshepard:
Interesting. The fact that it works fine when I take them out makes me think that they're probably not necessary, but I haven't researched this. It would be nice to not have to include these two lines in every file if we don't have to.

PR #138, opened 2020-05-14
Export project GeoJSON

### LINKED ISSUE #92: Export map to geojson
## Overview

We already have an endpoint that, given a district configuration will return geojson with no properties. This task is to create another endpoint that will look up a plan in the database, calculate demographic properties (through the use of the static data), and return the geojson. It will eventually used by the frontend to allow a user to download a geojson version of their map that contains demographic data. Eventually we will probably want to also include calculations within this geojson, but that isn't needed at this point. 

## Acceptance criteria

An endpoint exists that will take a project id and will return geojson of the districts with calculated demographic properties.

pcaisse:
## Overview
Export GeoJSON for a given project

### Demo
Viewing the GeoJSON for a project using Delaware data

#### Delaware with all districts unassigned
db_qgis_de_all_unassigned

#### Delaware with 3 districts
db_qgis_de_123_assigned

#### More complex map
db_qgis_de_complex

## Testing Instructions
* Delete all projects since we've added a new non-nullable field to the `project` table
* Run migrations
* Ensure you have a region config set up that corresponds to the hardcoded Delaware topology (`s3_uri` must be 'de')
```
districtbuilder=# select * from region_config where s3_uri = 'de';
-[ RECORD 1 ]+-------------------------------------
id           | 9490b40c-6468-4269-aa63-18cccc2a7d47
name         | Delaware
country_code | US
region_code  | DE
s3_uri       | de
version      | 2020-05-13 21:35:34.223246+00
```
* Create a project through the UI that uses that region config and has 3 districts
* Access the endpoint to produce GeoJSON for that project
```
$ http :3003/api/projects/<project_id>/export/geojson 'Authorization:Bearer $TOKEN' > ~/Downloads/myGeoJSON.json
```
* Check that the GeoJSON looks as you'd expect in QGIS
* Modify the districts via PATCH like so:
```
$ http PATCH :3003/api/projects/<project_id> districtsDefinition:='[1, 2, [1,1,1,2,2,2,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]]' Authorization:Bearer $TOKEN' > ~/Downloads/myOtherGeoJSON.json
```
* Check that the other GeoJSON looks as you'd expect in QGIS

Closes #92


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -55,7 +83,39 @@ export class ProjectsController implements CrudController<Project> {
   ): Promise<Project> {
     return await this.service.createOne(req, {
       ...dto,
-      user: req.parsed.authPersist.userId
+      user: req.parsed.authPersist.userId,
+      districtsDefinition: Array(dto.numberOfDistricts).fill(0)
```
maurizi, line 87: This is incorrect. The length of `districtsDefinition` is not based on `numberOfDistricts` but on the number of counties (or whatever the top-most geo unit is).

pcaisse: :facepalm: Woops. I went ahead and made `districtsDefinition` nullable in ec35055 since it didn't seem very straighforward to figure out how many of the top-most geounits there would be. The frontend should also know enough to produce valid districts definitions and be able to tell what a valid default value should be (an array with appropriate number of zeroes), so it seems okay for it to start as null.

pcaisse:
I pushed several commits to try to close the loop on this functionality, specifically to address the incorrect districts definition but mainly to integrate fully with the topology service.

The biggest change is b223c5c44a62c32a5dcba6916b3f1e4d5e976040 which uses the S3 URL as specified in the region config and gets rid of the hard-coded layers in the topology service. It also includes a change in functionality to go from fetching all region configs up front to fetching them only when needed and then caching them in memory them for reuse (going back to fetching them all when the app loads is easy if we want to go back to that, and perhaps coming up with an altogether different/better caching strategy would be a nice future step, but I wanted to at least get things working respecting the region configs in the database for now).

In any case, there is currently a bigger issue if you go through the full workflow of processing and publishing GeoJSON and then hitting the endpoint as per the testing instructions. The TopoJSON being produced is huge (~160MB, eg. `s3://global-districtbuilder-dev-us-east-1/regions/US/PA/2020-05-22T13:29:40.595Z/topo.json`) whereas the old, hard-coded TopoJSON data that was being used previously by the topology service was about 1MB (eg. `s3://global-districtbuilder-dev-us-east-1/pa/data/bg-lines.topojson`). I don't have enough context to know why it's so much bigger now but it's so big that it takes quite a while to download and more importantly the merge operation doesn't seem to be working for the new TopoJSON:
```
server_1    | [Nest] 416   - 05/23/2020, 1:44:04 AM   [ExceptionsHandler] Cannot read property 'length' of null +376ms
server_1    | TypeError: Cannot read property 'length' of null
server_1    |     at GeoUnitTopology.merge (/home/node/app/server/dist/server/src/districts/entities/geo-unit-topology.entity.js:52:44)
server_1    |     at ProjectsController.exportGeoJSON (/home/node/app/server/dist/server/src/projects/controllers/projects.controller.js:52:39)
server_1    |     at runMicrotasks (<anonymous>)
server_1    |     at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
The TopoJSON looks like TopoJSON and the geolevels hierarchy looks right. I tried reversing the hierarchy but that didn't seem to help. I also tried messing with the simplification option when processing the GeoJSON but I couldn't get the outputted TopoJSON to be any smaller. All of that to say that it _seems_ like my change is mostly working but it can't be fully tested due an issue with GeoJSON processing or there's some other pending incompatibility with `geounit-topology.entity.ts` that needs to be ironed out. It's also possible that I'm making some other mistake, so please let me know if that's the case.

pcaisse:
Okay, this should be working now :sweat_smile: 

I took these steps to make it work for Delaware (the merge operation takes too long for PA and the request times out):

1. `./scripts/manage process-geojson data/DE.geojson -o data/output-de -n 12,4,4 -x 12,12,12 -l block,tract,county`
1. `../scripts/manage publish-region data/output-de US DE Delaware`
1. Create a new project through the UI
1. Update the project's districts definition manually: `districtbuilder=# update project set districts_definition = '[0,0,0]' where id = 'ebbbbf49-0917-43ba-9fd4-8e1f31449833';
UPDATE 1
`
1. Make a request to `http://localhost:3003/api/projects/<project-id>/export/geojson` and you should get GeoJSON back

src/server/src/projects/entities/project.entity.ts
```diff
@@ -19,6 +27,14 @@ export class Project implements IProject {
   @Column({ name: "number_of_districts" })
   numberOfDistricts: number;
 
+  @Column({
+    type: "jsonb",
+    name: "districts_definition",
+    nullable: true
+    // transformer: districtsDefinitionTransformer
```
maurizi, line 34:
```suggestion
```

src/server/src/projects/entities/project.entity.ts
```diff
@@ -1,9 +1,17 @@
...
 import { User } from "../../users/entities/user.entity";
 
+// const districtsDefinitionTransformer = {
+//   from: (json: string) => {
+//     console.log("json", json);
+//     return JSON.parse(json);
+//   },
+//   to: (geounits: GeoUnitCollection) => JSON.stringify(geounits)
+// };
+
```
maurizi, line 14:
```suggestion
```

Issue #83, opened 2020-04-15
Barebones Project page


tgilcrest:
Notes:
* Hit backend API to get project data
* Load static data from S3 
* Load vector tiles from S3
* Show the map

Acceptance Criteria:
* Console print out demographic information when clicking on polygon
* Show only top-most geolevel (e.g. for states this is counties)

PR #139, opened 2020-05-15
Theme-UI componentization pass

### LINKED ISSUE #99: Componentization pass
## Overview

The pages existing at the moment were developed before we had a component library in place, and therefore are bare-bones to the point of not matching up structurally to the wireframes. We need to go through all the existing pages and apply UI components everywhere to begin matching up to wireframes.

## Acceptance criteria

All pages in the application are appropriately making use of the component library to match up structurally to the wireframes.

maurizi:
## Overview

Refactors some of the existing UI into shared components and reworks the markup of every screen to use `theme-ui`

#### Registration page

localhost_3003_projects_5c102dbc-4a92-47fd-a50a-fb8199cf7d24
localhost_3003_projects_5c102dbc-4a92-47fd-a50a-fb8199cf7d24 (3)

#### Create project page

localhost_3003_ (10)
localhost_3003_ (11)
localhost_3003_ (8)
localhost_3003_ (7)
localhost_3003_ (9)

#### Home page

localhost_3003_ (12)

### Notes

 - Not sure I necessarily found the right set of reusable components with the `Field` component(s) - up for other ideas.
 - Purposefully kept a light touch on the Project page, which is under development right now (#83) and the Home page, which wasn't built out enough to style more fully.

## Testing Instructions

- `scripts/update`
- Visit every page. Do all the things. They should work as before, but now look nicer.

Closes #99


src/client/App.css
```diff
@@ -7,31 +7,6 @@ body {
...
-}
-
-.App-link {
-  color: #09d3ac;
-}
-
-.App main {
-  flex: auto;
+button[disabled] {
+  opacity: 0.6;
```
maurizi, line 11: @jfrankl I couldn't find a way using Theme-UI to do this globally - do you know how?

jfrankl: Hey @maurizi, Scott found this issue which seems to present a way: https://github.com/system-ui/theme-ui/issues/892. Let me know if you're able to get this to work. If not, I can take a look later in the sprint.

maurizi: Nice, that worked!

src/client/components/Field.tsx
```diff
@@ -0,0 +1,100 @@
...
+  readonly field: keyof D;
+  readonly resource: WriteResource<D, R>;
+}
+
+export default function Field<D, R>({
+  field,
+  resource,
+  children
+}: {
+  readonly children?: (hasErrors: boolean) => React.ReactNode;
```
pcaisse, line 24: It seems slightly strange to me to have `children` be a function that returns a `ReactNode` instead of it being a `ReactNode` itself (what I would normally expect). Alternatively, we could have the components that wrap `Field` (`InputField` and `SelectField`) call `getFieldErrors` directly (or perhaps a specific `getFieldError` function which takes the resource and a field name) and pass the appropriate markup into `Field` as a `children: React.ReacNode` type.

I don't feel super strongly about that refactor since we won't be consuming `Field` directly, but I think it might make the components in this file a bit easier to make sense of.

tgilcrest:
The reason the divider is in the wireframes, I believe, is because of the Advanced Options section that will go below it. Since we don't have that now I think we should hold off on the dividing line. Rather make that line part of the Advanced Options implementation. @jfrankl ?

src/client/components/Field.tsx
maurizi: I don't think this is that strange, this is straight out of the React docs https://reactjs.org/docs/render-props.html#using-props-other-than-render

pcaisse: Yeah. It also says:
> Since this technique is a little unusual, you’ll probably want to explicitly state that children should be a function in your propTypes when designing an API like this.

towards the bottom. But yeah, I don't feel that strongly about it.

jfrankl:
> The reason the divider is in the wireframes, I believe, is because of the Advanced Options section that will go below it. Since we don't have that now I think we should hold off on the dividing line. Rather make that line part of the Advanced Options implementation. @jfrankl ?

@tgilcrest thanks for catching that. Yes, let's remove the horizontal line until we implement the advanced options.

PR #140, opened 2020-05-15
Add Terraform configuration for staging CLI

### LINKED ISSUE #72: Add Terraform configuration for staging CLI
## Overview

Import the following resources that were created out-of-band into the Terraform state:

- Amazon ECR
- Route 53
- ...

Make use of our existing Terraform modules to add the following new resources:

- ECS with Fargate
  - Pass environment secrets to the ECS task definition
- Include log aggregation with CloudWatch Logs

We should be able to apply migrations once this issue is complete.

rbreslow:
## Overview

The following changes were made to support the staging CLI on AWS:

- Add TypeScript build output to `districtbuilder-manage` container image
- Create new ECR repositories
- ECS with Fargate
  - Pass environment secrets to the ECS task definition
- Include log aggregation with CloudWatch Logs

Resolves #72 

## Testing Instructions

- Verify that the Terraform plan output is clean

```bash
$ docker-compose -f docker-compose.ci.yml run --rm terraform
bash-5.0# ./scripts/infra plan
. . .
```

- Build and publish the container images

```bash
$ ./scripts/cibuild
. . .

$ ./scripts/cipublish
. . .
```

- See the output of the ECS task that I invoked out-of-band:
https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/ecsStagingCluster/tasks/8943a75f-e053-4ea8-9e0a-a04e5cee840c/details

image

### DIFF
--- a/src/manage/.dockerignore
+++ b/src/manage/.dockerignore
@@ -1,0 +1,5 @@
+*
+!bin
+!lib
+!package.json
+!yarn.lock
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -31,7 +31,7 @@ $ export DB_AWS_ECR_ENDPOINT=123456789012.dkr.ecr.us-east-1.amazonaws.com
 $ ./scripts/cibuild
 ...
 Successfully built 20dcf93f6907
-Successfully tagged district-builder-server:a476b78
+Successfully tagged districtbuilder:a476b78
 $ ./scripts/cipublish
 ...
 ```
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -1,0 +1,64 @@
+#
+# Security Group Resources
+#
+resource "aws_security_group" "app" {
+  name   = "sg${var.environment}AppEcsService"
+  vpc_id = module.vpc.id
+
+  tags = {
+    Name        = "sg${var.environment}AppEcsService",
+    Project     = var.project
+    Environment = var.environment
+  }
+}
+
+#
+# ECS Resources
+#
+resource "aws_ecs_cluster" "app" {
+  name = "ecs${var.environment}Cluster"
+
+  setting {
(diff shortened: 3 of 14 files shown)


src/manage/.dockerignore
```diff
@@ -0,0 +1,5 @@
+*
+!bin
+!lib
+!package.json
+!yarn.lock
```
rbreslow, line 5: I had originally tried to do something like

```docker
COPY bin lib* /home/node/app/manage/
```

However, this is problematic because it will try to copy the contents of the `bin` and `lib` (if exists) directories into `/home/node/app/manage/`, where we want to retain the directory structure.

I can't copy them separately because `lib` may not exist and you can only copy a source that may or may not exist with a source that does exist.

```docker
COPY bin /home/node/app/manage/bin
COPY lib /home/node/app/manage/lib
```

The approach I'm taking here is to make it so the only files in the build context are `bin` and `lib` so that I can copy `.` and maintain directory structure.

hectcastro: Gotcha. Wasn't sure where `lib` got created, and in the process of trying to run `cibuild` locally, I ran into a bunch of sluggishness due to the known macOS file sharing issues.

If we can't depend on `lib` being there, I think this approach makes sense. My only concern is that it makes a little trickier to extend later (you need to know to look inside `.dockerignore`).

rbreslow: >Wasn't sure where `lib` got created

`lib` is the output directory of the compiled files from `manage`.

https://github.com/PublicMapping/districtbuilder/blob/9448c630f7d4e53ba5171fa95039591c3ac44ca0/src/manage/tsconfig.json#L9

>My only concern is that it makes a little trickier to extend later

Agreed. Although, I can't think of another way to do this with this project.

PR #141, opened 2020-05-20
Barebones project page

### LINKED ISSUE #83: Barebones Project page

kshepard:
## Overview

As a start for the project page we want to do the following, all of which has been implemented here:
 * Hit backend API to get project data
 * Load static data from S3
 * Load vector tiles from S3
 * Show the map
 * Console print out demographic information when clicking on polygon (for top geolevel only)

### Demo

project-pa

project-de

### Notes
 * It was easy enough to implement, so instead of showing only the topmost geolevel, all of them are being displayed (you'll have to zoom in far to see blocks)
 * I noticed the vector tiles still have demographic data on them. I left it since it's great for testing, but now that we're able to load the static data, we can reduce the tile file size by removing these properties.
 * I had to add a couple additional small pieces of data to `static-metadata.json` -- the bounding box, and an array defining the geolevel hierarchy. I suspect we will need to add a few more pieces of data as we develop this further (e.g. for zoom level boundaries of geolevels), but this was sufficient for the time being. 
 * I needed to create two separate axios instances, since the way it was working, an auth header was being added to all the requests, but we don't want this behavior when requesting things from S3.
 * Since out S3 paths are stored as `s3://` in the database, I needed to add a helper function to turn these into `https://` equivalents. I considered a couple different strategies here, but decided to go with this one for now, mainly because there's another open PR that will be making use of this field, and wanted to minimize friction. We may want to revisit this in the future when we have a better idea of the primary uses of this field will be. 
 *  I added `as Parameters<typeof TheFunctionBeingCalled>` to all of our `Cmd.run` `args`, since it ensures type safety for these calls (it's not checked otherwise, and this has bitten me in the past).
 * So far I'm pretty happy with the memory and performance characteristics of the static data array buffers. :tada:

## Testing Instructions
 * Follow the instructions in #91 to process geojson and then publish it. That PR has instructions for DE. I also tested with PA using the following commands:
    * `./scripts/manage process-geojson data/PA.geojson -b -o data/output-pa -n 12,4,4 -x 12,12,12`
    * `./scripts/manage publish-region data/output-pa US PA Pennsylvania`
 * The previous step is necessary, because some additional metadata has been added to `static-metadata.json`. Any previously published projects will not work correctly in the following steps.
 * Use the UI to create a new project using the region you just published
 * View this project and ensure the vector tiles are showing up correctly
 * Zoom all the way in to see blocks
 * Open up the developer console and click the border of a county. You should see two things logged: the data coming from the vector tile, and the data generated by aggregating the static data files. Verify that these values match up.

Closes #83


src/client/reducers/user.ts
```diff
@@ -27,7 +27,7 @@ const userReducer: LoopReducer<UserState, Action> = (
         Cmd.run(fetchUser, {
           successActionCreator: userFetchSuccess,
           failActionCreator: userFetchFailure,
-          args: []
+          args: [] as Parameters<typeof fetchUser>
```
maurizi, line 30: :fireworks:

src/client/reducers/projectData.ts
```diff
@@ -0,0 +1,181 @@
...
+              ...state,
+              staticMetadata: {
+                resource: action.payload
+              }
+            },
+            Cmd.action(
+              staticGeoLevelsFetch({
+                s3URI: state.project.resource.regionConfig.s3URI,
+                geoLevels: action.payload.geoLevels
+              })
```
maurizi, line 93: I think it's possible here to also request the demographic array buffers at this point too right? Would cut down on wait times a bit to parallelize the downloads if we can.

kshepard: Very nice suggestion. I've implemented it.

PR #142, opened 2020-05-20
Deploy staging application

### LINKED ISSUE #71: Add Terraform configuration for staging application
## Overview

Make use of our existing Terraform modules to add the following new resources:

- ECS with Fargate
- ACM certificate with [terraform-aws-acm-certificate](https://github.com/azavea/terraform-aws-acm-certificate)
- Layer in CloudFront distribution to front ALB

ColeKettler:
## Overview

Configures and deploys the application to Fargate in the staging environment.

### Notes

Migrations have been applied to the staging database.

There are some minor formatting fixes in this PR, you may want to hide whitespace in the diff.

## Testing Instructions

- Run a Terraform plan and verify that infrastructure is up to date:
```#bash
docker-compose -f docker-compose.ci.yml terraform scripts/infra plan
```
- Verify that the application is functional at https://staging.districtbuilder.azavea.com

Closes #71

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-142') {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
       // Publish container images built and tested during `cibuild`
       // to the private Amazon Container Registry tagged with the
       // first seven characters of the revision SHA.
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-142') {
       // Publish container images built and tested during `cibuild`
       // to the private Amazon Container Registry tagged with the
       // first seven characters of the revision SHA.
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -1,6 +1,16 @@
 #
 # Security Group Resources
 #
+resource "aws_security_group" "alb" {
+  vpc_id = module.vpc.id
+
+  tags = {
+    Name        = "sgAppLoadBalancer"
+    Project     = var.project
+    Environment = var.environment
+  }
+}
+
 resource "aws_security_group" "app" {
   name   = "sg${var.environment}AppEcsService"
   vpc_id = module.vpc.id
@@ -9,6 +19,59 @@ resource "aws_security_group" "app" {
     Name        = "sg${var.environment}AppEcsService",
     Project     = var.project
     Environment = var.environment
+  }
+}
+
+#
+# ALB Resources
+#
+resource "aws_lb" "app" {
+  name            = "alb${var.environment}App"
+  security_groups = [aws_security_group.alb.id]
+  subnets         = module.vpc.public_subnet_ids
+
+  tags = {
+    Name        = "alb${var.environment}App"
+    Project     = var.project
+    Environment = var.environment
+  }
+}
+
+resource "aws_lb_target_group" "app" {
+  name = "tg${var.environment}App"
+
+  health_check {
+    healthy_threshold   = "3"
+    interval            = "30"
+    matcher             = "200"
+    protocol            = "HTTP"
+    timeout             = "3"
+    path                = "/health-check/"
+    unhealthy_threshold = "2"
+  }
+
+  port     = "80"
+  protocol = "HTTP"
+  vpc_id   = module.vpc.id
+
+  target_type = "ip"
+
+  tags = {
+    Name        = "tg${var.environment}App"
+    Project     = var.project
+    Environment = var.environment
+  }
+}
+
+resource "aws_lb_listener" "app" {
+  load_balancer_arn = aws_lb.app.id
+  port              = "443"
+  protocol          = "HTTPS"
+  certificate_arn   = module.cert.arn
+
+  default_action {
+    target_group_arn = aws_lb_target_group.app.id
+    type             = "forward"
   }
 }
 
@@ -61,6 +124,33 @@ resource "aws_ecs_task_definition" "app" {
   }
 }
 
+resource "aws_ecs_service" "app" {
+  name            = "${var.environment}App"
+  cluster         = aws_ecs_cluster.app.id
+  task_definition = aws_ecs_task_definition.app.arn
+
+  desired_count                      = var.fargate_app_desired_count
+  deployment_minimum_healthy_percent = var.fargate_app_deployment_min_percent
+  deployment_maximum_percent         = var.fargate_app_deployment_max_percent
+
+  launch_type = "FARGATE"
+
+  network_configuration {
+    security_groups = [aws_security_group.app.id]
+    subnets         = module.vpc.private_subnet_ids
+  }
+
+  load_balancer {
+    target_group_arn = aws_lb_target_group.app.arn
+    container_name   = "app"
+    container_port   = var.app_port
+  }
+
+  depends_on = [
+    "aws_lb_listener.app",
+  ]
+}
+
 resource "aws_ecs_task_definition" "app_cli" {
   family                   = "${var.environment}AppCLI"
   network_mode             = "awsvpc"
@@ -95,6 +185,11 @@ resource "aws_ecs_task_definition" "app_cli" {
 #
 # CloudWatch Resources
 #
+resource "aws_cloudwatch_log_group" "app" {
+  name              = "log${var.environment}App"
+  retention_in_days = 30
+}
+
 resource "aws_cloudwatch_log_group" "app_cli" {
   name              = "log${var.environment}AppCLI"
   retention_in_days = 30
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -105,7 +105,7 @@ resource "aws_ecs_task_definition" "app" {
     postgres_user     = var.rds_database_username
     postgres_password = var.rds_database_password
     postgres_db       = var.rds_database_name
-    
+
     jwt_secret           = var.jwt_secret
     jwt_expiration_in_ms = var.jwt_expiration_in_ms
     client_url           = var.client_url
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -24,6 +24,43 @@ resource "aws_ecs_cluster" "app" {
   }
 }
 
+resource "aws_ecs_task_definition" "app" {
+  family                   = "${var.environment}App"
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = var.fargate_app_cpu
+  memory                   = var.fargate_app_memory
+
+  task_role_arn      = aws_iam_role.ecs_task_role.arn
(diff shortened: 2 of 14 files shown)


Jenkinsfile
```diff
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-142') {
```
ColeKettler, line 22: This will be removed prior to merge.

deployment/terraform/app.tf
```diff
@@ -1,6 +1,16 @@
 #
 # Security Group Resources
 #
+resource "aws_security_group" "alb" {
+  vpc_id = module.vpc.id
+
+  tags = {
+    Name        = "sgAppLoadBalancer"
```
ColeKettler, line 8: Just curious, is there a reason we don't typically namespace this tag by environment? Working off of FieldScope primarily.

deployment/terraform/cdn.tf
```diff
@@ -0,0 +1,85 @@
+resource "aws_cloudfront_distribution" "cdn" {
```
ColeKettler, line 1: Config in here represents a common set of defaults that we use for static files, e.g. in FieldScope or OAR.

deployment/terraform/task-definitions/app.json.tmpl
```diff
@@ -0,0 +1,59 @@
+[
+  {
+    "name": "app",
+    "image": "${image}",
+    "cpu": 0,
+    "essential": true,
+    "entryPoint": ["yarn"],
+    "command": ["run", "start:prod"],
```
ColeKettler, line 8: This ended up being tricky. While trying to apply migrations via command override in the ECS console, I realized the `ENTRYPOINT` for the base `node` image is `node`. This was messing up any calls to other binaries like `yarn`.

I set the `entryPoint` to `yarn` here since that's what we'd use to run the application and any administrative commands. Migrations can now be run by issuing a command override for `run, migration:run`.

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
+  type = number
+}
+
+variable "jwt_secret" {
+  type = string
+}
+
+variable "jwt_expiration_in_ms" {
+  type = number
+}
```
ColeKettler, line 267: Set to one week.

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
+  type    = string
+}
+
+variable "fargate_app_cpu" {
+  type = number
+}
+
+variable "fargate_app_memory" {
+  type = number
+}
```
ColeKettler, line 251: Set to `256` and `1024` respectively, identical to the CLI task config for now.

src/server/Dockerfile
```diff
@@ -14,6 +14,6 @@ COPY yarn.lock /home/node/app/server
 RUN set -ex \
     && yarn install
 
-COPY ormconfig.js dist* /home/node/app/server/
+COPY . /home/node/app/server
```
ColeKettler, line 17: This was the same problem that we ran into in https://github.com/PublicMapping/districtbuilder/pull/140#discussion_r426033834. I went with the same fix here in the interests of consistency.

src/server/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "prebuild": "rimraf dist",
+    "clean": "rimraf dist",
```
ColeKettler, line 6: `prebuild` is run implicitly before `build`, which was wiping out the static bundle built by the `client` container prior to building the `server` container. I realized this once I saw that there was no `index.html` page present in the deployed application.

By renaming this to `clean` and calling it explicitly, this preserves the behavior of cleaning up the `dist` directory while allowing us to call it explicitly at the very beginning of the build process and preserving the output of the `client` bundle.

scripts/update
```diff
@@ -20,6 +20,11 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
         # Ensure container images are current
         docker-compose build --pull
 
+        # Clean frontend static asset bundle
+        docker-compose \
+            run --rm --no-deps server \
+            bash -c "yarn clean"
+
```
ColeKettler, line 27: See https://github.com/PublicMapping/districtbuilder/pull/142/files#r428728193.

ColeKettler:
Realized that this was only passing due to state in the build environment, fixing it now.

src/server/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "prebuild": "rimraf dist",
+    "clean": "rm -rf dist",
```
ColeKettler, line 6: prebuild is run implicitly before build, which was wiping out the static bundle built by the client container prior to building the server container. I realized this once I saw that there was no index.html page present in the deployed application.

By renaming this to clean and calling it explicitly, this preserves the behavior of cleaning up the dist directory while allowing us to call it explicitly at the very beginning of the build process and preserving the output of the client bundle.

I've also changed this to use `rm -rf` instead of the cross-platform `rimraf` module. We're running all of this in Docker containers anyway so we control the runtime environment for the build. It also removes the need to install the server packages prior to `yarn clean` as a separate build step.

deployment/terraform/firewall.tf
```diff
@@ -66,3 +122,23 @@ resource "aws_security_group_rule" "ecs_rds_egress" {
   security_group_id        = aws_security_group.app.id
```
hectcastro, line 122: This change wasn't related to this PR, but since we setup the application in this PR, I wanted to test connecting to RDS from the bastion. The PostgreSQL client tooling wasn't installed, so I installed it and confirmed that connecting to PostgreSQL from the bastion works.

That led to a follow-up question regarding the application's need for PostGIS. I'm not sure we need it yet (or at all), but it would be worth confirming that, and enabling the PostGIS extension on RDS, if necessary.

deployment/terraform/task-definitions/app.json.tmpl
hectcastro: This makes sense. My only concern is that it introduces enough of a discrepancy between local development and AWS that it might be worth opening an issue to adjust the entrypoint upstream in the `Dockerfile` and Docker Compose configurations to align. That way, we're using `yarn` as the entrypoint everywhere without having to override it.

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
- }
+  type = string
+}
 
- variable "fargate_app_cli_cpu" {
-   type = number
- }
+variable "fargate_app_desired_count" {
+  default = "1"
+  type    = string
```
hectcastro, line 232: Since these are expected to be integers, should they be `number` instead of `string`?

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
+}
 
- variable "fargate_app_cli_memory" {
-   type = number
- }
+variable "fargate_app_deployment_min_percent" {
+  default = "100"
+  type    = string
```
hectcastro, line 237: Similar comment regarding `string` vs. `number`.

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
+}
 
- variable "aws_ecs_task_execution_role_policy_arn" {
-   default = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
-   type    = string
- }
+variable "fargate_app_deployment_max_percent" {
+  default = "200"
+  type    = string
```
hectcastro, line 242: Similar comment regarding `string` vs. `number`.

deployment/terraform/variables.tf
hectcastro: That seems to match what's in the Docker Compose configuration, so makes sense.

deployment/terraform/variables.tf
```diff
@@ -216,19 +224,59 @@ variable "rds_cpu_credit_balance_threshold" {
...
+  type = number
+}
+
+variable "client_url" {
+  type = string
+}
+
+variable "app_port" {
+  default = "3005"
+  type    = string
```
hectcastro, line 275: Similar comment regarding `string` vs. `number`.

src/server/package.json
```diff
@@ -50,7 +50,6 @@
     "passport-jwt": "4.0.0",
     "pg": "7.17.1",
     "reflect-metadata": "0.1.13",
-    "rimraf": "3.0.0",
```
hectcastro, line 53: How was this module removed? Asking because I would have expected corresponding changes to `yarn.lock`.

deployment/terraform/certificate.tf
```diff
@@ -0,0 +1,16 @@
+#
+# ACM resources
+#
+module "cert" {
+  source = "github.com/azavea/terraform-aws-acm-certificate?ref=2.0.0"
+
+  providers = {
+    aws.acm_account     = "aws"
```
hectcastro, line 8: Looks like these quotes are not needed in Terraform 0.12 syntax either.

src/server/Dockerfile
rbreslow: Was it that this didn't work, or you wanted to align solutions for consistency? This should have worked, because `ormconfig.js` is an individual file, I observed it copy the directory `dist` rather than the contents of.

src/server/package.json
ColeKettler: It was removed from `package.json` and then I reran the install. Just realized I neglected to commit the updated `yarn.lock` file, pushing that up now.

deployment/terraform/variables.tf
ColeKettler: I was referring to FieldScope config when creating these variables, where they're strings, but on checking the docs more closely Terraform does indeed expect them to be numbers. I'll get the types of all of these in line with the official docs.

deployment/terraform/firewall.tf
ColeKettler: Good foresight, raising that question.

deployment/terraform/task-definitions/app.json.tmpl
ColeKettler: That's a good idea. I don't think that's worth a new issue though. Since we're already using `yarn start:dev` in the `docker-compose.yml` file and relying on the command that's aliased to `yarn start:prod` in the `Dockerfile`, I'm going to change it as part of this PR.

ColeKettler: Scratch that, there's way more surface area for the entrypoint than I had expected. I created #148 to track it.

deployment/terraform/firewall.tf
ColeKettler: I installed the PostGIS extension on RDS per this [guide](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.html#Appendix.PostgreSQL.CommonDBATasks.PostGIS).

src/server/Dockerfile
ColeKettler: @rbreslow Both, it wasn't setting up the directory structure that `yarn start:prod` is expecting. After it got built by Jenkins, I pulled down the image from ECR and saw this:

```
root@9b28a4bdeebe:/home/node/app/server# ls -la
total 1792
drwxr-xr-x   1 root root    4096 May 26 20:30 .
drwxr-xr-x   1 root root    4096 May 26 20:30 ..
drwxr-xr-x 863 root root   28672 May 26 15:01 node_modules
-rw-r--r--   1 root root     550 May 22 22:47 ormconfig.js
-rw-r--r--   1 root root    3062 May 26 14:58 package.json
drwxr-xr-x   4 root root    4096 May 26 20:28 server
drwxr-xr-x   2 root root    4096 May 26 20:28 shared
drwxr-xr-x   2 root root    4096 May 26 20:28 templates
-rw-r--r--   1 root root 1429356 May 26 20:28 tsconfig.build.tsbuildinfo
-rw-r--r--   1 root root  346576 May 26 15:00 yarn.lock
```

The `dist` directory wasn't getting preserved and its contents were all getting dumped into the same level as `ormconfig.js`.

ColeKettler:
@hectcastro Addressed the most recent round of feedback, coming back to you ⚾

PR #151, opened 2020-05-29
Fix converting S3 URIs to HTTP URIs

maurizi:
This was actually generating strings starting like https:/s3-buck-name/...
which worked when hosting from localhost but not when from staging

Switching from path.join which is meant for filepaths to url.resolve which
is designed for URIs fixed the issue

## Overview

Brief description of what this PR does, and why it is needed.

### Demo

### Notes

## Testing Instructions

- How to test this PR
- Prefer bulleted description

Closes #XXX

### DIFF
--- a/src/client/s3.ts
+++ b/src/client/s3.ts
@@ -1,6 +1,5 @@
 import axios from "axios";
-import { join } from "path";
-import { parse } from "url";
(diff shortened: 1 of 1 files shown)


maurizi:
Trivial, merging

PR #153, opened 2020-06-01
Document running RDS migrations using ECS Fargate

### LINKED ISSUE #46: Determine best approach for executing management commands in ECS
Without Django, we cannot make use of `django-ecsmanage` to facilitate the execution of management commands in ECS. This issue is to determine the lowest friction way to invoke Nest.js commands (migrations, ad-hoc tasks) in ECS without going through the AWS console.

I think that it would be good to shoot to encapsulating this logic in a STRTA.

ColeKettler:
## Overview

Documents the process for running migrations as ECS Fargate tasks. Instructions are currently for the `Staging` environment, but a similar approach can be used for the `Production` environment once it exists.

### Notes

I made a fair bit of progress with both shell scripting using AWS CLI commands, and with adapting https://github.com/azavea/django-ecsmanage to run as a standalone script.

The shell script approach quickly ran into a lot of edge cases around parsing quotes, JSON, and token separators. It became clear that I was in the process of writing a parser so I abandoned that approach.

The `django-ecsmanage` approach was more successful but also ran into problems around parsing out arguments to `ecsmanage` vs. passing them along to ECS. I think this is a viable approach, but it required more adaptation than I expected to remove it from Django. By this point the task was running well over estimate and was starting to block development. I think we should revisit a similar approach if we need to run management commands more frequently but it should be estimated for more time.

## Testing Instructions

- Verify documentation by running a task in the DistrictBuilder AWS ECS console.

Closes #46

### DIFF
--- a/deployment/README.md
+++ b/deployment/README.md
@@ -83,7 +83,7 @@ This will attempt to apply the plan assembled in the previous step using Amazon'
 
 ## RDS Migrations
 
-Migrations on RDS should be run through ECS using the app task definition.
+Migrations on RDS should be run through ECS using the [StagingApp](https://console.aws.amazon.com/ecs/home?region=us-east-1#/taskDefinitions/StagingApp) task definition.
 
 To run migrations on the staging environment, run a new task using the following parameters:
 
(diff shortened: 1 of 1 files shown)


ColeKettler:
Yeah, very disappointing. It's totally doable and I think it would be a good investment but I couldn't justify extending this task anymore given the rate of schema changes being introduced right now. Looking forward to revisiting it in the future.

PR #158, opened 2020-06-04
Send email via SES

### LINKED ISSUE #54: Add Terraform configuration necessary to send email via Amazon SES
- [ ] Add the Terraform resources necessary to support DKIM authenticated email sending via Amazon SES.
- [ ] Ensure that the email address necessary to send emails is provided to the application via the environment.
- [ ] Ensure that the SES transport for NestJS is configured appropriately.

ColeKettler:
## Overview

Creates SES verification and DKIM records in AWS and configures the application to use SES for email transport.

### Notes

~I've opened up an AWS support case to have SES taken out of sandbox mode so that we can send registration emails to arbitrary recipients. I'll indicate when the ticket goes through in a comment.~

Our SES sandbox restrictions have been removed. Our new SES email limit is 50,000 per day at a rate of 14 per second.

~The verification link in the registration email will not be valid. I set up #163 to track this bug.~

## Testing Instructions

- Run `docker-compose -f docker-compose.ci.yml run --rm terraform`.
- Run `GIT_COMMIT=d3ad718 ./scripts/infra plan`. Confirm that there are no new changes to be applied.
- Log in to the DistrictBuilder AWS Console.
- Go to https://staging.districtbuilder.azavea.com/register and register a new user.
- Confirm that you receive a verification email from `no-reply@staging.districtbuilder.azavea.com`.
- Start up a local development server with `scripts/setup` and `scripts/server`.
- Go to the local `server` running on `http://localhost:3005`.
- Perform the same registration process.
- Check the `server` console output and check that an email from `no-reply@staging.districtbuilder.azavea.com` was logged.

Closes #54

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-158') {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
       // Publish container images built and tested during `cibuild`
       // to the private Amazon Container Registry tagged with the
       // first seven characters of the revision SHA.
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-158') {
       // Publish container images built and tested during `cibuild`
       // to the private Amazon Container Registry tagged with the
       // first seven characters of the revision SHA.
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -43,7 +43,7 @@ data "aws_iam_policy_document" "scoped_email_sending" {
       test     = "StringEquals"
       variable = "ses:FromAddress"
 
-      values = [var.from_email]
+      values = [var.default_from_email]
     }
   }
 }
--- a/deployment/terraform/iam.tf
+++ b/deployment/terraform/iam.tf
@@ -30,3 +30,26 @@ resource "aws_iam_role_policy_attachment" "ecs_task_execution_role_policy" {
   role       = aws_iam_role.ecs_task_execution_role.name
   policy_arn = var.aws_ecs_task_execution_role_policy_arn
 }
+
+data "aws_iam_policy_document" "scoped_email_sending" {
+  statement {
+    effect = "Allow"
+
+    actions = ["ses:SendRawEmail"]
+
+    resources = ["*"]
+
+    condition {
+      test     = "StringEquals"
+      variable = "ses:FromAddress"
+
+      values = [var.from_email]
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "scoped_email_sending" {
+  name   = "ses${var.environment}ScopedEmailSendingPolicy"
+  role   = aws_iam_role.ecs_task_role.name
+  policy = data.aws_iam_policy_document.scoped_email_sending.json
+}
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -32,8 +32,8 @@
         "value": "${aws_region}"
       },
       {
-        "name": "FROM_EMAIL",
-        "value": "${from_email}"
+        "name": "DEFAULT_FROM_EMAIL",
+        "value": "${default_from_email}"
       },
       {
         "name": "JWT_SECRET",
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -32,6 +32,10 @@
         "value": "${aws_region}"
       },
       {
+        "name": "NODE_ENV",
+        "value": "${node_env}"
+      },
+      {
         "name": "DEFAULT_FROM_EMAIL",
         "value": "${default_from_email}"
       },
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -28,6 +28,10 @@
         "value": "${postgres_db}"
       },
       {
+        "name": "AWS_REGION",
+        "value": "${aws_region}"
+      },
+      {
         "name": "FROM_EMAIL",
         "value": "${from_email}"
       },
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -28,6 +28,10 @@
         "value": "${postgres_db}"
       },
       {
+        "name": "FROM_EMAIL",
+        "value": "${from_email}"
+      },
+      {
         "name": "JWT_SECRET",
         "value": "${jwt_secret}"
       },
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -29,7 +29,7 @@
       },
       {
         "name": "NODE_ENV",
-        "value": "${node_env}"
+        "value": "${environment}"
       },
       {
         "name": "AWS_REGION",
--- a/deployment/terraform/task-definitions/app.json.tmpl
+++ b/deployment/terraform/task-definitions/app.json.tmpl
@@ -28,12 +28,12 @@
         "value": "${postgres_db}"
       },
       {
-        "name": "AWS_REGION",
-        "value": "${aws_region}"
-      },
-      {
         "name": "NODE_ENV",
         "value": "${node_env}"
+      },
+      {
+        "name": "AWS_REGION",
+        "value": "${aws_region}"
       },
       {
         "name": "DEFAULT_FROM_EMAIL",
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -258,11 +258,6 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
-variable "node_env" {
-  default = "staging"
-  type    = string
-}
-
 variable "default_from_email" {
   type = string
 }
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -258,6 +258,10 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
+variable "from_email" {
+  type    = string
+}
+
 variable "jwt_secret" {
   type = string
 }
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -258,6 +258,11 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
+variable "node_env" {
+  default = "staging"
+  type    = string
+}
+
 variable "default_from_email" {
   type = string
 }
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -259,7 +259,7 @@ variable "fargate_app_cli_memory" {
 }
 
 variable "default_from_email" {
-  type    = string
+  type = string
 }
 
 variable "jwt_secret" {
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -258,7 +258,7 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
-variable "from_email" {
+variable "default_from_email" {
   type    = string
 }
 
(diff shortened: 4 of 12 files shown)


Jenkinsfile
```diff
@@ -19,7 +19,7 @@ node {
       }
     }
 
-    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/')) {
+    if (env.BRANCH_NAME == 'develop' || env.BRANCH_NAME.startsWith('test/') || env.BRANCH_NAME.startsWith('release/') || env.BRANCH_NAME.startsWith('hotfix/') || env.BRANCH_NAME == 'PR-158') {
```
ColeKettler, line 22: This will be removed before this PR is merged.

deployment/terraform/iam.tf
```diff
@@ -30,3 +30,26 @@ resource "aws_iam_role_policy_attachment" "ecs_task_execution_role_policy" {
...
+      values = [var.default_from_email]
+    }
+  }
+}
+
+resource "aws_iam_role_policy" "scoped_email_sending" {
+  name   = "ses${var.environment}ScopedEmailSendingPolicy"
+  role   = aws_iam_role.ecs_task_role.name
+  policy = data.aws_iam_policy_document.scoped_email_sending.json
+}
```
ColeKettler, line 55: I saw this more tightly scoped policy being used in FieldScope and though that the extra constraints on the `FromAddress` were a good idea.

deployment/terraform/task-definitions/app.json.tmpl
```diff
@@ -27,6 +27,14 @@
         "name": "POSTGRES_DB",
         "value": "${postgres_db}"
       },
+      {
+        "name": "AWS_REGION",
+        "value": "${aws_region}"
+      },
```
ColeKettler, line 33: The AWS SDK for Node [does not set the region by default](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html), but it will use this environment variable. It also [does not recognize `AWS_DEFAULT_REGION`](https://github.com/aws/aws-sdk-js/pull/2930); you have to use `AWS_REGION`.

docker-compose.yml
```diff
@@ -21,6 +21,8 @@ services:
       - POSTGRES_PASSWORD=districtbuilder
       - POSTGRES_DB=districtbuilder
       - AWS_PROFILE=${AWS_PROFILE:-district-builder}
+      - AWS_REGION=${AWS_REGION:-us-east-1}
+      - DEFAULT_FROM_EMAIL=no-reply@staging.districtbuilder.azavea.com
```
ColeKettler, line 25: This staging address is being used because its domain is verified in SES, allowing us to use SES in local development as well.

src/server/src/app.module.ts
```diff
@@ -19,12 +17,10 @@ import { UsersModule } from "./users/users.module";
 
 import { join } from "path";
 
-// In development the email service is a no-op that only logs
-// TODO: Use SESTransport.Options instead if using SES in #54
-const mailTransportOptions: StreamTransport.Options = {
-  streamTransport: true,
-  buffer: true,
-  newline: "unix"
```
ColeKettler, line 27: I think it could be useful to preserve this local stream behavior that will send emails to `stdout` instead of actually sending them. However this gets into application-level configuration concerns and I think it's better left for application developers to determine if they should and how to condition behavior based on the environment.

I've created #165 to keep track of this in the event that we want to reintroduce this behavior at a later time.

src/server/src/auth/services/auth.service.ts
```diff
@@ -83,7 +83,7 @@ export class AuthService {
       template: "verify"
     });
 
-    this.logger.debug(info.envelope, info.message.toString());
+    this.logger.debug(info.envelope, info.raw.toString());
```
ColeKettler, line 86: The properties available on SES email responses are different from the stream transport. Instead of including a message, the SES email includes a raw buffer for the message body.

src/server/src/common/constants.ts
```diff
@@ -1,3 +1,3 @@
 export const BCRYPT_SALT_ROUNDS = 10;
 export const EMAIL_VERIFICATION_TOKEN_LENGTH = 20;
-export const FROM_EMAIL = "no-reply@districtbuilder.org";
+export const DEFAULT_FROM_EMAIL = process.env.DEFAULT_FROM_EMAIL;
```
ColeKettler, line 3: I changed the name of this constant for sake of consistency with projects like OAR and FieldScope.

deployment/terraform/task-definitions/app.json.tmpl
rbreslow: What happens if the region is not set, for something like SES?

deployment/terraform/variables.tf
```diff
@@ -258,6 +258,10 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
+variable "default_from_email" {
+  type    = string
```
rbreslow, line 262: Nitpick, but you should run `terraform fmt` on this file.

src/server/src/app.module.ts
rbreslow: I think that we should put our foot down here and use the [`stream`](https://nodemailer.com/transports/stream/) transport in local development. 

This the precedent we've set with Django. It allows anyone (including outside collaborators) to test without an AWS profile, and provides a much faster feedback loop for testing.

src/server/src/auth/services/auth.service.ts
rbreslow: Could you share a `JSON.stringify()` representation of the `info` object? I'm curious to see what the difference is between the `raw` object and the `message` object, if any.

ColeKettler:
The SES sandbox restrictions have been lifted and our limit has been increased to 50,000 emails per day (14 per second). We can now send emails to unverified addresses. I'll simplify the testing instructions to account for this 📬

deployment/terraform/task-definitions/app.json.tmpl
ColeKettler: The SDK will immediately throw an error complaining that the region isn't set; it won't try to default to anything:

```
server_1    | [Nest] 66   - 06/09/2020, 8:44:27 PM   [ExceptionsHandler] Missing region in config +86131ms
server_1    | ConfigError: Missing region in config
server_1    |     at Request.VALIDATE_REGION (/home/node/app/server/node_modules/aws-sdk/lib/event_listeners.js:92:45)
server_1    |     at Request.callListeners (/home/node/app/server/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
server_1    |     at callNextListener (/home/node/app/server/node_modules/aws-sdk/lib/sequential_executor.js:96:12)
server_1    |     at /home/node/app/server/node_modules/aws-sdk/lib/event_listeners.js:86:9
server_1    |     at finish (/home/node/app/server/node_modules/aws-sdk/lib/config.js:379:7)
server_1    |     at /home/node/app/server/node_modules/aws-sdk/lib/config.js:397:9
server_1    |     at SharedIniFileCredentials.get (/home/node/app/server/node_modules/aws-sdk/lib/credentials.js:127:7)
server_1    |     at getAsyncCredentials (/home/node/app/server/node_modules/aws-sdk/lib/config.js:391:24)
server_1    |     at Config.getCredentials (/home/node/app/server/node_modules/aws-sdk/lib/config.js:411:9)
server_1    |     at Request.VALIDATE_CREDENTIALS (/home/node/app/server/node_modules/aws-sdk/lib/event_listeners.js:81:26)
```

I had to [rethrow the exception in the auth controller](https://github.com/PublicMapping/districtbuilder/blob/41f9c48d1ef4432488c12da393a0e6bbe374cfc0/src/server/src/auth/controllers/auth.controller.ts#L68-L69) in order to see this error. The original registration error is swallowed and a new one is thrown. There's a comment in there explaining that it's to prevent unintentionally logging passwords. I don't think that's the case given that the passwords are stored hashed, but I left it as-is since we probably don't want to be leaking those either.

deployment/terraform/variables.tf
ColeKettler: Gah, missed that towards the end, thanks for catching that!

src/server/src/auth/services/auth.service.ts
ColeKettler: Certainly! Here's what the SES `info` object looks like with the `raw` object and (truncated) message buffer:

```#json
{
  "envelope": {
    "from": "no-reply@staging.districtbuilder.azavea.com",
    "to": [
      "ckettler@azavea.com"
    ]
  },
  "messageId": "<010001729adfdd3b-4afdf18a-ac2f-4d85-831d-a2c6aec23211-000000@email.amazonses.com>",
  "response": "010001729adfdd3b-4afdf18a-ac2f-4d85-831d-a2c6aec23211-000000",
  "raw": {
    "type": "Buffer",
    "data": [
      67,
      111,
      110
    ]
  }
}
```

... and here's the stream `info` object that uses `message` instead:

```#json
{
  "envelope": {
    "from": "no-reply@staging.districtbuilder.azavea.com",
    "to": [
      "ckettler@azavea.com"
    ]
  },
  "messageId": "<0d71d805-eb46-b18e-3b0a-23be237d6438@staging.districtbuilder.azavea.com>",
  "message": {
    "type": "Buffer",
    "data": [
      67,
      111,
      110
    ]
  }
}
```

Similar in structure but with different property names.

src/server/src/app.module.ts
ColeKettler: That makes sense, I agree that we should keep that precedent.

My one concern here is that there's two possible configs with two possible types and a few logging statements scattered around. Sprinkling conditionals based on something like `NODE_ENV` along with property guards would work but I don't think that's the neatest approach. I think pairing with a GA dev would result in a better solution.

What do you think about me queuing #165 up for next sprint, putting a pair tag on it, estimating it, and resolving it early next sprint with a GA dev?

Also, if you've got a fix in mind I'm definitely interested!

rbreslow: > Sprinkling conditionals based on something like `NODE_ENV` along with property guards would work but I don't think that's the neatest approach.

I would argue that this is what we usually do when configuring things based on environment:

https://github.com/open-apparel-registry/open-apparel-registry/blob/48cb643cb0057190b54daf30a015a2aeef198783/src/django/oar/settings.py#L273

I would consider implementing something like that. It looks like you'll only need one conditional, here. 

Then, after this PR is merged, we could queue up a pairing session to see if there is a neater language construct.

src/server/src/auth/services/auth.service.ts
rbreslow: Got it. Weird that the property name changes but they're both buffers.

src/server/src/app.module.ts
ColeKettler: That approach makes sense. I've pushed up some changes with a quick fix. Hopefully soon we'll be able to set up an appropriate type / wrapper for the mail object in #165 (which I will update the description of). I'd love it if the caller didn't have to worry about the environment or which properties are present 🤞

docker-compose.yml
rbreslow: Can we do something like `DEFAULT_FROM_EMAIL = process.env.DEFAULT_FROM_EMAIL || "no-reply@staging.districtbuilder.azavea.com"`, in the constants file, so we don't need to specify this here?

docker-compose.yml
```diff
@@ -21,6 +21,9 @@ services:
       - POSTGRES_PASSWORD=districtbuilder
       - POSTGRES_DB=districtbuilder
       - AWS_PROFILE=${AWS_PROFILE:-district-builder}
+      - AWS_REGION=${AWS_REGION:-us-east-1}
+      - NODE_ENV=development
```
rbreslow, line 25: Nitpick, but can we put `NODE_ENV` above `AWS_PROFILE`, per [usual](https://github.com/open-apparel-registry/open-apparel-registry/blob/48cb643cb0057190b54daf30a015a2aeef198783/docker-compose.yml#L24-L26)?

deployment/terraform/variables.tf
```diff
@@ -258,6 +258,15 @@ variable "fargate_app_cli_memory" {
   type = number
 }
 
+variable "node_env" {
+  default = "staging"
+  type    = string
+}
```
rbreslow, line 264: Can we remove the `node_env` variable and thread `environment` through to the task definition, like we do [here](https://github.com/open-apparel-registry/open-apparel-registry/blob/48cb643cb0057190b54daf30a015a2aeef198783/deployment/terraform/task-definitions/app.json#L14)?

src/server/src/app.module.ts
```diff
@@ -19,13 +18,21 @@ import { UsersModule } from "./users/users.module";
...
 
+let mailTransportOptions: StreamTransport.Options | SESTransport.Options;
 // In development the email service is a no-op that only logs
-// TODO: Use SESTransport.Options instead if using SES in #54
-const mailTransportOptions: StreamTransport.Options = {
-  streamTransport: true,
-  buffer: true,
-  newline: "unix"
-};
+if (process.env.NODE_ENV === "development") {
```
rbreslow, line 23: I would consider adding something in the constants file like `DEBUG = (process.env.ENVIRONMENT === "Development")`. This approach means there's a single source of truth for what enables debug-type behavior, and is what we'd usually do in a Django-based project.

src/server/src/auth/services/auth.service.ts
```diff
@@ -83,7 +83,8 @@ export class AuthService {
       template: "verify"
     });
 
-    this.logger.debug(info.envelope, info.message.toString());
+    const infoBody = info.raw || info.message;
```
rbreslow, line 86: Nice use of `||`.

ColeKettler:
Awesome, thank you for helping me polish this one!

Regarding the activation link not matching an internal URL, I hadn't thought about that case. Currently that can be configured by setting the `CLIENT_URL` environment variable which is used to generate the emailed URLs. I'm leaning towards handling this sort of thing in application config, e.g. with Terraform on AWS, vs. dynamically for sake of predictability.

If there's a useful case I'm not thinking of though, I'd be happy to set up an issue to track it!

PR #159, opened 2020-06-04
Switch from tslint to eslint

### LINKED ISSUE #128: tslint -> eslint
`tslint` is deprecated.

The recommended upgrade path is to convert the project to `eslint` (see https://github.com/typescript-eslint/tslint-to-eslint-config)

KlaasH:
## Overview

Switches the linting from `tslint` to `eslint`, since the former is deprecated.

I started out with [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config), but found it not very helpful.  My issues with it:
- It produced a very long and inscrutable `.eslintrc.js` file that included a bunch of "actually run TSLint inside ESLint to apply the rules that we didn't know how to translate" items, which seems like it doesn't quite fulfill the purpose of stopping using TSLint.  
- The resulting configuration didn't actually work.  It had all kinds of missing rules, and when I got around those it had hundreds of violations, which doesn't make sense given that it was supposed to be replicating a TSLint config that worked fine.  I don't really know if there was an underlying problem that would have cleared everything up if I'd found it.  I tried for a while to get it working, but ultimately gave up.
- Precisely replicating the TSLint config doesn't actually seem like a useful goal in this case.  The [`tslint-to-eslint-config` FAQ](https://github.com/typescript-eslint/tslint-to-eslint-config/blob/master/docs/FAQs.md#should-i-use-tslint-to-eslint-config) notes that exactly replicating your TSLint rules isn't necessarily the way to go for most projects, and recommends evaluating them and ultimately migrating to the `typescript-eslint` recommendations.

So I bailed on the conversion and started over, putting together an `.eslintrc.js` for the `client` package based on standard recommendations for a Typescript/React/Prettier project.  Then I ran it, turned off rules that had several violations, and fixed some one-off violations.

For the plugins, I kept `import`, `jsdoc`, and `prefer-arrow`, since they didn't cause any trouble.  `tslint-immutable` has no direct port to ESLint.  The successor is [`eslint-plugin-functional`](https://github.com/jonaskello/eslint-plugin-functional), by the same author, but it's not a drop-in replacement.  So I turned on the recommended rules and tried to adjust things so that they'd pass and wouldn't be dropping any rules that were enforced under `tslint`.

For the `server` and `manage` packages, I worked off of the new `client` configuration and adjusted based on the differences between the old `tslint.json` files.

### Demo

image

### Notes

I ran into an issue where running `eslint` for `client` worked fine, but when `react-scripts build` ran it, it was unable to find the `functional` rules (so it crashed when it encountered the `eslint-disable` comments for them).  @maurizi found that [it had been reported as an issue](https://github.com/jonaskello/eslint-plugin-functional/issues/105), but neither of us could figure out what's going on or how to make it work.  Except you can work around it by just doing `// eslint-disable-next-line` without naming a specific rule.  Which means other linting errors on the line wouldn't be caught, but that seems like a fairly small price to pay to get it unblocked.  So I changed those comments and added comments above noting what rule it was supposed to be overriding.

## Testing Instructions

- Run `scripts/update` to install the new packages. It will also run `eslint`, since that's included in the `build` command for the containers.
- Running the `lint` and `fix` yarn commands for each container should also work.
- Introduce linting errors and confirm that they get flagged.
- I don't actually know how to create an auto-fixable linting error. Most of the rules that can be automatically fixed get turned off by `extends: "prettier/@typescript-eslint"`, i.e. they're reserved for Prettier to handle and aren't ESLint's problem.

Closes #128


KlaasH:
Oops, I rebased onto develop and didn't re-run linting. Will fix.

Issue #97, opened 2020-04-28
Project page sidebar


tgilcrest:
## Overview

See the wireframe for the sidebar on the left of the page [here](https://app.abstract.com/projects/9f988240-1dad-11e8-9a40-cba930331af7/branches/master/commits/238ea75510f11bde6dcba5f841bd99aa326bd167/files/DCA38482-FC7F-4F48-BECD-44E07D9F7443/layers/6C5663D9-F7D8-4075-AA3B-41F4532437ED). For now, we just need something simple that will display all the districts with colors, along with population (all districts will be 0 at first, with the unassigned district having the total population count).

## Acceptance criteria

The project page has a sidebar on the left listing districts with colors, populations, and population deviation.

PR #162, opened 2020-06-08
Project demographics export

### LINKED ISSUE #160: Add demographic information to GeoJSON export
Follow-on from #138

maurizi:
## Overview

Adds demographic information support to the backend GeoJSON project districts export.

### Demo

image

### Notes

 - Originally I had been planning on calculating these client-side as part of #97, but the district definition doesn't have enough information in it to query the static data ArrayBuffers on page load (we don't have the feature's ID / index). We do have this in the vector tile layer, so we can update demographic data dynamically without needing to query the server as the user interacts with the map.
 - Snuck a few unrelated fixes in here https://github.com/PublicMapping/districtbuilder/commit/fc6902239243159e49063a1073b82c3d255cab54, https://github.com/PublicMapping/districtbuilder/commit/041772e9141425d3659c9bb4f678f8a4d01a7609

## Testing Instructions

- `./scripts/server`
- Query the projects GeoJSON export endpoint. You should see demographic information included in the output.

Closes #160


maurizi:
Fixed in https://github.com/PublicMapping/districtbuilder/pull/162/commits/9a5e22029fa38ec48d176afe826aae2be355db48

I was treating features with an id of `0` as having no id at all.

maurizi:
Actually... something strange is still going on, even though the values are no longer zero.

Unassigned district values haven't gone down after excluding Philly:
image

Philly population is _way_ too small, as is the state population.

I'm going to take another look at this, I'll re-assign it when its ready.

maurizi:
@kshepard this is ready for another look, numbers look correct now:
image

PR #183, opened 2020-06-22
Add ability to draw districts by selecting single geounits

### LINKED ISSUE #96: Draw a district
## Overview

We need to allow adding geounits to a district via a simple set of drawing tools. For now this will only cover selecting districts by clicking on them (no rectangle, etc. tools).

## Acceptance criteria

 * A district can be selected from the sidebar
 * The top-most geolevel geounits can be clicked on (e.g. counties)
 * Population figures are updated in the sidebar accordingly
 * There is an accept changes button that will save the plan to the database

### LINKED ISSUE #147: Save and load districts to/from database

kshepard:
## Overview

Allows interacting with the top-most geolevel only (as a first step) to select geounits via a single geounit selection tool. The selection can be canceled, or approved, which will save the district definition to the database and re-fetch the geojson to update the map and sidebar.

### Demo

draw-districts

### Notes

A few additional things were done here:
 * Cancel/Approve buttons only show up when there's an active selection
 * Wired up sidebar to allow selecting a district for editing
 * Added some basic district color swatches to the sidebar 
 * Added some css cursor:pointer changes to relevant sections of the sidebar
 * Left-aligned district ids in sidebar
 * Rounded population deviation
 * Added link from project page to homepage
 * Enabled the `downlevelIteration` compiler flag to allow iterating over Sets
 * Initialized an array filled with zeros for an empty set of districts, and changed the district definition type to not be nullable. Aside from it being the technically correct way to refer to an empty set of districts, this is also important, because there's no other way for the client to know how to generate an appropriate district definition, since this information is only available in the topology. Note: we'll probably need additional metadata for the other geolevels when we start interacting with them, since I think we'll encounter a related problem.

I've added several TODOs, and will create corresponding issues for future tasks. Most notably:
 * After saving district edits, the map and sidebar perform a full refresh, which we'll want to eliminate
 * There is no intermediate updating of district population figures as the geounits are being selected. The population changes are only shown after Approve is clicked. We'll want to show these intermediate changes.
 * The `projectData` reducer is getting pretty large. I think it will make sense to refactor some of the actions out to a separate module, but didn't do so in the interest of time.
 * We can use a status area to show various errors, like failing to save a set of districts

## Testing Instructions

- `./scripts/server`
- Make sure you have published a region that's available in your database. If you haven't, there are some instructions in previous PRs
- Create a new project (this is necessary, because old projects had null district definitions, which has been changed here) * The `projectData` reducer is getting pretty large. I think it will make sense to refactor some of the actions out to a separate module, but didn't do so in the interest of time.
- Select some geounits to add to a district and click `Approve`
- Select another district, add some geounits to that one, and click `Approve`
- Select some more geounits, and click `Cancel`

Closes #96
Closes #147


kshepard:
Thanks for the review. Good call on allowing deselection. I see an easy way to make that happen, so I'm going to go ahead and do that quickly before merging.

PR #184, opened 2020-06-23
Create ADR for server-side static data loading

### LINKED ISSUE #176: Static Data Loading Architecture ADR
* Conversation with Ops
* Write up ADR

ColeKettler:
## Overview

Establishes an ADR to provide rationale for our short-term approach for quickly manipulating TopoJSON data server-side.

## Testing Instructions

- Review for editorial and technical accuracy.
- Provide feedback on the decision that was reached in the ADR.

Closes #176


ColeKettler:
That's a really good point. Doubling the memory required to support recent historical data should trigger a discussion about an alternate approach (and likely a superseding ADR).

hectcastro:
> One thing that we'll need to be cautious of is that at some point we may want to be able to support multiple versions of data for each state. For example, we'll be launching all states using 2010 census data, but when 2020 census data becomes available, we'll want to load it, and then there would be two versions of data for each state, which would effectively double the amount of memory needed.

This seems like a good addition to the **Consequences** section.

ColeKettler:
I created #192 to track the implementation of the decision in this ADR.

Issue #194, opened 2020-07-02
Add spinner


pcaisse:
## Detailed Description
Add a nice looking spinner to the frontend to be displayed while awaiting responses from the API.

Should consult with @jfrankl about this

## Context
The prototype is all frontend so it doesn't really have to deal with the network latency issue

PR #195, opened 2020-07-02
Seamlessly rerender the map after editing districts

### LINKED ISSUE #185: Map and sidebar re-render after saving a district edit
## Overview

After a user selects geounits and clicks `Approve`, the new GeoJSON is fetched, which ends up causing a full re-render of both the map and the sidebar. I believe this may have to do with how the React components are wired up, which leads to them being fully reinitialized. 

## Acceptance Criteria

Both the map and sidebar are not fully rebuilt after districts are edited. Instead a seamless transition is observed.

pcaisse:
## Overview

Seamlessly rerender the map after editing districts

### Demo

district_builder_seamless_districts2

### Notes

This PR is kind of a grab bag of changes to achieve the desired effect (how this should work was informed by discussions with @jfrankl )

It includes a change to the loading state (now "Loading..." is shown while waiting for district definition saving and loading new GeoJSON, though that was broken out into a separate `Loading.tsx` component so it can easily be improved later; see https://github.com/PublicMapping/districtbuilder/issues/194)

Part of making the seamless refresh functionality work required a reducer state change to both persist the current GeoJSON and show a loading state during the refresh (see change to `resource.ts`).

I added f562a11 to try to mitigate flicker (see that commit description). It's still slightly flickery and could be improved but I think it's a lot better than showing the white map underneath for a second (much worse flicker). (You may want to revert f562a11 to see the difference.)

## Testing Instructions
Test:
* Adding an unassigned geounit
* Changing the district for an assigned geounit
* Removing a geounit from district
* Test selecting multiple geounits and performing the above operations
* Test for flicker (see f562a11) 

Closes #185

### DIFF
--- a/src/client/components/Loading.tsx
+++ b/src/client/components/Loading.tsx
@@ -1,0 +1,7 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+
+// TODO: Make this into a nice-looking spinner
+const Loading = () => <span>Loading...</span>;
+
+export default Loading;
--- a/src/client/components/Loading.tsx
+++ b/src/client/components/Loading.tsx
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { jsx } from "theme-ui";
 
-// TODO: Make this into a nice-looking spinner
+// TODO (#194): Make this into a nice-looking spinner
 const Loading = () => <span>Loading...</span>;
 
 export default Loading;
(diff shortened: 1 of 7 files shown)


src/client/components/Loading.tsx
```diff
@@ -0,0 +1,7 @@
+/** @jsx jsx */
+import { jsx } from "theme-ui";
+
+// TODO: Make this into a nice-looking spinner
```
kshepard, line 4: Should update this comment to refer to issue #194

kshepard:
This does fix the issue, but now there's a warning in the console: 
```
./src/client/components/Map.tsx
  Line 216:6:  React Hook useEffect has a missing dependency: 'selectedDistrictId'. Either include it or remove the dependency array  react-hooks/exhaustive-deps
```

PR #196, opened 2020-07-07
Display intermediate changes to population in sidebar

### LINKED ISSUE #186: Display intermediate changes to population as geounits are selected
## Overview

We need to display in-progress population changes in the district sidebar as geounits are being selected, as is done in the prototype. The selected district should show increases in population, while any districts that geounits are being pulled from (including the unassigned district) should show decreases in population. These increases and decreases should be color coded accordingly (green/red).

Note: this will involve making use of the district definition, to drill down and aggregate new results. I suspect there will be some non-trivial logic involved to work it out.

## Acceptance Criteria

As districts are edited, the sidebar updates with colored population deviations as is done in the prototype.

### LINKED ISSUE #197: Cancel button no longer clears selected geounits
## Overview

Pressing the cancel button with selected geounits no longer clears them from the map. I believe this was introduced in #195, and I further suspect it has something to do with the call to `map.once("idle"...)`. My guess is that since nothing on the map is changing, the idle call is never triggering, and thus never executing the follow-up logic of clearing the selection.

## To reproduce
 * Select a geounit
 * Press the cancel button (they still show as selected)

## Acceptance criteria

Pressing the cancel button clears selected geounits from the map

kshepard:
## Overview

This displays in-progress population changes in the district sidebar as geounits are being selected. The selected district shows increases in population, while any districts that geounits are being pulled from (including the unassigned district) show decreases in population. These increases and decreases are color coded accordingly (green/red).

### Demo

screenshot_13

### Notes

 * Also fixed a problem with the unassigned district being counted as a normal district when calculating population deviations
 * Also fixed a problem with cancel button not clearing selection
 * This is only currently working for the top-most geolevel, which is the only one we can currently select. I spent a lot of time attempting to get it to work with districts comprised of other geolevels too (via manually POSTing district definitions), but ran into a road block, since we don't currently have enough information on the front-end to determine the full composition of base geounits for a given district. I've added some comments explaining the problem, and had a chat with developers. In order to get this working with multiple geolevels, it looks like we are going to have to generate a slimmed-down version of the hierarchy (with no shapes) and ship it over to the front-end.

## Testing Instructions

- `./scripts/server`
- View a project page
- Interact with the map, selecting various geounits and districts, and verify that the numbers are adjusting accordingly

Closes #186 
Closes #197

### DIFF
--- a/src/client/components/Map.tsx
+++ b/src/client/components/Map.tsx
@@ -210,9 +210,11 @@ const Map = ({
       (selectedDistrictId === 0
         ? removeSelectedFeatures(map)
         : // When adding or changing the district to which a geounit is
-          // assigned, wait until districts GeoJSON is updated before removing
-          // selected state.
-          map.once("idle", () => removeSelectedFeatures(map)));
+        // assigned, wait until districts GeoJSON is updated before removing
+        // selected state.
+        map.isStyleLoaded()
+        ? removeSelectedFeatures(map)
+        : map.once("idle", () => removeSelectedFeatures(map)));
     // We don't want to tigger this effect when `selectedDistrictId` changes
     // eslint-disable-next-line
   }, [map, selectedGeounitIds, topGeoLevel]);
--- a/src/client/components/Map.tsx
+++ b/src/client/components/Map.tsx
@@ -212,7 +212,7 @@ const Map = ({
         : // When adding or changing the district to which a geounit is
         // assigned, wait until districts GeoJSON is updated before removing
         // selected state.
-        map.isStyleLoaded()
+        map.isStyleLoaded() && map.isSourceLoaded("districts")
         ? removeSelectedFeatures(map)
         : map.once("idle", () => removeSelectedFeatures(map)));
     // We don't want to tigger this effect when `selectedDistrictId` changes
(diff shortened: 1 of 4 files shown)


src/client/components/Map.tsx
```diff
@@ -210,9 +210,11 @@ const Map = ({
       (selectedDistrictId === 0
         ? removeSelectedFeatures(map)
         : // When adding or changing the district to which a geounit is
-          // assigned, wait until districts GeoJSON is updated before removing
-          // selected state.
-          map.once("idle", () => removeSelectedFeatures(map)));
+        // assigned, wait until districts GeoJSON is updated before removing
+        // selected state.
+        map.isStyleLoaded()
```
pcaisse, line 215: This might not be necessary, but there's also: https://docs.mapbox.com/mapbox-gl-js/api/map/#map#issourceloaded

We could potentially (also?) check for `map.isSourceLoaded("districts")`. Might be worth checking into case you haven't already.

kshepard: Thanks, I'm testing it out, and it seems to be functioning the same way, though is maybe a little safer.

Issue #200, opened 2020-07-07
Implement geolevel selector 


kshepard:
## Overview

Note: this issue supersedes #113 (looking that over may help provide additional context).

This issue is to create a geolevel selector component (i.e. to choose what is being drawn on the map -- counties/block groups/ blocks). When the user selects a geolevel from this component, the drawing tools will operate on that geolevel, and they will be able to save districts containing those geounits. One important thing to note is that constructing the new district definition will involve new logic. As it currently stands, we only allow operating on counties, which makes constructing the new district definition trivial. This logic will need to be expanded.

## Acceptance criteria

 * There is a geolevel selector in place
 * Users are able to select geounits from the selected geolevel
 * Districts with these geounits are able to be saved to the database

PR #203, opened 2020-07-09
Feature/mvm/map demographic labels

### LINKED ISSUE #189: Display vector tile demographic counts on the map
## Overview

#181 adds demographic counts to the vector tiles. These need to be displayed on the map, as is done in the prototype. In order to handle this, we'll also need to add a dropdown selector for which demographic labels to show. See wireframes:

demographic-labels-selector

## Acceptance criteria

A demographic label dropdown selector exists, and interacting with it causes appropriate counts to be displayed on the map.

maurizi:
## Overview

This PR adds a control to the project page to display population / other demographic data as labels on the map.

### Demo

localhost_3003_projects_31b83fa2-4fcd-4bbb-954f-275bb2059af1

### Notes

 - This is hardcoded to only display the county labels - this should be changed after or as part of #200 to show the current geolevel
 - I had to make a small change to the `process-geojson` script to prevent labels from being dropped at low zoom levels

## Testing Instructions

- Rerun `process-geojson` & `publish-region` and create a new project for that region. You should be able to use the label control to display labels on the map

Closes #189


src/client/screens/ProjectScreen.tsx
```diff
@@ -36,14 +37,41 @@ const MapContainer = ({ children }: { readonly children: React.ReactNode }) => {
   return <Flex sx={{ flexDirection: "column", flex: 1 }}>{children}</Flex>;
 };
 
-const MapHeader = () => {
-  return <Box sx={{ variant: "header.app", backgroundColor: "burlywood" }}>MapHeader</Box>;
+const MapHeader = ({
```
kshepard, line 40: Consider moving this to another file, since it's no longer a simple one-liner component.

maurizi: I'm going to leave this as-is to make things simpler for @pcaisse to rebase on top of in #206

PR #208, opened 2020-07-15
Fix rectangle selection tool bug with multiple selections

pcaisse:
## Overview

Small bug fix that @kshepard noticed related to the newly-implemented rectangle selection tool when making multiple, separate selections.

This fix should make the rectangle selection tool work the same as the prototype.

Bug fix is in 63a5616

### Demo

**Before**
db_rectangle_selection_bug_before

**After**
db_rectangle_selection_bug_after

## Testing Instructions
* Make a selection using the rectangle tool, making sure to select at least one feature (mousing over the not yet selected feature should deselect it)
* Make a separate selection, mousing over and off the originally selected feature and newly selected features (original should stay selected, new features should be selected/unselected as per prototype)

### DIFF
--- a/src/client/components/map/RectangleSelectionTool.ts
+++ b/src/client/components/map/RectangleSelectionTool.ts
@@ -54,7 +54,7 @@ const RectangleSelectionTool: ISelectionTool = {
       document.addEventListener("mouseup", onMouseUp);
 
       setOfInitiallySelectedFeatures = featuresToSet(
-        getFeaturesInBoundingBox().filter(feature => isFeatureSelected(feature) === true)
+        getFeaturesInBoundingBox().filter(feature => isFeatureSelected(feature))
       );
 
       // Capture the first xy coordinates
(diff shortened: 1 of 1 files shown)


src/client/components/map/RectangleSelectionTool.ts
```diff
@@ -51,13 +53,17 @@ const RectangleSelectionTool: ISelectionTool = {
       document.addEventListener("mousemove", onMouseMove);
       document.addEventListener("mouseup", onMouseUp);
 
+      setOfInitiallySelectedFeatures = featuresToSet(
+        getFeaturesInBoundingBox().filter(feature => isFeatureSelected(feature) === true)
```
kshepard, line 57: Possibly a style preference that I never noticed you had before, but `isFeatureSelected` returns a boolean, so the ` === true` is redundant.

pcaisse: Yeah, I don't really have a good reason for being explicit about the condition here :shrug:

PR #222, opened 2020-07-23
Add zoom geolevel restrictions and auto-saving when switching geolevels

### LINKED ISSUE #201: Implement zoom/geolevel restrictions
## Overview

Note: this issue supersedes #114  (looking that over may help provide additional context).

There are two ways in which we want to restrict zoom and geolevels:
 * The geolevel selector (#200) is grayed out when the geolevel in question is not visible on the map
 * When geounits are selected, the map zoom is restricted so it can't be zoomed out to a point where that geolevel isn't visible

The information needed for this will be made available in the implementation of #199

In addition, when geounits are selected, and a different geolevel is chosen, changes need to be automatically accepted. See: https://github.com/PublicMapping/districtbuilder/issues/113#issuecomment-649072679

## Acceptance criteria
 * Geolevel selections are greyed out when the geolevel isn't visible
 * Map zoom is restricted when geounits have been selected, so a user isn't able to zoom out past the point where that geolevel is no longer visible
 * If a user has geounits selected and chooses a different geolevel, changes are automatically accepted

pcaisse:
## Overview
Add zoom geolevel restrictions and auto-saving when switching geolevels

### Demo
db_selector_constraints

### Notes
There is a small amount of weirdness introduced in `reducers/districtDrawing.ts` where we need state held in `reducers/projectData.ts` to call the API function to save the updated district definition, so instead we kick off an action, passing the missing data along to the reducer that has the rest of the data to make the API call (previously we were getting around this by having the necessary data where the action is dispatched at the call site but now it needs to be called in two places to accommodate the auto-saving). I suppose this is the price we pay for encapsulation, but it does add complexity. The other alternative is to recombine the two reducers into one.

FWIW I still think the auto-saving is confusing and would rather just not let the user switch geolevels with some messaging if they have a selection instead of auto-saving for them.

## Testing Instructions
* Make sure you can only select the block level after zooming in enough to see blocks
* Make sure once you select a block, you can't zoom out so far as to no longer be able to see blocks
* Switch geolevels and make sure your select was saved

Closes #201

### DIFF
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -16,6 +16,6 @@ export const setSelectionTool = createAction("Set selection tool")<SelectionTool
 
 export const setGeoLevelIndex = createAction("Set geoLevel index")<number>();
 
-export const setBaseGeoUnitVisible = createAction("Set base geounit visible")<boolean>();
+export const setGeoLevelVisibility = createAction("Set geolevel visibility")<readonly boolean[]>();
 
 export const saveDistrictsDefinition = createAction("Save districts definition")();
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -1,5 +1,5 @@
 import { createAction } from "typesafe-actions";
-import { GeoUnits, GeoUnitHierarchy, IProject } from "../../shared/entities";
+import { GeoUnits } from "../../shared/entities";
 
 export enum SelectionTool {
   Default = "DEFAULT",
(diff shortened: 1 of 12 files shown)


src/client/actions/districtDrawing.ts
```diff
@@ -12,18 +12,10 @@ export const addSelectedGeounitIds = createAction("Add selected geounit ids")<Ge
...
->();
-export const patchDistrictsDefinitionFailure = createAction("Patch districts definition failure")<
-  string
->();
-
 export const setSelectionTool = createAction("Set selection tool")<SelectionTool>();
 
 export const setGeoLevelIndex = createAction("Set geoLevel index")<number>();
+
+export const setBaseGeoUnitVisible = createAction("Set base geounit visible")<boolean>();
```
kshepard, line 19: I think this is fine for now, but it has an assumption baked in that we need to keep in mind, and I suspect we'll want to modify down the road. The assumption being that the base geolevel is the only one that gets the special treatment of potentially being hidden. This won't work as-is if we have multiple geolevels that aren't able to be displayed when fully zoomed-out. Just wanted to make a note of that, but I think it's alright to keep it this way for now in the interest of time, since that assumption will likely hold for quite a while at least.

pcaisse:
I went ahead and made some changes to make the restricting of geolevels/zoom dynamic based on the static metadata since it was fresh in my mind and wasn't much more work. I think the behavior is actually a bit more correct now since it respects the min _and_ max zooms for each geolevel and derives the min/max map zooms from the geolevels (this seems to also be something we want to respect from the data). I also went ahead and added a lookup table for the button/tooltip geolevel labels so they are more in keeping with the wireframes and made the labels dynamic.

I tested everything out and it seems to work well but it'd probably be good for you to review and test again as well.

src/client/components/MapHeader.tsx
```diff
@@ -6,18 +6,35 @@ import Icon from "./Icon";
...
+    case "block":
+      return "Blocks";
+    case "tract":
+      return "Tracts";
+    case "blockgroup":
+      return "Blockgroups";
+    case "county":
+      return "Counties";
+    default:
+      return "";
```
kshepard, line 20: I wonder if a better default would simply be returning the geolevel id. Maybe with a simple heuristic like capitalizing the first letter and adding an `s`. I just think it'd be more confusing if nothing is displayed.

PR #234, opened 2020-07-29
Calculate Polsby Popper compactness for drawn districts

### LINKED ISSUE #118: Compactness/Contiguity metric
Something we need DistrictBuilder to measure is compactness. This will go in the sidebar along with the other metrics. We want to calculate the [Polsby-Popper compactness](https://fisherzachary.github.io/public/r-output.html#polsby-popper) of the district and display it in a column like the prototype. 

AC:
1. There is a Compactness column added to the side bar
2. When a district is drawn it calculates the Polsby-Popper compactness of the district and a tooltip that says "Polsby-Popper score"
3. If a district in not contiguous it replaces the compactness metric with an `x` and a tooltip that says "Non-contiguous" (but not `(invalid)` like in the prototype)

When a contiguous district is drawn:

When a non-contiguous district is drawn:

pcaisse:
## Overview
Calculate Polsby Popper compactness for drawn districts

### Demo
db_polsby_popper

### Notes

I've never used Turf before but it's recommended that you install each package separately, which I did. Each package comes with its own TS types.

Also fixed a small alignment bug with the label dropdown in Chrome

I added a little `assertNever` function that makes it easier to do exhaustive checking. In this case it's used to ensure we return something for each possible value that `CompactnessScore` could take on

## Testing Instructions
* Draw some districts and verify that the more compact ones generally show a higher percentage
* Draw a non-contiguous district and ensure that the 'X' icon is shown with the appropriate tooltip

Closes #118


src/server/src/districts/entities/geo-unit-topology.entity.ts
```diff
@@ -144,9 +180,19 @@ export class GeoUnitTopology {
...
+    return {
+      ...featureCollection,
+      features: featureCollection.features.map(feature => ({
+        ...feature,
+        properties: {
+          ...feature.properties,
+          compactness: calcPolsbyPopper(feature)
+        }
+      }))
+    };
```
maurizi, line 196: Consider instead of modifying the feature collection to add compactness here, assigning compactness together with demographics up on line 180:
```
mutableGeom.properties = {
   ...getDemographics(baseIndices, this.staticMetadata, this.demographics),
   compactness: calcPolsbyPopper(feature)
};
```

pcaisse: So that was my initial approach, but I don't _think_ we actually have features to work with at that point (`topojson.mergeArcs` returns a TopoJSON MultiPolygon). It seems like we need to work with the output of `topojson.feature` which is a `FeatureCollection`. Am I missing something?

maurizi: Ah woops, carry on then...

PR #244, opened 2020-07-31
Use humanized numbers for demographic labels

### LINKED ISSUE #190: Use human readable population counts in vector tile labels
## Overview

In #181, population count labels are added to the vector tiles, but they aren't using the abbreviated values that are used in the prototype (see: `https://github.com/azavea/districtbuilder-2/blob/486e50d974cc61d77b939b765ad1ed4de454c8b6/public/process/process.sh#L12`). I believe we will need to run the `abbreviatePopulationValues` script to accomplish this.

## Acceptance Criteria

Population counts in vector tile labels are abbreviated to a human readable format.

maurizi:
## Overview

Ports over the prototype code to abbreviate demographic labels.

### Demo

image

## Testing Instructions

- Rebuild & republish data for a state, then create a new project for that state
- Labels should be abbreviated

Closes #190

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -168,7 +168,7 @@ const Map = ({
       map.setLayoutProperty(
         `${selectedGeolevel.id}-label`,
         "text-field",
-        label ? `{${label}}` : ""
+        label ? `{${label}-abbrev}` : ""
       );
     }
   }, [map, label, staticMetadata, selectedGeolevel]);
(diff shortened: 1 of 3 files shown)


tgilcrest:
In the screenshot is there a label of `0.1m` for the county but 500k people in the district it's assigned to?

maurizi:
> Looks good, tested in both DE and PA. I did notice one kind of weird thing: for PA, the labels for blockgroups don't seem to show up (counties and blocks are fine though). It looks like this was the case before this change as well though. Not sure if you have any insight into that one.

No clue, unfortunately

PR #258, opened 2020-08-03
Update human readable label processing

tgilcrest:
## Overview

I'm suggesting a couple of changes to the way human-readable labels are generated from the prototype.

1. Change the way we determine the suffix to use for the label. The existing method was using `.length` to determine which label to use. This works for integers, but breaks down for decimal numbers. We don't expect to have decimals for population, but this covers us in case they show up unexpectedly, and makes this more robust in case we do. I instead use `log10` to more reliably determine the 000s bin.
2. Change the way we show hundreds of units, e.g. from 0.1m to 100k. The existing approach displays hundreds of thousands as tenths of millions. I've heard, and agree with, feedback that doing so makes it hard to compare between geounits, e.g. 90k vs. 0.1m is a 10k difference but the different unit makes it hard to recognize that. It's also subject to rounding that can make a significant difference. 0.1m can be 100k or 150k, either a 10k or 60k difference from 90k. 

### Demo
I didn't load up the app but this is what they look like in the prototype.

image

## Testing Instructions

- Rebuild & republish data for a state, then create a new project for that state
- Labels should be abbreviated

### DIFF
--- a/src/manage/src/commands/process-geojson.ts
+++ b/src/manage/src/commands/process-geojson.ts
@@ -228,9 +228,7 @@ it when necessary (file sizes ~1GB+).
 
         // Aggregate all desired demographic data
         for (const demo of demographics) {
-          const value = this.aggProperty(geoms, demo);
-          merged.properties[demo] = value;
-          merged.properties[`${demo}-abbrev`] = abbreviateNumber(value);
+          merged.properties[demo] = this.aggProperty(geoms, demo);
         }
 
         // Set the geolevel keys for this level, as well as all larger levels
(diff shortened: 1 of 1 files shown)


jfrankl:
Note: I didn't actually run this script, I am just giving feedback on the overall concept and the `abbreviateNumber`. I'm hoping someone else who is more familiar with the script could test it.

tgilcrest:
Thanks @jfrankl. I overlooked that in testing.

tgilcrest:
I added one more little thing to shorten possible decimal single digits to one digit. Not sure if this is how we want to handle everything, but figured it was better to be safe and shorten it to one digit than possibly have a big long decimal show up in our labels.

kshepard:
Testing this out now in DE and PA. It's looking good. I noticed one problem (unrelated to this PR) with abbreviations not being added at the base geounit level. I'll make that change and will also cleanup the formatting that CI is complaining about.

PR #261, opened 2020-08-03
Lock districts for editing

### LINKED ISSUE #112: Lock district for editing
AC:
* There is a lock button that shows up when you hover over the district
* Locking a district means that geounits assigned to it can't be selected when drawing. 
  * So that you can't remove geounits from the district (take them from the district and assign them to another district or unassigned). 
  * You can still add geounits to the district if it's locked
* Locking creates a pattern overlay on the district to indicate it's locked (see screenshot)

pcaisse:
## Overview
Lock districts for editing

### Demo
db_lock_districts

## Notes
This copies over the sprites from the prototype for the dotted pattern overlay for locked districts.

This deviates from the prototype in that selecting a partially locked geometry will not select anything instead of selecting only the unlocked sub-geounits. Making it work like the prototype would be much more involved since we'd need to query for multiple geolevel layers to get all the constituent features to enable selecting sub-geounits of partially locked geometries. 

## Testing Instructions
* Lock and unlock a district with some geometries assigned to it to check the dotted overlay is shown/hidden correctly
* Make sure white lock is shown for a row when mousing over and then a filled in lock is shown when a district is locked
* Test that geometries inside of a locked district cannot be selected with any selection tool
* Test that partially locked geometries cannot be selected (eg. select county with locked blockgroups/tracts)

Closes #112


tgilcrest:
>This deviates from the prototype in that selecting a partially locked geometry will not select anything instead of selecting only the unlocked sub-geounits. 

So if you have a locked district partially covering a county, you can't select the whole county but can you change to blockgroups and select all of the blockgroups of that county?

pcaisse:
> So if you have a locked district partially covering a county, you can't select the whole county but can you change to blockgroups and select all of the blockgroups of that county?

Yes, all of the unlocked blockgroups

pcaisse:
> I haven't taken a look at the code yet, but took a quick spin of the app and have a couple changes to request:

Nice job testing. I fixed all the issues you called out.

Issue #241, opened 2020-07-30
Only show lines for selected geolevel


tgilcrest:
Currently we show lines for every geolevel on the map (subject to allowed zoom level). This can be confusing when drawing because you expect to draw with the shapes you can see. However, the shapes you see aren't always the same as the geolevel you have selected, so it's easy to think you will select blockgroups when you actually have counties selected. Because of this, we should only show lines for the selected geolevel on the map (edit: and higher geolevels; so, if you have blockgroups selected, you should see blockgroups and counties).

jfrankl:
Slight change on this @tgilcrest. We should show lines for the selected geolevel and any level above it. So, if you have blockgroups selected, you should see blockgroups and counties. If you have blocks selected, you should see blocks, blockgroups, and counties.

We can use different styling on each to make it more clear which is which when multiple are on the screen.

PR #270, opened 2020-08-04
Auto-select next highest geolevel if current geolevel is not visible

### LINKED ISSUE #245: Improve UX for zooming with restricted geolevels
## Description

We prevent selecting the blocks geolevel at zoom levels below a certain threshold. However, if you zoom in, select the blocks geolevel, and then zoom out, blocks are still selected but they are grayed out and we don't clearly explain what changed and what a user needs to do to continue working with blocks. The user can't select geounits on the map until they manually select an available geolevel. For example:

To remedy this, we want to make it more clear that the user needs to zoom back in to see and interact with blocks by showing a them a message (see screenshot). We also want to not gray-out the blocks geolevel so it's clear the blocks geolevel is still active.

AC:
1. When a user selects a restricted geolevel (i.e. Blocks) and zooms out past the min-zoom, we show a message that explains: `Zoom in to work with [geolevel]`. (See screenshot)
2. Also when this happens, the blocks geolevel button is still active and indicates blocks are still selected (rather than being grayed out).
3. When zoomed out, if a user selects a non-restricted geolevel  the restricted geolevel button is deactivated and returns to a state like in #268 / #287. 
4. If a user has a restricted geounit selected, the user can zoom out past the min-zoom (rather than be limited to it) and the selection is maintained (i.e. sidebar shows proposed change, if you zoom back in the selection is still there)

## Screenshots

image

pcaisse:
## Overview
Make sure a visible geolevel is selected by auto-selecting the next highest geolevel if the current geolevel is not visible

### Demo
db_auto_select_geolevel

## Testing Instructions
* Testing manually switching between geolevels works as before
* Test that auto-accepting works as before
* Test that if you zoom to the block level and select block level, if you then zoom out such that blocks are hidden, the next highest geolevel is selected automatically 

Closes #245

### DIFF
--- a/src/client/reducers/districtDrawing.ts
+++ b/src/client/reducers/districtDrawing.ts
@@ -70,14 +70,26 @@ const districtDrawingReducer: LoopReducer<DistrictDrawingState, Action> = (
         ...state,
         selectionTool: action.payload
       };
-    case getType(setGeoLevelIndex):
+    case getType(setGeoLevelIndex): {
+      // Make sure a visible geolevel is selected by auto-selecting the next highest geolevel if the
+      // current geolevel is not visible
+      const newGeoLevelIndex = [
+        ...state.geoLevelVisibility.slice(0, action.payload + 1).keys()
+      ].reduce(
+        (finalIndex, currentIndex) =>
+          state.geoLevelVisibility[currentIndex] ? currentIndex : finalIndex,
+        action.payload
+      );
       return loop(
         {
           ...state,
-          geoLevelIndex: action.payload
+          geoLevelIndex: newGeoLevelIndex
         },
-        state.selectedGeounits.size > 0 ? Cmd.action(saveDistrictsDefinition()) : Cmd.none
+        state.selectedGeounits.size > 0 && state.geoLevelIndex !== newGeoLevelIndex
+          ? Cmd.action(saveDistrictsDefinition())
+          : Cmd.none
       );
+    }
     case getType(saveDistrictsDefinition):
       return loop(
         state,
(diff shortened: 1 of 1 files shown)


jfrankl:
@pcaisse thanks for working on this. We've been rethinking how we handle the geounit switcher, and we made a decision that affects this PR.

Instead of having the application change geounits when the user zooms out, we are hoping to display a small message that explains that a user needs to zoom in to see the no-longer-visible geounits. We think that this will clearly communicate why the geounits are no longer visible and what the user needs to do to see them again. It also avoids an issue with automatically changing geounits, which is when a user accidentally zooms out too far.

image

I'm sorry we were not able to get this change reflected in the issue before you worked on it and for the annoyance of asking you to redo work. I am hoping that you will be able to use most of this code to implement the logic behind this message.

cc @tgilcrest

pcaisse:
@jfrankl So if you zoom in to the block level and select 'Blocks' and then back out past where blocks are shown, you would want to keep the block level selected -- i.e. the user can't make a selection -- and show the message for them to zoom in again?

My 2 cents is that works well for the case where they accidentally zoomed too far out, but not for the case where they zoomed out because they didn't want to work with blocks anymore.

It makes me wonder about what fundamentally the interaction with the geolevel selector is for the user. Is the user is expected to be actively changing that all the time as they're building out their districts or should they mostly be interacting with the map and not have to think about selecting geolevels all the time? As a user, part of my confusion around this centers around seeing multiple geolevel geometries at once and being unsure as to which level I'm on.

jfrankl:
> @jfrankl So if you zoom in to the block level and select 'Blocks' and then back out past where blocks are shown, you would want to keep the block level selected -- i.e. the user can't make a selection -- and show the message for them to zoom in again?

Yes, that is correct.

> My 2 cents is that works well for the case where they accidentally zoomed too far out, but not for the case where they zoomed out because they didn't want to work with blocks anymore.

It's true that the direction I described does not automate switching geounits on zoom, which is intentional. The decision about what geounits you are selecting is not always related to your zoom level. We should allow users to move freely around the map without making assumptions about the geounits and changing it automatically. The old version of DB worked in the way that you described, and we received feedback that this was annoying and did not provide flexibility around how users wanted to draw. The direction we're going in intentionally sacrifices some time/automation for control/flexibility.

> As a user, part of my confusion around this centers around seeing multiple geolevel geometries at once and being unsure as to which level I'm on.

I believe this will be more clear when we implement #241, but if that's not the case, we can make adjustments from there.

pcaisse:
Gotcha, makes sense. I think in that case we can just close this out as well as the accompanying issue since it doesn't apply and write up a new one.

PR #277, opened 2020-08-06
Map tooltip

### LINKED ISSUE #218: Drawing cursor tooltip
When selecting geounits, we want to display a tooltip alongside the cursor that shows some additional details that are helpful for drawing. This provides some demographic information about the blockgroups that the user will select for drawing. It changes based on the cursor used.

AC:
1. For the single-select cursor, it displays the geounit, geounit ID, total population, and graph of the demographic information of the geounit that is being hovered
2. For the rectangle select, it displays the same on hover, but then displays a count of the geounits selected, total population, and updated graph of demographic information when multiple geounits are selected
3. When the cursor moves to the right-hand or bottom edge of the screen, the tooltip moves to another side of the cursor to stay visible (e.g. if cursor moves to right end of screen, tooltip moves to left side of cursor. If cursor moves to bottom edge of screen, tooltip moves above cursor. Bottom right corner of screen tooltip moves to top left side of cursor.)

Single-select popup:

Rectangle select:

Rectangle select with multiple selected:

maurizi:
## Overview

Adds a tooltip to the map page when hovering over features that displays demographic information.

### Demo

Screenshot from 2020-08-06 08-51-42
image

### Notes

In consulation with @jfrankl & @tgilcrest I made a few changes from the way the prototype works / the wireframes on Abstract:
 - When selecting with the rectangle tool, we do not show e.g. "2 blockgroups" as a header as in the wireframes, only population
 - The total demographics shown when using the rectangle tool only represents _new_ geounits that will be added by the currently drawn bounding box, not those contained in the bounding box which were already selected, which is a change from how the prototype functioned

You'll note that there is code for displaying a name instead of an ID if the feature has a `name` property, which we'll be able to support once we have new data with county names available (https://github.com/azavea/districtbuilder-data-pipeline/issues/9)

Additionally, I'm showing percentages for demographics, while the wireframes only show the bar chart, because it enabled me to reuse the tooltip component from the sidebar.

## Testing Instructions

- `./scripts/update`
- `./scripts/server`

Closes #218


tgilcrest:
This looks great, Mike. I think when we meet @jfrankl said that we should replace "2 blockgroups" in the prototype with "Multiple blockgroups" so the height of the tooltip and the position of the graph don't change.

PR #286, opened 2020-08-10
Show dropdown menu when user is logged in with an option to logout

### LINKED ISSUE #248: Show avatar instead of login/register buttons when user is logged in
## Description
We should show the avatar dropdown instead of login/register buttons when user is logged in. The dropdown, at first, will probably only have an option to "Logout". Later, it will have a way to change your account settings, etc.

## Steps To Reproduce

1. Go to https://staging.districtbuilder.azavea.com/
2. Login

## Expected behavior
You should see an avatar dropdown, as is pictured in this wireframe, instead of Login / Register
image

## Actual behavior
There is a login and register button. This is confusing and initially made me think that I was not logged in.

maurizi:
## Overview

Switches out the login / register buttons for a dropdown menu with an option to logout when the user is logged in.
Eventually we intend for this dropdown menu to house other things, but for now it only has logout.

### Demo

Peek 2020-08-10 12-06

### Notes

 - Theme-UI doesn't have any components suitable for use as a dropdown menu, so I decided to use `rc-menu` for the dropdown menu because I've used it before and it has TS support.

## Testing Instructions

- `scripts/update`
- Go to the home screen both logged out & logged in Use the option in the dropdown menu to logout.

Closes #248


src/client/screens/HomeScreen.tsx
```diff
@@ -1,35 +1,118 @@
...
   readonly projects: ProjectsState;
+  readonly user: Resource<IUser>;
 }
 
-const HomeScreen = ({ projects }: StateProps) => {
+enum UserMenuKeys {
+  Logout = "logout"
+}
+
+const getInitials = (name: string): string =>
```
kshepard, line 28: I think this could use a little more standardization on the set of initials that is generated. As it stands, users have the ability to provide different numbers of words in the name field. This will cause the avatar to have a different size depending on how many words there are. Similarly, capitalization is currently left as-is in the avatar. I'd vote that we fix the abbreviation at two characters (so perhaps the first words plus the last word, and pad with a space if there's only one word), also capitalize these characters.

examples: John Bob Smith -> "JS", john bob Frank Smith -> "JS", johnny -> "J "

maurizi: Woops this is an oversight on my part, I meant to do that through CSS styling and then... forgot....

jfrankl: I'd consider using https://www.npmjs.com/package/react-avatar. We used it on a project and it worked well. It's smart enough to do what Kenny is talking about, where it can pick initials based on a name string. Also opens up the door to using profile photos in the future.

maurizi: I like `react-avatar`!

localhost_3003_ (16)

src/client/screens/HomeScreen.tsx
```diff
@@ -1,35 +1,118 @@
...
+import { IUser } from "../../shared/entities";
+import { resetState } from "../actions/root";
 import { projectsFetch } from "../actions/projects";
+import { userFetch } from "../actions/user";
+import { clearJWT, getJWT } from "../jwt";
 import { State } from "../reducers";
 import { ProjectsState } from "../reducers/projects";
+import { Resource } from "../resource";
 import store from "../store";
+import theme from "../theme";
```
kshepard, line 19: `Button` and `theme` aren't being used.

```suggestion
import { Box, Divider, Flex, Heading, jsx, Styled } from "theme-ui";

import { IUser } from "../../shared/entities";
import { resetState } from "../actions/root";
import { projectsFetch } from "../actions/projects";
import { userFetch } from "../actions/user";
import { clearJWT, getJWT } from "../jwt";
import { State } from "../reducers";
import { ProjectsState } from "../reducers/projects";
import { Resource } from "../resource";
import store from "../store";
```

PR #287, opened 2020-08-10
Select geometries at any geolevel

### LINKED ISSUE #267: Make transition between non-zoom-level-limited geounits seamless
Currently, if a user has an active geounit selection and switches the geolevel, we automatically accept the changes in the selection. This automatic accepting can be unexpected and frustrating. In the prototype, the user can select geounits on one level and switch to another level and continue selecting geounits, without needing to Accept/Reject changes. We should emulate that interaction.

The one wrinkle is the full version in block support. In order to visualize blocks, we've had to implement zoom level limitations so we don't overload the browser. So we have two types of geolevels: 1) loaded and visible at every zoom level, and 2) only loaded at a limited set of zoom levels. 

We should implement seamless district drawing for geolevels in 1), because they'll always be available. We expect geolevels that fall under 2) will be for more advanced, so we'll create some more friction to get to them, and don't need this same seamlessness/fluidity.

AC:
1. For zoom levels in 1), users can select geounits in any geolevel and then switch to a different level and continue selecting units. The selection will stay in progress and will not be automatically accepted/rejected.
2. When a user accepts the selection, all the selected geounits of all levels are added to the district
3. Rejecting the selection will clear all selected geounits

Note: The interaction for geolevels in 2) will change, but will be handled by a separate issue.

### LINKED ISSUE #268: Add friction to switching between zoom-level-limited geolevels
As explained in #267, some geolevels are limited to certain zoom levels. Because these zoom-level-limited geolevels are generally more advanced, we should add some friction/explanation so less experienced users don't get confused.

AC:
1. Users can only switch to zoom-level-limited geolevels if they 1) are zoomed in far enough and 2) if they have no active selection in other geolevels
2. If either of those criteria are not met, the button to switch to the geolevel is deactivated. When they are met the button is activated.
3. There is a tooltip when hovering over the geolevel to indicate which of these criteria are met/not, similar to the password validation tooltip
4. If the user selects the ZLL geolevel and makes a selection, the non-ZLL geolevels buttons are deactivated.
5. There is a tooltip over the non-ZLL geolevels that indicates the criteria for switching to them (no active selections), similar to 3.

@jfrankl has a design for how this should look.

pcaisse:
## Overview
Allow for selecting geometries at any geolevel (counties + tracts/blockgroups at the same time, or just blocks) and add a tooltip so it's clear to the user why certain geolevels are deactivated.

### Demo
db_geolevel_editing

db_subgeounit_selection

### Notes
This includes two different issues since they were related and #267 turned out to basically be a one-line change to get rid of auto-accepting since everything about selecting features at different geolevels as part of the same selection already worked (EDIT: except what commenters pointed out). #268 included some rules about not being able to select both the base geolevel (blocks) and some other geolevel at the same time as well as informative tooltips with messaging to explain what was going on to the user.

After asking @jfrankl about the tooltip, we decided to move away from showing which criteria were met vs not met and new mocks were created for the tooltip (see https://github.com/PublicMapping/districtbuilder/issues/268#issuecomment-671511801) 

These mocks don't cover the acceptance criteria in #268 numbers 4 and 5 which describe the tooltip for the non-zoom level limited geolevels, but I kept it consistent in terms of the `<conditions> to edit <geolevel>` pattern.

This also includes two very small unrelated changes (cb838ed and f9cb79e)

## Testing Instructions
* While zoomed out, select a blockgroup/tract and a county as part of the same selection
* Ensure the tooltip text for "Blocks" tells you to zoom in and resolve the changes
* Accept the changes and make sure that works fine
* Now the 'Blocks' tooltip text should tell you just to zoom in
* Select a county/blockgroup/tract
* Now the 'Blocks' tooltip text should tell you just to resolve changes
* Cancel the selection
* Zoom to the block level and now you should be able to select 'Blocks'
* Select a block and mouse over the disabled buttons and you should see the tooltip tell you to zoom out to edit those geolevels
* Accept or cancel and all geolevels should be enabled
* Select an intermediate geolevel geounit (blockgroups/tracts)
* Switch to 'County' and select the whole county and ensure that the county selection envelops the smaller geounit selection correctly (need to test with both selection tools)
* Select an entire county and then switch to intermediate geolevel and ensure geounits can be selected/deselected correctly and that this saves correctly
* Check that demographic totals are being summed up correctly when a selection includes geounits from multiple geolevels

Closes #267 
Closes #268


jfrankl:
This does not quite mimic the functionality in the prototype around #267. A user should be able to select a county and then unselect block groups within the county, as demonstrated in the below videos.

**Prototype**
prototype

**Staging**
staging

The way this is handled in the prototype is that when a user clicks on a county, behind the scenes the application is selecting block groups. It uses the index to convert from counties to block groups to do this. That way, when you switch from county to block group, it feels seamless.

pcaisse:
> A user should be able to select a county and then unselect block groups within the county

Ah, nice. There's that complexity I'd forgotten about! Thanks, I'll add to that to this.

pcaisse:
@kshepard Okay, I think this is ready for another look :sweat_smile:

pcaisse:
> Let me know if you'd like a GIF demonstration.

Yes, please. It seemed to me like it was working well for selecting counties and tracts at the same time for my test DE data.

kshepard:
> > Let me know if you'd like a GIF demonstration.
> 
> Yes, please. It seemed to me like it was working well for selecting counties and tracts at the same time for my test DE data.

Peek 2020-08-13 10-39

pcaisse:
There is one outstanding issue related to this (which in fact isn't related to this changes but was a potential problem before and only just now revealed itself). If you select county 0 and then switch to tracts, thereby selecting all tracts in the county, and then save, tract 0 is "missing" from the selection and isn't saved:
db_duplicate_feature_id
This is actually due to the duplicate feature id (both county 0 and tract 0 have a feature id of 0) which is a problem because we were assuming feature ids to be unique (when we remove the county 0 we accidentally end up removing tract 0 as well). Previously we could never select a county and tract at the same time, so there never was a conflict. We should address this in a separate issue which I can write up.

pcaisse:
Geounit selection should be much zippier now:
db_selection_fast

pcaisse:
> Definitely write up that other issue you came across.

See https://github.com/PublicMapping/districtbuilder/issues/307

PR #291, opened 2020-08-11
Redirect to main home screen on registration

### LINKED ISSUE #246: User should be sent to their map list view after registering
## Description
A user should be sent to their map list view after registering.

## Steps To Reproduce

1. Go to https://staging.districtbuilder.azavea.com/register
2. Fill out information and click "Let's go!" button

## Expected behavior
After entering information to create an account, the user should be dropped directly into the application. This wireframe shows what that looks like:

image

## Actual behavior
I am taken to a screen that tells me to check my email.

image

## Additional context
We want to get users into the application as quickly as possible, without jumping through hoops.

maurizi:
## Overview

 - Redirect to main home screen on registration
 - Show an alert that the user needs to verify their email address on the home screen
   - With a link to resend the verification email

### Demo

localhost_3003_(iPhone 5_SE) (2)
localhost_3003_(iPad Pro) (1)
localhost_3003_(iPad Pro)
localhost_3003_(iPhone 5_SE) (1)
localhost_3003_(iPhone 5_SE)
localhost_3003_ (17)

### Notes

 - This is just using the default `<Alert>` styling, I haven't made any attempt to match the wireframes as the whole Home screen needs a design pass
 - The copy could use a look by @tgilcrest 

## Testing Instructions

- Register a new user
- Observe the alert banner
- Try using the "Resend email" button
- Verifying your email should make the banner go away

Closes #246

### DIFF
--- a/src/client/api.ts
+++ b/src/client/api.ts
@@ -74,6 +74,15 @@ export async function initiateForgotPassword(email: string): Promise<void> {
   });
 }
 
+export async function resendConfirmationEmail(email: string): Promise<void> {
+  return new Promise((resolve, reject) => {
+    apiAxios
+      .post(`/api/auth/email/resend-verification/${email}`)
+      .then(() => resolve())
+      .catch(error => reject(error.response && error.response.data));
+  });
+}
+
 export async function activateAccount(token: string): Promise<JWT> {
   return new Promise((resolve, reject) => {
     apiAxios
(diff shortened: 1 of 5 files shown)


tgilcrest:
Looks good. The one change I would make is to the error text. I think it should say: "Error resending email. If this error persists, please contact us at support@districtbuilder.org."

One other question: if they click `Resend email` a second time while the error message is showing, will it give any indication that it's been clicked again? If not, it may seem like it's not responding. Even if we just close and re-open the error message I think that would be sufficient.

maurizi:
> One other question: if they click Resend email a second time while the error message is showing, will it give any indication that it's been clicked again? If not, it may seem like it's not responding. Even if we just close and re-open the error message I think that would be sufficient.
It does close & reopen the error message :+1:

PR #308, opened 2020-08-17
Document how to get set up with data for development

pcaisse:
## Overview

Document how to get set up with data for development.

Currently this is only documented in past PRs via testing instructions. It would be nice to have the commands handy as a reference and for spinning up new developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,27 @@ BASE_URL=https://staging.districtbuilder.azavea.com docker-compose up client
 
 This will proxy local all requests directed at `/api` to `https://staging.districtbuilder.azavea.com`.
 
+### Development Data
+
+To have data to work with, you'll need to do a two step process:
+
+1. Process the GeoJSON for your state/region (this outputs all the static files DistrictBuilder needs to work in a local directory) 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,27 @@ BASE_URL=https://staging.districtbuilder.azavea.com docker-compose up client
...
+
+To have data to work with, you'll need to do a two step process:
+
+1. Process the GeoJSON for your state/region (this outputs all the static files DistrictBuilder needs to work in a local directory) 
+1. Publish the resulting files (upload to S3 for use by the app)
+
+To process PA data (note that the `data` directory is relative to `src/manage`):
+
+```
+$ ./scripts/manage process-geojson data/PA.geojson -b -o data/output-pa -n 12,4,4 -x 12,12,12
```
kshepard, line 69: Maybe add a note about needing to copy the geojson file to this directory first. And another note about creating the output directory. I think those are the full set of steps.

pcaisse: Sounds good. The script will also yell with a helpful message if the output directory is missing.

PR #309, opened 2020-08-17
Provide math for scaling task memory with additional states

### LINKED ISSUE #192: Increase memory allocation for Fargate tasks to load TopoJSON for additional states
## Overview

We established [an ADR](https://github.com/PublicMapping/districtbuilder/blob/fc3d86fb8a66ff823e4473a8a34c4561e447f8f8/doc/arch/adr-04-server-side-static-data-loading.md) in #184 to determine our approach for loading TopoJSON data in memory server-side. At this time, we are planning to launch with a subset of all 50 states, which will be increased over time.

To implement the decision in ADR-04, we need to:

- Increase the amount of memory available to each Fargate task to accomodate additional states
- Only declare the application healthy via health check once states are loaded into memory

### Notes

We have established 12 GiB as an upper bound for the amount of memory required. This is unlikely to be the actual required amount of memory, since that estimate assumed that all states were as complex as the most complex state, California. We can tune the scaling amount of allocation as we add states.

## Acceptance Criteria

Each Fargate task should be running with an amount of available memory that scales with the declared number of states to be loaded, and TopoJSON should be loaded before the application is declared healthy.

rbreslow:
## Overview

This PR adds `districtbuilder_state_count` in an effort to introduce a tunable parameter for scaling task memory.

Because Fargate has specific [supported configurations](https://aws.amazon.com/fargate/pricing/), my math takes a conservative approach where we put **1024 MiB** on the table, **3** states eat it up, and the next state will add another **1024 MiB**.

## Testing Instructions

- Set `districtbuilder_state_count` to `3`. 
- Cycle Terraform.
```bash
     ~ memory                   = "2048" -> "3072" # forces replacement
```

- Set `districtbuilder_state_count` to `4`. 
- Cycle Terraform.
```bash
     ~ memory                   = "2048" -> "4096" # forces replacement
```

- Set `districtbuilder_state_count` to `1000`. 
- Cycle Terraform. 
- See that memory is capped to `8GB (1 vCPU)`.
```bash
      ~ memory                   = "2048" -> "8192" # forces replacement
```

Closes #192

### DIFF
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -87,24 +87,28 @@ resource "aws_ecs_cluster" "app" {
   }
 }
 
+// PA uses an additional ~600MiB on my workstation. It is estimated that some
+// larger states could use up to a gig. This math will add an additional
+// 1024MiB for every state. This math will also cap memory by vCPUs to not
+// exceed Fargate limits.
+locals {
+  fargate_app_memory = min(var.fargate_app_base_memory + var.districtbuilder_state_count * 1024, (var.fargate_app_cpu / 1024) * 8192)
+}
+
 resource "aws_ecs_task_definition" "app" {
   family                   = "${var.environment}App"
   network_mode             = "awsvpc"
   requires_compatibilities = ["FARGATE"]
   cpu                      = var.fargate_app_cpu
-  // PA and DE used an additional 620MiB of memory on my workstation. However,
-  // we can't multiply states by 310MiB because Fargate requires you to provide
-  // memory in specific increments. This math will add an additional 1024MiB for
-  // every 3 states  (e.g. 3 states = 1024 MiB, 4 states = 2048 MiB, 5 states =
-  // 2048 MiB). This math will also cap memory by vCPUs to not exceed Fargate
-  // limits.
-  memory = min(var.fargate_app_memory + ceil(var.districtbuilder_state_count / 3) * 1024, (var.fargate_app_cpu / 1024) * 8192)
+  memory = local.fargate_app_memory
 
   task_role_arn      = aws_iam_role.ecs_task_role.arn
   execution_role_arn = aws_iam_role.ecs_task_execution_role.arn
 
   container_definitions = templatefile("${path.module}/task-definitions/app.json.tmpl", {
     image = "${module.ecr.repository_url}:${var.image_tag}"
+
+    memory = local.fargate_app_memory
 
     postgres_host     = aws_route53_record.database.name
     postgres_port     = module.database.port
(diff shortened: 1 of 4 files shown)


deployment/terraform/app.tf
```diff
@@ -92,7 +92,13 @@ resource "aws_ecs_task_definition" "app" {
...
   requires_compatibilities = ["FARGATE"]
   cpu                      = var.fargate_app_cpu
-  memory                   = var.fargate_app_memory
+  // PA and DE used an additional 620MiB of memory on my workstation. However,
+  // we can't multiply states by 310MiB because Fargate requires you to provide
+  // memory in specific increments. This math will add an additional 1024MiB for
+  // every 3 states  (e.g. 3 states = 1024 MiB, 4 states = 2048 MiB, 5 states =
+  // 2048 MiB). This math will also cap memory by vCPUs to not exceed Fargate
+  // limits.
+  memory = min(var.fargate_app_memory + ceil(var.districtbuilder_state_count / 3) * 1024, (var.fargate_app_cpu / 1024) * 8192)
```
hectcastro, line 101: This Slack message from yesterday may have an impact on the analysis used to derive the 310MB and 620MB numbers: https://azavea.slack.com/archives/C6UD5FG3T/p1597784330054400

deployment/terraform/task-definitions/app.json.tmpl
```diff
@@ -29,6 +29,10 @@
         "name": "NODE_ENV",
         "value": "${environment}"
       },
+      {
+        "name": "NODE_OPTIONS",
+        "value": "--max_old_space_size=${memory}"
+      },
```
rbreslow, line 35: I figured out that the default heap size for V8 is around ~3GB when I was testing DistrictBuilder locally. Based on Mike's ballpark of ~10 states around launch, we're almost certainly going to hit this.

deployment/terraform/app.tf
```diff
@@ -87,19 +87,29 @@ resource "aws_ecs_cluster" "app" {
   }
 }
 
+// PA uses an additional ~600MiB on my workstation. It is estimated that some
+// larger states could use up to a gig. This math will add an additional
+// 1024MiB for every state. This math will also cap memory by vCPUs to not
+// exceed Fargate limits.
```
hectcastro, line 93: It seems like the `(var.fargate_app_cpu / 1024) * 8192)` expression being passed to `min()` works for most of the supported Fargate configurations, but doesn't work in a 4 vCPU configuration (where `districtbuilder_state_count` is high) because it leads to 32GB and the Fargate maximum is 30GB. I think we can supply a third argument to `min()` of `32768`) to account for this case:

```hcl
fargate_app_memory = min(
  // It is estimated that some larger states could use up to a gig.
  // This math will add an additional 1024MiB for every state.
  var.fargate_app_base_memory + var.districtbuilder_state_count * 1024,
  // This math will also cap memory by vCPUs to not exceed Fargate
  // limits.
  (var.fargate_app_cpu / 1024) * 8192,
  // This ensures no configuration exceeds the upper bound of memory
  // a Fargate container can have.
  30720
)
```

PR #322, opened 2020-08-18
Feature/mvm/update simplification params

### LINKED ISSUE #145: Experiment with Simplification paramters
County and blockgroup/block don't always line up, so maybe we need to tweak the parameters.

### LINKED ISSUE #255: Demographic labels don't show up for PA blockgroups
## Overview

For some reason, demographic labels are not showing up for blockgroups in PA. All other labels are showing up correctly (PA - counties, blocks; DE - counties, tracts, blocks).

## Acceptance criteria

Demogaphic labels are being displayed for PA blockgroups

maurizi:
## Overview

Reworks the `writeVectorTiles` function of the `process-geojson` script to make boundary lines match precisely, with a greater level of detail.

This started as a basic adjustment of simplification parameters, but in order to address boundary lines not matching properly, I had to run all layers through `tippecanoe` together in order for `detectSharedBorders` to operate on all of the layers at once. This in turn increased memory consumption of the script, which required rewriting several parts to stream data instead of loading it all into memory at once.

### Demo

No "tiny triangles" and borders all align perfectly:
image

### Notes

My process iterating on the simplification parameters was as follows:
 - I first adjusted the `minWeight` parameter to `topojson-simplify`, reducing it until I could no longer visually distinguish between the original block GeoJSON and the simplified TopoJSON.
 - I then adjusted the `simplification` parameter to `tippecanoe`, starting low and increasing it until just before fine-detailed borders around rivers became too blocky and triangular

By happenstance, this also fixes #255 , which seems to have been caused by OoM issues when generating the labels.

File sizes _have_ increased with these changes, bringing the PA TopoJSON from about 180Mb to about 270Mb, with a slightly smaller percentage increase in the vector tile sizes.

## Testing Instructions

- `./scripts/manage process-geojson data/PA.geojson -b -o data/output-pa -n 12,4,4 -x 12,12,12`

Closes #255 
Closes #145


jfrankl:
The blocks look great! 👌

PR #328, opened 2020-08-19
Fix linting issue

jfrankl:
Replace var with const. Could have sworn I did this already, but maybe lost it fixing a merge conflict.


jfrankl:
Thanks @pcaisse!

PR #344, opened 2020-08-25
Fix displaying demographics in rectangle select tooltip

### LINKED ISSUE #236: Rectangle tool acts like invisible single select
## Description

When the rectangle tool is selected, it is possible to use it like the single select tool to select geounits individually. The problem, however, is that the geounit does not look like it is selected. 

## Steps To Reproduce

1. Go to a map
2. Select the rectangle tool
3. Click on a geounit

## Actual behavior

The geounit does not appear to be selected because it is not colored gray. However, the population updates in the sidebar and the Accept/Reject buttons appear, indicating that something has been selected. Clicking Accept will add the geounit to the selected district, confirming that it had been selected.

## Expected behavior

This should do one of two things. Either:
1. The geounit turns gray indicating it is selected
2. The geounit is not selected (it does not turn gray, Accept/Reject do not appear, and no changes are show in the sidebar)

I have a preference for 1. We expect the geounit to be selected as soon as the user moves the mouse. For the average user with a mouse or a trackpad, they are unlikely to be clicking without moving the mouse slightly (they may not even be able to). So the difference between a click and click+drag is minimal and we should treat them as the same action.

## Screenshots

I have the rectangle tool selected and clicked on the geounit in the red box. The Accept/Reject buttons appear and the sidebar updates, but the geounit is not highlighted.

## Your environment

 - OS: Windows
 - Browser: Happens in both Chrome and Firefox

### LINKED ISSUE #324: Rectangle tooltip doesn't update totals
## Description
The population and demographic information in the tooltip does not update as expected when using the rectangle select.

## Steps To Reproduce

1. Open project
2. Switch to blockgroups/tracts
3. Switch to rectangle tool
4. Select multiple geounits with the rectangle

## Expected behavior
The population and demographic information in the tooltip should reflect the totals.

## Actual behavior
The population and demographic information in the tooltip currently reflects the geounit under the cursor.

## Screenshots
tooltip-rectangle

## Your environment

 - OS: Mac
 - Browser: Chrome
 - Version: Latest

maurizi:
## Overview

Peek 2020-08-25 09-57

## Testing Instructions

- Use the rectangle select on the project screen

Closes #324

### DIFF
--- a/src/client/components/map/MapTooltip.tsx
+++ b/src/client/components/map/MapTooltip.tsx
@@ -133,7 +133,9 @@ const MapTooltip = ({
   ) {
     const geoLevel = staticMetadata.geoLevelHierarchy[invertedGeoLevelIndex].id;
     const selectedGeounits =
-      feature && featuresToGeoUnits([feature], staticMetadata.geoLevelHierarchy);
+      highlightedGeounits.size > 0
+        ? highlightedGeounits
+        : feature && featuresToGeoUnits([feature], staticMetadata.geoLevelHierarchy);
     const demographics =
       selectedGeounits &&
       getDemographics(staticMetadata, geoUnitHierarchy, staticDemographics, selectedGeounits);


pcaisse:
@maurizi Should this be for #324 instead of #236 ?

maurizi:
@pcaisse Yes... yes it should.

Issue #348, opened 2020-08-26
Exclude large datasets from Redux dev tools


maurizi:
## 
See this warning in the Crhome developer tools when viewing the project screen:
![image](https://user-images.githubusercontent.com/4432106/91357260-f1843f80-e7be-11ea-93ad-ed7d1fb7a428.png)

PR #349, opened 2020-08-26
Improve map screen performance

### LINKED ISSUE #285: District calculations run on all state changes
## Description
Many actions trigger the population and demographic calculations for all districts. I suspect that any change to state triggers these calculations. Here are some actions that result in district recalculation, but should not:

- Zooming in or out on the map
- Changing the "Label" dropdown
- Changing drawing tools
- Locking a district

## Steps To Reproduce

1. Add a `console.log` inside the `getDemographics` function in `functions.ts`
2. Open console
3. Use any of the actions listed in the description (zoom in, for example)

## Expected behavior
I wouldn't expect to see any console messages from these actions.

## Actual behavior
Console messages indicating that we are recalculating district populations unnecessarily.

### LINKED ISSUE #288: Map tooltip causes performance issues
## Description
The map tooltip is currently a cause of lag across the application. The movement of the tooltip itself feels slower than in the prototype.

## Steps To Reproduce

1. View the application on staging
2. Move your cursor around on the map quickly, causing the tooltip to display and update frequently
3. View the application on the prototype
4. Move your cursor around on the map quickly, causing the tooltip to display and update frequently
5. To quantify this, open the[ FPS meter in Chrome](https://medium.com/@gilfink/quick-tip-using-the-chrome-devtools-fps-meter-1bb400b63f7); for me, the protoype runs at 55+ fps and staging runs below 5 fps (it slows down significantly with developer tools open, which is curious; but it's slower than the prototype when the developer tools are not open as well)

## Expected behavior
The tooltip should move with the same speed on both applications.

## Actual behavior
The tooltip on staging is slowing down the browser, causing a frame rate decrease and other elements on the page to lag.

## Screenshots
https://www.youtube.com/watch?v=Z1Fk6pcBBy8
- Note the frame rate drop on the spinner when hovering on the map
- Note that the tooltip does not move as smoothly in the staging version

## Your environment

 - OS: MacOS
- Browser: Chrome
- Version: Latest

## Notes
- On development, `emotion` seems to be adding one or more new elements to the DOM each time the cursor moves, which seems to be increasing lag further; however, this does not happen on staging, and there is still significant lag
- The way the prototype handles the tooltip, at least for the pointer, is that it pulls the population and demographic data directly from the polygon tiles; on staging, we do not include the pop/demo data in the polygon tiles, but we might want to do that if it will improve the performance of the tooltip
- #219 will implement rectangle mode calculating the population in real time; we should leverage this in the tooltip and make sure not to duplicate those calculations

maurizi:
## Overview

A variety of performance improvements to the Project screen:
 - Moved dynamic CSS properties out of Theme-UI `sx` property to avoid recalculating styles
 - Added memoization in a few places
 - Added a  separate throttle to the map tooltip for its _contents_ to provide smooth motion w/o needing as many recalculations of data
 - Fixed up some event handlers that were getting disabled/enabled on every render

### Checklist

- ~[ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible~

### Demo

Peek 2020-08-26 17-07

### Notes

Note that there is still a significant performance hit when the chrome dev tools are open (partially but not fully caused by https://github.com/PublicMapping/districtbuilder/issues/348). This seems to go away when the dev tools are closed so I'm not that worried about it.

I think any further performance enhancements on this page will likely require the use of web workers to offload demographics calculations to a background thread

## Testing Instructions

- Go to the project screen
- Observe the improved performance, particularly with the rectangle selection

Closes #285 
Closes #288

### DIFF
--- a/src/client/components/DemographicsTooltip.tsx
+++ b/src/client/components/DemographicsTooltip.tsx
@@ -39,8 +39,10 @@ const Row = ({
     <Styled.td sx={style.label}>{label}</Styled.td>
     <Styled.td sx={{ minWidth: "50px", py: 0 }}>
       <Box
+        style={{
+          width: `${percent}%`
+        }}
         sx={{
-          width: `${percent}%`,
           height: "10px",
           backgroundColor: color,
           borderRadius: "1px"
--- a/src/client/components/ProjectSidebar.tsx
+++ b/src/client/components/ProjectSidebar.tsx
@@ -14,7 +14,8 @@ import {
   IStaticMetadata,
   LockedDistricts
 } from "../../shared/entities";
-import { assertNever, getDemographics, getTotalSelectedDemographics } from "../../shared/functions";
+import { assertNever } from "../../shared/functions";
+import { getDemographics, getTotalSelectedDemographics } from "../functions";
 import {
   getDistrictColor,
   negativeChangeColor,
--- a/src/client/components/map/MapTooltip.tsx
+++ b/src/client/components/map/MapTooltip.tsx
@@ -68,16 +68,16 @@ const MapTooltip = ({
     ? staticMetadata.geoLevelHierarchy.length - geoLevelIndex - 1
     : undefined;
 
-  const throttledSetFeature = throttle((point, geoLevel) => {
-    const features =
-      map &&
-      map.queryRenderedFeatures(point, {
-        layers: [levelToLineLayerId(geoLevel), levelToSelectionLayerId(geoLevel)]
-      });
-    features && setFeature(features[0]);
-  }, SET_FEATURE_DELAY);
-
   useEffect(() => {
+    const throttledSetFeature = throttle((point, geoLevel) => {
+      const features =
+        map &&
+        map.queryRenderedFeatures(point, {
+          layers: [levelToLineLayerId(geoLevel), levelToSelectionLayerId(geoLevel)]
+        });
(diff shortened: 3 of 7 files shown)


maurizi:
@kshepard I rebased this on top of `develop`, which fixess the Map Tooltip issue (needed the fix in #344).

> Also, I added a console.log as is described in #285, and I'm still seeing the calculation run frequently (in the ways that are laid out in that issue).

I'm not sure what you're referring to here though - because of the memoization in place, you shouldn't be seeing multiple calls into `getDemographics` without relevant data changing (e.g. selected geounits), and my own testing has shown that to be working.

There are almost certainly still some unneeded re-renders going on (I fixed those for `ProjectSidebar` for the most part, but not other components), but that shouldn't be reflected in duplicate calls to `getDemographics`.

kshepard:
I edited the file: `src/shared/functions.ts`, and put a `console.log` in the first line of the `getDemographics` function. Here's a GIF showing the many calls to `getDemographics` when just hovering over the map:

get-demographics

maurizi:
@kshepard those calls are to be expected - they're getting demographics for the _tooltip_ which is why it is happening on hover.

The interactions in https://github.com/PublicMapping/districtbuilder/issues/285 were for things that _shouldn't_ need to call `getDemographics`

kshepard:
Thanks, that makes sense.

PR #351, opened 2020-08-27
Increase max zoom level

### LINKED ISSUE #231: Increase max zoom level
## Description
While testing the new drawing changes, I tried to zoom into blocks. I was able to zoom to the point where blocks appeared, but could zoom no further. We should make it possible to zoom in further so that users can work with block effectively. We want to make sure this is accessible for a wide range of users, like those who need larger shapes to make them discernible, so they need to be able to zoom far enough in so the smallest of blocks can be made large.

AC:
1. When a user switches to blocks, they are able to zoom far enough into the map so that all the blocks are clearly visible and can easily be selected with the pointer tool.

## Screenshots
This image demonstrates a significant range of block sizes. The blocks in the red box are too small to see and to work with. The blocks in the yellow box are just too small to work with, and the blocks in the green are comfortably sized. In fact, I would suggest we even let users go a bit larger. 

image

jfrankl:
## Overview

Increases the max zoom level of the map while maintaining the max zoom level for tiles.

### Demo

Map that zooms in farther than before

### Notes

To determine whether to enable a geolevel at a certain zoom level, we were checking that the current zoom level is smaller than the maxzoom for that geolevel. I needed to remove this to allow us to enable a geolevel even if we were zoomed past the maxzoom. Maxzoom, in this case, is about tile availability, not layer visibility.

To set the new maxzoom level for the map, I am adding 4 to the tile max zoom. This is a number that looked good to me for Philadelphia, but we might need to adjust after testing with additional states. I think a better long term solution would be to add this value to the state config, so we can adjust it on a case-by-case basis. I wasn't sure how to do that, but if people feel that this would be good to do now, I'm happy to make this change with support.

## Testing Instructions

- Open a project
- Press the + zoom buttom to zoom in one level at a time
- Zoom in until the "Blocks" button becomes enabled
- You should be able to continue to zoom in four additional levels after you reach this point

Closes #231

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -76,6 +76,9 @@ const Map = ({
 
   const minZoom = Math.min(...staticMetadata.geoLevelHierarchy.map(geoLevel => geoLevel.minZoom));
   const maxZoom = Math.max(...staticMetadata.geoLevelHierarchy.map(geoLevel => geoLevel.maxZoom));
+
+  // While a geolevel has tiles up to the maxZoom level, we want the enable the user to zoom in
+  // beyond that zoom level. Using lower zoom tiles at higher zoom levels is called overzoom.
   const overZoom = maxZoom + 4;
 
   // Add a color property to the geojson, so it can be used for styling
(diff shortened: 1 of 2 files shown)


jfrankl:
Thanks for reviewing

Issue #326, opened 2020-08-19
Migrate districtbuilder.org hosted zone to Route 53


hectcastro:
Ensure that all of the DNS records in GoDaddy are migrated to Route 53, then update the `districtbuilder.org` name server configuration on GoDaddy to delegate to Route 53.

This will likely require Terraform state manipulation for both the production and staging AWS accounts.

rbreslow:
In addition, we should update the `r53_public_hosted_zone` variable to reflect the new domain in both staging and production.

ColeKettler:
@hectcastro Nameservers have been switched over in GoDaddy DNS management to point to the nameservers in our districtbuilder.org hosted zone. Everything looks to be functioning as expected.

I've also removed the districtbuilder.azavea.com hosted zone from Terraform state, manually deleted and Terraform-managed records, and retained the pa.districtbuilder.azavea.com NS records.

This is ready for review.

hectcastro:
Everything looks good. I think both versions of DB are functioning as expected with the update domains, and the DTL site is still working well. 👍

PR #360, opened 2020-08-28
Migrate DNS records to districtbuilder.org hosted zones

ColeKettler:
## Overview

These changes support us migrating DNS records from GoDaddy to AWS Route 53, and using `districtbuilder.org` DNS records for the application. The application will now be hosted at `app.staging.districtbuilder.org` and `app.districtbuilder.org`.

### Checklist

### Notes

Production records have not been altered yet. Staging NS records have been added to GoDaddy for testing.

## Testing Instructions

- Verify that the staging application is functioning at https://app.staging.districtbuilder.org/
- Verify that all `districtbuilder.org` records have been copied from GoDaddy to the `districtbuilder.org` domain in Route 53.

Connects #326


ColeKettler:
Thanks for the extra scrutiny! I'll get those values updated in the tfvars.

Looks like the JWT secret is just an arbitrary secret value used for HMAC signing. I don't think it can be dependent on the `CLIENT_URL`.

I'll remove any Terraform-managed records from the `districtbuilder.azavea.com` hosted zone and remove it from the Terraform state.

PR #395, opened 2020-09-09
Realtime sidebar updates

### LINKED ISSUE #219: Calculate sidebar updates in real-time as rectangle tool is dragged
## Overview

In the prototype, the sidebar updates in real-time with the pending rectangular selection. In our app, the sidebar only updates after the user completes their rectangular selection. We would like this to behave in a similar manner as the prototype. However, there are some additional considerations to account for, since we are dealing with a more complicated set of calculations than the prototype needs. Because of this, we expect that a naive approach here would result in sluggish behavior as many potentially hefty calculations will be triggered in quick succession. An approach to improve the user experience may involve something such as the following:
 1) Debounce the calculation requests
 2) Make use of web workers to ensure the user experience doesn't suffer as a calculation is churning

## Acceptance criteria

The sidebar is updating in real-time as the rectangle tool is dragged. And there is no major UI sluggishness resulting from this change.

maurizi:
## Overview

Refactors demographics calculations to happen in a web worker, and updates the project sidebar to update demographic data in as the rectangle selection is used.

### Checklist

### Demo

Peek 2020-09-09 15-05

### Notes

I originally tried using `worker-loader` & `comlink` for the web worker communication before switching over to `workerize-loader`, which I had an easier time getting to work with TypeScript.

## Testing Instructions

- `scripts/update`

Closes #219

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-react": "^7.20.0",
     "http-proxy-middleware": "0.20.0",
-    "prettier": "1.19.1"
+    "prettier": "1.19.1",
+    "workerize-loader": "^1.3.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "@tippyjs/react": "^4.1.0",
     "@types/jwt-decode": "2.2.1",
     "axios": "0.19.0",
+    "json-stable-stringify": "^1.0.1",
     "jwt-decode": "2.2.0",
     "lodash": "^4.17.19",
     "mapbox-gl": "1.6.1",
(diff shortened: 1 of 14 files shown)


jfrankl:
The numbers update as expected in the sidebar! 

Unfortunately, this PR seems to add a significant drop in performance on my frontend, resulting in a noticeable lag when using it. I'm surprised to see this, since we're using web workers. Do you experience this on your end?

It's okay for there to be a delay on when the sidebar/tooltip update (in fact, I'd recommend debouncing those), but the tooltip movement itself shouldn't be slowing down or reducing in frame rate. That is a sign of a bigger performance issue.

jfrankl:
I haven't looked into the performance issue in this case, but something I learned from working with web workers on the prototype is that there can be performance issues around sending information back and forth from the web worker. There are ways to send information faster using `transferable objects` and sending as arrayBuffer. Sorry if you already know that or are doing that.

kshepard:
Superseded by #452

Issue #374, opened 2020-09-01
Pre-launch fixups


tgilcrest:
Covers any last little fixes we need to make for things we find in testing.

- [ ] Change election data labels from `votedem`, `voterep`, `voteoth`, to `Voted Dem.`, `Voted Rep.`, `Voted Oth.` (tentative)
  - (k.s.) I don't think we need to do this right now, since none of our states have election data loaded

### kshepard working on
 - [x] Re-load PA data for new labels
 - [x] Re-load VA data for new labels
 - [x] Re-load DE data for new labels
 - [x] Similar to the change in #356, some counties already end in "County", so don't append an additional "County" in this case. See New Jersey for an example. Edit: Luke has made a change so the "County" suffix shows up when it should, so we can remove our formatting logic for this and just show what's there. Adding additional data reloading tasks below to account for this.
 - [x] On map page, change button text from `Approve` to `Accept` for pending changes
 - [x] On create map page, change button text from `Create Project` to `Create Map`
 - [x] On create map page, change header text from `New Project` to `New Map`
 - [x] On create map page, under Custom, change "Define other types of districts" to "Define a custom number of districts"
 - [x] Lazily cache new region data. Context from Slack message: 
>I think we may need a way to tell the system that there's a new region to cache. Right now, if I add a new region to the database, I can then select it in the frontend and it takes a very long time to load (because it's pulling down the files), and then it takes a very long time on each save (presumably because the initial pulling of the data doesn't save it in memory). So I've been needing to stop the ECS task and have a new one come up to cache it. Maybe we just make a change to save it in memory after initial load even if it wasn't performed on startup -- I think that would essentially solve it: just the first person (one of us) would take the initial performance hit.
- [x] Add a warning to the page if a user tries to leave without accepting/canceling changes 
![Screen Shot 2020-09-08 at 12 38 34 PM](https://user-images.githubusercontent.com/1809908/92504150-3dc47c00-f1d0-11ea-9b24-d5005a73f6de.png)
- [x] Fix map tooltip flicker per https://github.com/PublicMapping/districtbuilder/issues/279#issuecomment-689894115
- [x] Configure production data

tgilcrest:
@jfrankl Proposed change for warning text:
> You have unsaved changes on your map. Are you sure you want to leave this page?
>
>Accept or reject changes to save your map.

jfrankl:
[...]

PR #396, opened 2020-09-09
Update for county suffixes

kshepard:
## Overview

The data processing pipeline now passes through the "County" suffixes, so we no longer need to add them to the label after the fact. I also performed some additional minor text changes:
 * On map page, change button text from Approve to Accept for pending changes
 * On create map page, change button text from Create Project to Create Map
 * On create map page, change header text from New Project to New Map
 * On create map page, under Custom, change "Define other types of districts" to "Define a custom number of districts"

### Checklist

### Demo

city-suffix

county-suffix

text-canges

## Testing Instructions

- Make sure you have loaded one of the new region files that has county suffixes passed through
- Verify that you are not seeing doubled-up county suffixes
- Verify the minor text changes described in the overview are present

### DIFF
--- a/src/client/components/ProjectSidebar.tsx
+++ b/src/client/components/ProjectSidebar.tsx
@@ -229,7 +229,7 @@ const SidebarHeader = ({
             }}
           >
             <Icon name="check" />
-            Approve
+            Accept
           </Button>
         </Flex>
       ) : null}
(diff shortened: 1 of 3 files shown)


pcaisse (approved):
Possible nitpick: I see that you've kept the case consistent for the headings and buttons, but that seems like it differs from what's specified in #374 . Might be something we verify with UX.

kshepard:
> Possible nitpick: I see that you've kept the case consistent for the headings and buttons, but that seems like it differs from what's specified in #374 . Might be something we verify with UX.

Nice catch, I'll ask.

Issue #111, opened 2020-05-02
Basemap switcher


tgilcrest:

jfrankl:
I just opened a pull request that adds a basic basemap to the application #401. That will be suitable for launch.

In terms of adding a basemap switcher, I think we should spend some time thinking through our needs here. Getting the aerial imagery view to work well might be a big challenge, and I think it would be worth confirming its value. Instead of having 2 or 3 basemaps to choose from, we may want to explore an option where the user can choose between aspects of the basemap. For example: toggle street labels, toggle city labels, toggle hillshade.

PR #401, opened 2020-09-10
Add basemap

jfrankl:
## Overview

Adds a custom basemap from Mapbox Studio.

### Demo

Screen Shot 2020-09-10 at 6 06 00 PM

Screen Shot 2020-09-10 at 6 08 42 PM

Screen Shot 2020-09-10 at 6 07 31 PM

### Notes

To use a Mapbox Studio basemap, I had to change the way we define a style in the application. Instead of generating a style json, we are now using a style from Mapbox Studio, and we use `addSource` and `addLayer` to add additional elements. We were using that technique in some places already, so now it is more consistent. I made a `generateMapLayers` function to move this outside of the crowded `Map.tsx` file, but let me know if you'd prefer a different approach.

I am using a filter to only show labels for the selected region. Right now this will work for any state, but when we support other types of regions, we'll need to make some minor changes. I figure there are a lot of little things like that.

I think this is in a good enough place to launch, but I hope to continue to refine it over time. That being said, let me know if you see anything super weird. 

In the future, I want to allow users to have more control over which layers they see. 

## Testing Instructions

- Open a project from 3 different states
- For each state, zoom around, check out counties, blockgroups, and blocks
- Let me know if anything looks super messed up
- Confirm that you only see city/poi labels from the selected state

Connected to #111

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -93,6 +93,7 @@ const DistrictsMap = ({
       return;
     }
 
+    // eslint-disable-next-line
     MapboxGL.accessToken =
       "pk.eyJ1IjoiZGlzdHJpY3RidWlsZGVyIiwiYSI6ImNrZXZzeXlvMjIxb2QycW1yeGpuMDJ2ZGwifQ.FdOGNk3y1BPkGvF_yCjjGQ";
 
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -23,11 +23,9 @@ import { areAnyGeoUnitsSelected, getSelectedGeoLevel } from "../../functions";
 import { getAllIndices } from "../../../shared/functions";
 import {
   GEOLEVELS_SOURCE_ID,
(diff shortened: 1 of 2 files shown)


tgilcrest (approved):
I can't load it locally but from what I can tell of the screenshots I really like the context it adds to the map. The blank version feels incomplete compared to this.

The one thing that's confusing to me is how the transparency affects the district colors. In the last image it's hard to tell which districts correspond to the ones on the map. Is the red color 8, 9, or 13? Is the blue color 7 or 16?

I don't think it's worth worrying about for now, but something we should think about soon is a good way to match the map to the districts other than color alone. One option could be adding the assigned district to the tooltip for when you hover over shapes, if it's not a performance hit.

kshepard (approved):
Looks amazing, huge improvement.

jfrankl:
@tgilcrest I agree with you about needing to find ways to connect the map and sidebar. I'll open a separate issue about that.

jfrankl:
@tgilcrest here are some ideas about connecting the sidebar and map #404. Feel free to add other thoughts.

PR #406, opened 2020-09-11
Warn when attempting to leave page with selection

kshepard:
## Overview

We want to give the user a chance to save active selections before accidentally closing the page.

### Checklist

### Demo

screenshot_7

### Notes

I provided a custom message, but no browser I tested with is able to display it because of the following.

```
Some browsers used to display the returned string in the confirmation dialog, enabling the event handler to display a custom message to the user. However, this is deprecated and no longer supported in most browsers.
```
Source: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event

## Testing Instructions

- ./scripts/update
- Load a project
- Verify you can close the page without a confirmation dialog when no geounits are selected
- Select some geounits
- Verify you get a confirmation dialog when attempting to close the page


pcaisse (approved):
:rocket:

jfrankl:
This functions as I'd expect when the user tries to close the tab or window, but not if they click their browser back button or click on the DistrictBuilder logo in the app, which are other common ways a user might leave this page without saving. In working on this issue, did you learn anything about the possibility of doing this for the browser back function? I looked for a minute and it seemed like it might not be possible.

PR #410, opened 2020-09-11
Fix S3 permissions to allow CloudFront logging

hectcastro:
## Overview

It appears as though something changed with Terraform's ability to detect out-of-band grants, or perhaps a CloudFront service level change led to different behaviors than before (they recently added real-time logs). That's left us with a handful of CloudFront logs in S3 from May, but nothing between May and now.

When logging to S3 from CloudFront is configured via the AWS console, a specific set of grants that allow CloudFront to interact with an S3 bucket are added. These grants appear to get wiped away by Terraform unless they are explicitly defined in the Terraform configuration.

The changes in this PR aim to persist the two grants (one is the AWS canonical user ID, and the other is the CloudFront canonical user ID) in the Terraform project.

### Checklist

## Testing Instructions

- Disable CloudFront logging for the staging distribution via the [AWS console](https://console.aws.amazon.com/cloudfront/v2/home#/logs/E3QOQU5DX6A1SS)
- Execute a Terraform plan/apply cycle using `develop` and see that 2 grants are being removed from the `aws_s3_bucket.logs` resource
- Execute a Terraform plan/apply cycle using the changes in this branch and ensure that no grants are removed the the `aws_s3_bucket.logs` resource

### DIFF
--- a/deployment/terraform/config.tf
+++ b/deployment/terraform/config.tf
@@ -1,6 +1,6 @@
 provider "aws" {
   region  = var.aws_region
-  version = "~> 3.1.0"
+  version = "~> 3.6.0"
 }
 
 terraform {
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -295,3 +295,8 @@ variable "aws_ecs_task_execution_role_policy_arn" {
   default = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
   type    = string
 }
+
+variable "aws_cloudfront_canonical_user_id" {
+  default = "c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0"
+  type    = string
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fix S3 permissions to allow CloudFront logging [#410](https://github.com/PublicMapping/districtbuilder/pull/410)
 
 
 ## [0.1.0] - 2020-09-14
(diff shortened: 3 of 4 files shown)


deployment/terraform/config.tf
```diff
@@ -1,6 +1,6 @@
 provider "aws" {
   region  = var.aws_region
-  version = "~> 3.1.0"
+  version = "~> 3.6.0"
```
hectcastro, line 3: I made this change first to see if maybe it was a bug in the provider, but that didn't yield any change in behavior. I kept the change in the PR because no other resources seemed to be impacted by the upgrade.

deployment/terraform/variables.tf
```diff
@@ -295,3 +295,8 @@ variable "aws_ecs_task_execution_role_policy_arn" {
   default = "arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy"
   type    = string
 }
+
+variable "aws_cloudfront_canonical_user_id" {
```
hectcastro, line 299: Unfortunately, there isn't a data resource for this like there is for the `aws_canonical_user_id`. https://github.com/terraform-providers/terraform-provider-aws/issues/12512 exists to add it to the Terraform AWS provider.

hectcastro:
A heads up @ColeKettler because this PR doesn't show up on our board and may have slipped through the cracks.

deployment/terraform/variables.tf
ColeKettler: That'd be nice, hope that change gets worked in soon!

ColeKettler (approved):
Wow, that's a subtle one. Looks like this fix works!

I feel like this should become part of our standard CloudFront config. Is it possible that this could be happening in other projects and we haven't noticed yet?

hectcastro:
> Wow, that's a subtle one. Looks like this fix works!

I think it does, but was hopeful there might be something a bit clearer out there. Thus far, I haven't found it.

> I feel like this should become part of our standard CloudFront config. Is it possible that this could be happening in other projects and we haven't noticed yet?

I think it very well may be. I opened a few issues in other projects that have similar configurations. It would be better to fix sooner rather than later so that information is available in the event that we need to troubleshoot things.

PR #413, opened 2020-09-14
Blocks button enabled by default

### LINKED ISSUE #407: Make blocks button enabled by default
## Detailed Description
When a user opens a new project, the "Blocks" button is disabled. They need to zoom in and resolve any changes before it becomes enabled, due to technical restrictions to using this feature. @alexelash made a great point at design crit that by showing this button as disabled, we might be inadvertently signaling to the user that this feature is not actually implemented, or not available to them for a harder-to-change reason. If the user hovers over the button, they will see a tooltip that explains how to activate block mode, but we can expect that some users will not find that.

A slight variation on this approach, which I think strikes the right balance between adding some friction for the advanced mode while also making it clear that it's available, would be to only disable the button when geounits have been selected, not based on zoom level.

When the user loads a project, they have access to all three geounit types. They could pick blocks right off the bat, if they wanted. Or, if they don't want it, the button being enabled will indicate that it is available in the future.
Screen Shot 2020-09-11 at 3 11 52 PM

If the user makes a selection of a county or blockgroup, the button does become disabled. The user might notice the link and understand that there is a connection between making selections and being able to use block mode. And hovering over the block button will confirm.
Screen Shot 2020-09-11 at 3 12 06 PM

If the user does select block mode, they will see the existing modal. I think this will be enough friction to stop users who might get confused by block mode.
Screen Shot 2020-09-11 at 3 17 56 PM

Since the user is zoomed out too far, we show them the "Zoom in to view blocks" message that also already exists.
Screen Shot 2020-09-11 at 3 13 16 PM

The user can zoom in until they get to where the message disappears and the blocks appear. Now we're ready to edit blocks!
Screen Shot 2020-09-11 at 3 13 42 PM

jfrankl:
## Overview

A user can enable block mode without zooming in. This makes it more clear that the block editing mode is available. Previously, when this button was disabled by default, it sent mixed signals—maybe this feature isn't built yet, or maybe I don't have access to it for a difficult-to-change reason. Depending on a tooltip to straighten that out wasn't enough.

Now a user can immediately click on the blocks button if they want. And they'll receive an appropriate warning to let them know the takeoffs of block editing. 

## Testing Instructions

- Create a new project
- In the new project, confirm that the "blocks" button visually appears enabled
- Click on the "blocks" button; you should see the blocks modal; click cancel
- Select a county on the map
- Confirm that the "blocks" button looks disabled
- Hover over the disabled "blocks" button; confirm that you see a tooltip that explains you need to resolve changes
- Resolve changes (click cancel or accept)
- Click on the "blocks" button and this time click "Start block editing"
- You should see a message overlay on the map that says you need to zoom in to edit blocks
- Zoom in; at some point, the message should disappear and you should see blocks!

Closes #407

### DIFF
--- a/src/client/components/MapHeader.tsx
+++ b/src/client/components/MapHeader.tsx
@@ -61,11 +61,9 @@ const style: ThemeUIStyleObject = {
 const buttonClassName = (isSelected: boolean) => `${isSelected ? "selected" : ""}`;
 
 const GeoLevelTooltip = ({
-  isGeoLevelHidden,
   areChangesPending,
   label
 }: {
-  readonly isGeoLevelHidden: boolean;
   readonly areChangesPending: boolean;
   readonly label: string;
 }) => {
@@ -73,14 +71,7 @@ const GeoLevelTooltip = ({
   return (
     <span>
       <strong>Disabled: </strong>
-      {isGeoLevelHidden && areChangesPending ? (
-        <span>{zoomText} and resolve changes</span>
-      ) : isGeoLevelHidden ? (
-        <span>{zoomText}</span>
-      ) : (
-        <strong>Resolve changes</strong>
-      )}
-      &nbsp;to edit {label.toLowerCase()}
+      Resolve changes to edit {label.toLowerCase()}
     </span>
   );
 };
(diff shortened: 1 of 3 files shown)


kshepard (approved):
Looks good. I just pushed a small commit to address some linter warnings and a formatting difference.

alexelash (approved):
Works as described 👍

jfrankl:
@kshepard thanks for the review and code cleanup

@alexelash thanks for the review and for suggesting this change, I think it makes the block mode way more clear.

PR #414, opened 2020-09-14
Style pass on map tooltip

### LINKED ISSUE #279: Style pass: map tooltip
### Description
Style cleanup on the map tooltip. The design in the mockup is slimmer. It's important for this component to maintain a small footprint so it doesn't cover too much of the map.

### Design assets

- Tooltip: https://share.goabstract.com/5691f15f-619c-4426-b48d-8835ca32c8f3?mode=design&sha=a408aeb9a005792b7061a451337c3a93a62c95c9

### Notes
Ignore the content highlighted in this image. We don't have that in the tooltip yet.
image

jfrankl:
## Overview

Style cleanup on the map tooltip. Tighten line-height to make it easier to compare bars in race chart, set fixed size so the size of the tooltip doesn't change as much, making it easier to quickly scan as it changes. I made the colors match for the map tool tips and tooltips that appear elsewhere in the application. I also removed the repositioning when the tooltip approaches the side of the screen (more on that in the notes).

### Demo

Screen Shot 2020-09-13 at 5 15 01 PM

### Notes

@maurizi I removed the tooltip repositioning in this PR. I wanted to address that since the repositioning was a feature I had requested and I believe you implemented. While this feature made sense to me in abstract, in practice I found it annoying to have the tooltip jump around unexpectedly. I think you did a great job implementing it, and I wanted to be clear that this isn't a knock on your code. Just one of those iterative moments where I realize one of my ideas doesn't work out as well as I had hoped.

## Testing Instructions

- Open a project
- Make sure the tooltip looks similar to the screenshot
- Confirm that when you look at the tooltip for many geounits, the width of the tooltip remains consistent
- When you move to the right side of the screen, the tooltip should no longer reposition itself

Closes #279

### DIFF
--- a/src/client/App.css
+++ b/src/client/App.css
@@ -16,3 +16,8 @@
 .mapboxgl-canvas {
   outline: none;
 }
+
+.tippy-box {
+  background-color: #141414 !important;
+  border-radius: 2px;
+}
--- a/src/client/components/DemographicsTooltip.tsx
+++ b/src/client/components/DemographicsTooltip.tsx
@@ -9,15 +9,14 @@ const style: ThemeUIStyleObject = {
     textAlign: "left",
     py: 0,
     pr: 2,
-    textTransform: "capitalize",
-    fontSize: 1
+    textTransform: "capitalize"
   },
   number: {
     flex: "auto",
     textAlign: "right",
     fontVariant: "tabular-nums",
     py: 0,
-    fontSize: 1
+    fontWeight: "light"
   }
 };
 
@@ -72,7 +71,7 @@ const DemographicsTooltip = ({
   ));
   return (
     <Box sx={{ width: "100%", minHeight: "100%" }}>
-      <Styled.table sx={{ margin: "0" }}>
+      <Styled.table sx={{ margin: "0", width: "100%" }}>
         <tbody>{rows}</tbody>
       </Styled.table>
     </Box>
--- a/src/client/components/map/MapTooltip.tsx
+++ b/src/client/components/map/MapTooltip.tsx
@@ -18,9 +18,6 @@ import { State } from "../../reducers";
 import DemographicsTooltip from "../DemographicsTooltip";
 import { levelToLineLayerId, levelToSelectionLayerId } from ".";
 
-const X_BUFFER = 300;
-const Y_BUFFER = 300;
-
 const getDemographics = memoize(getTotalSelectedDemographics);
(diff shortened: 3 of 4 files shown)


maurizi (approved):
This looks good to me!

One thing to consider instead though: changing the position of the tooltip so it never gets covered by the bottom/right edge of the screen, and letting it go over top of the sidebar & header:
image

jfrankl:
@maurizi That's a really interesting idea. Is that screen from the actual application with some code modifications, or is it a mockup?

maurizi:
From the application. Here is the diff:
```diff
diff --git a/src/client/components/map/MapTooltip.tsx b/src/client/components/map/MapTooltip.tsx
index 445d667..aa2fb15 100644
--- a/src/client/components/map/MapTooltip.tsx
+++ b/src/client/components/map/MapTooltip.tsx
@@ -22,7 +22,7 @@ const getDemographics = memoize(getTotalSelectedDemographics);
 
 const style: ThemeUIStyleObject = {
   tooltip: {
-    position: "absolute",
+    position: "fixed",
     margin: "20px",
     backgroundColor: "gray.8",
     color: "muted",
@@ -33,7 +33,7 @@ const style: ThemeUIStyleObject = {
     overflow: "hidden",
     pointerEvents: "none",
     p: 2,
-    zIndex: 1,
+    zIndex: 10000,
     lineHeight: 1.3,
     fontSize: 0,
     willChange: "transform"
@@ -164,10 +164,12 @@ const MapTooltip = ({
           ).toLowerCase()}`
         : featureLabel();
 
-    const x = point.x;
-    const y = point.y;
+    const tooltipWidth = tooltipRef.current && tooltipRef.current.offsetWidth;
+    const tooltipHeight = tooltipRef.current && tooltipRef.current.offsetHeight;
+    const x = tooltipWidth && point.x - tooltipWidth - 40;
+    const y = tooltipHeight && point.y - tooltipHeight - 40;
 
-    return demographics ? (
+    return demographics && x !== undefined && y !== undefined ? (
       <Box
         ref={tooltipRef}
         style={{ transform: `translate3d(${x}px, ${y}px, 0)` }}
```

jfrankl:
Thanks @maurizi. It felt off to me having it on the right. But I might try this again in the future. It's a great idea and it does solve the issue of the tooltip being out of view—there might be a way I can refine it to have it look better.

PR #415, opened 2020-09-14
Style map message and add zoom functionality

### LINKED ISSUE #372: Style Map Message Component

alexelash:
## Overview
This PR styles the "zoom in to work with blocks" message to pop more against its background. It also turns it into a button so that a user can quickly get to the correct zoom level to work with blocks.

### Checklist

### Demo
Screen Shot 2020-09-14 at 1 17 42 PM

districtbuilder_zoom_button

### Notes
N/A

## Testing Instructions
- `git pull`
- Open up a map. Zoom in close enough that the "Blocks" button is enabled. Click "Blocks" to begin editing at the block level.
- Now use the - arrow to zoom out.
- You should see the map message that says "zoom in to work with blocks". It should have a hover state and display a pointer cursor.
- Click the button.
- The message should disappear and you should see blocks again.

Closes #372


src/client/components/map/MapMessage.tsx
```diff
@@ -44,10 +68,12 @@ const MapMessage = ({
...
 
   return geoLevelVisibility[geoLevelIndex] ? null : (
-    <Box sx={style.message}>
-      <span>
-        <strong>+</strong> Zoom in to work with {levelLabel}
-      </span>
+    <Box
+      sx={style.message}
+      as="button"
+      onClick={() => (map !== undefined ? map.zoomTo(maxZoom) : null)}
```
jfrankl, line 74: Nice, this works great. You might want to consider using a [&& operator](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator) instead of the ternary operator because there is no "else" aspect of this condition.

jfrankl (approved):
Hey Alex,

jfrankl:
Hey @alexelash this looks great. One thing I noticed, which wasn't captured in this issue, is that this component isn't actually horizontally centered. Would you be able to horizontally center it before merging?

alexelash:
@jfrankl Good catch, I didn't notice that. I've updated to resolve both suggestions in 54fddee.

Issue #417, opened 2020-09-14
Add messaging when selecting county/blockgroup containing locked block


pcaisse:
## Detailed Description
This is needed because we've decided to not allow for partial selection that would result in selecting blocks since block selections are not available at all geolevels. Here is the direction from UX:

> For blocks, we are okay with handling them differently, since they are a more advanced part of the tool. For these, we should show a toast message when we are unable to make a selection due to locked blocks. We should throttle the message or find another way for it not to appear over and over if the user is selecting with the rectangle tool 

Additionally, we decided that the message should link to the docs (placeholder here: https://github.com/PublicMapping/districtbuilder/wiki/Block-Editing)

PR #420, opened 2020-09-14
Allow for selecting partially locked districts

### LINKED ISSUE #323: Support selecting partially locked geounits

pcaisse:
## Overview
Allow for selecting partially locked districts.

### Checklist

### Demo
Selecting/deselecting locked counties
db_partially_locked_final

Partial selection with locked blocks (messaging will be added to improve the UX of this in https://github.com/PublicMapping/districtbuilder/issues/417)
db_partially_locked_selection_ui_weirdness

### Notes
There were a couple of existing bugs that I ran into which had to be smoothed out to get everything working:
* For the rectangle selection tool, `getFeaturesInBoundingBox` only returned features at the currently selected geolevel, though since we made the change to allow for working with multiple levels at once, we need to get all features from all levels (that's what the new `getAllSelectedFeatures` does)
* `isGeoUnitLocked` wasn't quite working as advertised and would quit recursing too early if the district definition was jagged which could result in incorrect behavior

Much of the meat of the changes was to `onlyUnlockedGeoUnits` which needed to be modified to break out a geounit such as a county into its constituent unlocked sub-geounits. I kept things imperative-style since the `Map` API is imperative, for better or for worse. I'm not totally sure about the name `removeLockedGeoUnits` since it removes geounits that are partially locked and _adds_ the unlocked child geounits, but I think it does get the idea across and there is a docstring to try to explain it.

There is a good amount of complexity within the selection tools. There already was a good amount of complexity to the rectangle selection tool, though it didn't have to change all that much (complexity stayed about the same). The default selection tool did need to become significantly more complex because it needed to support the selection _and_ deselection of partially locked geounits. I added quite a few comments there to make that as clear as I could.

I ended up adding a helper function `getChildGeoUnits` which is used liberally for implementing this functionality. It returns child geounits but no further (no grandchildren). This seems fine for now since we're only dealing with breaking up counties into blockgroups at the moment. This function is also now used instead of the now defunct `findSelectedSubFeatures` which queried the Mapbox map. Given the limitations around Mapbox only returning features within the current viewport it seems preferable to use our own static metadata for feature querying whenever possible (and I would think it would be faster though I'm not sure about that).

## Testing Instructions
* Select a blockgroup and assign it to a district
* Lock that district
* Using the default selection tool, select the county containing the locked blockgroup
* Ensure all non-locked blockgroups are selected
* Deselect the county and all of the blockgroups should be deselected
* Select a single blockgroup and then select the whole county containing the blockgroup
* Ensure the single blockgroup is not double-selected
* Assign a block to the locked district
* Try to select the blockgroup which contains it (nothing should happen)
* Try to select the county which contains the blockgroup (all the blockgroups except the one containing the locked block should be selected)

Closes #323

### DIFF
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -207,7 +207,7 @@ export function destructureResource<T extends object>(
   return "resource" in resourceT ? resourceT.resource[key] : undefined;
 }
 
-export function mergedGeoUnits(geoUnitsA: GeoUnits, geoUnitsB: GeoUnits): GeoUnits {
+export function mergeGeoUnits(geoUnitsA: GeoUnits, geoUnitsB: GeoUnits): GeoUnits {
   const mergedGeoUnits: MutableGeoUnits = {};
   Object.entries(geoUnitsA).forEach(([geoLevelId, geoUnitsForLevel]) => {
     const mergedGeoUnitsForLevel = new Map(geoUnitsForLevel);
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -10,7 +10,8 @@ import {
   GeoUnitIndices,
   GeoUnitHierarchy,
   IStaticMetadata,
-  NestedArray
+  NestedArray,
+  MutableGeoUnits
 } from "../shared/entities";
 import { Resource } from "./resource";
 import { getDemographics as getDemographicsBase } from "../shared/functions";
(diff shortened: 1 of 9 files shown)


src/client/functions.ts
```diff
@@ -205,3 +210,16 @@ export function destructureResource<T extends object>(
...
+  Object.entries(geoUnitsA).forEach(([geoLevelId, geoUnitsForLevel]) => {
+    const mergedGeoUnitsForLevel = new Map(geoUnitsForLevel);
+    geoUnitsB[geoLevelId].forEach((geoUnitIndices, featureId) => {
+      mergedGeoUnitsForLevel.set(featureId, geoUnitIndices);
+    });
+    // eslint-disable-next-line
+    mergedGeoUnits[geoLevelId] = mergedGeoUnitsForLevel;
+  });
+  return mergedGeoUnits;
+}
```
maurizi, line 225: I don't think this function will correctly handle the case where there are levels present in `geoUnitsB` that are not in `geoUnitsA`. You could copy the implementation I used here https://github.com/PublicMapping/districtbuilder/pull/395/files#diff-1b8d94cb7809453fae3e461753b8abecR22-R27 which should handle that.

tgilcrest:
I noticed an issue with the current implementation of locking where if you have an unlocked block inside a locked blockgroup it allows you to select the whole county. Is this addressed in this PR?

Green is blockgroup that has a block in another, unlocked district. When selecting counties it's possible to select the county with the partial locked blockgroup in it. With the changes in this PR, I would expect it to only select the blockgroups around the green blockgroup, so it may be fixed but I wanted to check.
locking-bug

src/client/functions.ts
pcaisse: Yeah, you're right. I guess I was assuming they'd have the same keys which I think they always will in practice but I like your implementation better anyway.

pcaisse:
@tgilcrest Yeah, I just tested it and that seems to be fixed :+1: I believe that is related to the second of the two bugs I mentioned in the notes

jfrankl:
This is so excited! Great looking work @pcaisse.

kshepard (approved):
Excellent, seems to work great.

PR #423, opened 2020-09-15
Add wiki images

tgilcrest:
## Overview

I need to add screenshots to the documentation in the wiki. This PR uploads the images I need for the Getting Started Gude.


tgilcrest:
Added images directly to wiki instead.

PR #436, opened 2020-09-21
Bigger click area and hover state for lock button

### LINKED ISSUE #424: Improve lock button UX
## Detailed Description
The lock button does not have a hover effect and has a small click area. We should make it easier to click and also make it clear when your click will toggle the lock or select the district.

alexelash:
## Overview
Makes a larger click area and adds a hover state to the lock button in the sidebar. I also added a slight transition for button hover states and applied our `:focus` state to the logo link in the project header. 

### Checklist

### Demo
districtbuilder_hover_click_area

### Notes
- I added a new `icon`button variant for these. 
- The width of the sidebar had to get a bit wider in order to fit the new width of the button.

## Testing Instructions
- `git pull`
- Enter a map
- Hover on rows in the sidebar. Notice that the lock icon appears.
- Hover on a lock icon to see its `:hover` state. Click to see the `:active` state.
- Note there is more padding about the the icon so it's easier to click.
- Take a look at this in Firefox. There shouldn't be a horizontal scroll bar in the sidebar.

Closes #424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Allow for selecting partially locked districts [#420](https://github.com/PublicMapping/districtbuilder/pull/420)
 
 ### Changed
+- Improved lock button UX [#436](https://github.com/PublicMapping/districtbuilder/pull/436)
 - Reduce noise in log output [#399](https://github.com/PublicMapping/districtbuilder/pull/399)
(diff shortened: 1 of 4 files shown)


jfrankl (commented):
This looks great from a visual perspective and the click area feels like the right size. As I'm using it, I'm finding the background-color transition slightly frustrating because it introduces a delay before I can see whether I'm hovering in the right place. When I'm trying to select that small lock button, I find that I want immediate feedback. What do you think?

alexelash:
@jfrankl That makes sense to me. I removed the "hover on" effect, but retained "hover off":

hover-on-off

PR #437, opened 2020-09-21
Show toast for errors

### LINKED ISSUE #188: Show toast message for errors on saving districts
## Overview

There currently isn't a good place to notify users of errors that occur on the map page. When errors are returned from the server (for example, when saving districts), we need a place to show them. This will probably involve some design time, since it isn't currently clear where these errors should be displayed.

## Acceptance criteria

When an error is encountered while saving a map, display a toast message to the user "Something went wrong. Please try again."

maurizi:
## Overview

Shows toast for errors triggered by user actions (e.g. the "Accept" button) on the project screen screen as well as for failed API resource requests throughout the application (e.g. a 502 downloading the region configs, which will leave the "District" dropdown empty).

### Checklist

### Demo

Failure to save project changes:
localhost_3003_projects_9e24a1cc-c583-436e-b371-decc4572c797

Failure to download project data (e.g. `/geojson` endpoint returns a 502):
localhost_3003_projects_d802c294-034b-4d20-96d7-0cd1ed3a7e95

Failure to download region configs:
localhost_3003_create-project

### Notes

@tgilcrest Note the different message used for failures when loading the page "Something went wrong, please refresh the page".

We'd probably prefer to not leave the page in a potentially broken state and instead redirect to a dedicated error page like Github's https://github.com/503.html , but this was easy to do while I was working on this feature and seemed like a good bandaid for now.

## Testing Instructions

- `scripts/update`
- Loading the a page after the webpack server was up but while the Nest server was still coming up was a reliable way for me to test resources failing to load.
- I simply turned off my server or used the Network tab to switch to "Offline" to test error handling for the "Accept" button

Closes #188


tgilcrest:
Awesome, glad we can catch these errors now.

A couple of questions:
* Does the message disappear when you click "Accept" again and the request succeeds? (In the other case, I assume it disappears when you refresh the page)
* Do these messages have any animation or do they just appear on the screen?

One request:
* Could you make the notification wide enough so the "refresh the page" message doesn't break onto 2 lines?

maurizi:
> A couple of questions:
>
>  -  Does the message disappear when you click "Accept" again and the request succeeds? (In the other case, I assume it disappears when you refresh the page)
>  -  Do these messages have any animation or do they just appear on the screen?


No, the messages dissapear after a short timeout, or when you click the "x" button. There is a fade in/out animation.


> Could you make the notification wide enough so the "refresh the page" message doesn't break onto 2 lines?

:+1: will do

tgilcrest:
>No, the messages dissapear after a short timeout, or when you click the "x" button. There is a fade in/out animation.
What happens in the unlikely event that they get two of these in a row? Is it clear that the error occurred again rather than the notification hasn't timed out yet? Could we make it so the notification is removed when you hit "Accept" again, if it's easy? If it's not really easy then we can just leave it as is.

src/client/functions.ts
```diff
@@ -218,3 +219,7 @@ export function mergeGeoUnits(a: GeoUnits, b: GeoUnits): GeoUnits {
     })
   );
 }
+
+export const showActionFailedToast = () => toast.error("Somthing went wrong, please try again.");
```
kshepard, line 223:
```suggestion
export const showActionFailedToast = () => toast.error("Something went wrong, please try again.");
```

src/client/functions.ts
```diff
@@ -218,3 +219,7 @@ export function mergeGeoUnits(a: GeoUnits, b: GeoUnits): GeoUnits {
...
+export const showResourceFailedToast = () =>
+  toast.error("Somthing went wrong, please refresh the page.");
```
kshepard, line 225:
```suggestion
  toast.error("Something went wrong, please refresh the page.");
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Allow for selecting partially locked districts [#420](https://github.com/PublicMapping/districtbuilder/pull/420)
+- Show toast on errors
```
kshepard, line 12:
```suggestion
- Show toast on errors [#437](https://github.com/PublicMapping/districtbuilder/pull/437)
```

kshepard (approved):
Looks good. Probably also worth removing the `toast-demo` page now that we have it integrated.

PR #439, opened 2020-09-22
Add "saved" notification

### LINKED ISSUE #431: Display "Saved" after accepting changes
It would help users to give an indication that their changes are saved after accepting. A possible implementation that could work is to show "Saved" after the spinner disappears on a successful PATCH request.

After hitting Accept, see the spinner:

After the request completes successfully, see "Saved", something like this:

When a new selection is made, remove Saved and show Accept/Cancel:

Unclear what should happens if the user hits cancel. Could see an argument for it either way.

jfrankl:
## Overview

Adds "Saved" notification to the sidebar, which appears when the map reflects what is saved on the server. This reinforces the connection between the "Accept" button and saving, and confirms to the user that their map is saved.

### Checklist

### Demo

Screen Shot 2020-09-22 at 10 12 17 AM

### Notes

@tgilcrest I added tooltips for the "Saved" text as well as the "Cancel" and "Accept" buttons. Could you take a look at that and let me know if you have any edits.

Screen Shot 2020-09-22 at 10 39 47 AM
Screen Shot 2020-09-22 at 10 12 02 AM
Screen Shot 2020-09-22 at 10 11 58 AM

## Testing Instructions

- Open a project
- You should see a "Saved" indicator in the sidebar
- Make a selection on the map; this should remove the "Saved" indicator and show the "Cancel" and "Accept" buttons instead
- Press cancel; the "Saved" indicator should reappear
- Make a new selection on the map; this time click "Accept"; you should see a brief spinner and then the "Saved" indicator should return after the save is complete and the map updates

Closes #431


tgilcrest:
Looks great, and think it will really help communicate what's happening.

Here are my thoughts on the copy. Feel free to take them or leave them.

>Your map is saved to the cloud
>save your map to the cloud

What do you think about removing "to the cloud" in the Saved and Accept tooltips? I'm not sure it adds important details and seems a little buzz-wordy for us/our brand. 

The first one would become "Your map is saved". If you feel like that's too brief or redundant with the text, you could reference "changes" like you do in the other tooltips, e.g. 
>All changes are saved to your map
>All changes to your map are saved
>Your changes have been saved

For Accept/Cancel tooltips, I think I prefer is changing the `:` to `and`. It adds a word but it seems to read a little more fluid to me.

jfrankl:
@tgilcrest Thanks for taking a look. I removed the "cloud" term.

For removing the colon, what do you think of using the word "to" instead of "and"? That seems to fit better to me because it connects an action to the affect, where "and" to me suggests a second action.

<strong>Accept changes</strong> to save your map
<strong>Accept changes</strong> and save your map

I think "Your map is saved" is okay on its own. I'm trying to avoid using the term "changes" for that one because the Saved message will not only appear after a user makes changes, but also when they initially load a totally empty project. In the future, we could potentially enhance this so the "Saved" message doesn't appear until after the first save.

tgilcrest:
"to" sounds good to me.

alexelash (approved):
👍 Works as described. I like the blue spinner 🔵

Issue #428, opened 2020-09-16
Resolve 502 errors generated by ELB


ColeKettler:
## Overview
We're seeing `502 Bad Gateway` errors crop up inconsistently in both staging and production. We need to figure out and address the source of these errors, since they are user-facing.

There are no application logs or Rollbar items being created, and the source of the error is likely infrastructural rather than solely at the application level.

## Supporting Evidence
- Our [CloudWatch logs](https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#metricsV2:graph=~(metrics~(~(~'AWS*2fApplicationELB~'HTTPCode_ELB_502_Count~'LoadBalancer~'app*2falbStagingApp*2f5a0ddc7dc30ea04e)~(~'.~'HTTPCode_Target_5XX_Count~'.~'.))~view~'timeSeries~stacked~false~region~'us-east-1~start~'2020-09-11T16*3a15*3a00.999Z~end~'2020-09-11T16*3a45*3a00.000Z~stat~'Sum~period~60);query=~'*7bAWS*2fApplicationELB*2cLoadBalancer*7d) confirm that the 502s were generated by the ELB, rather than the target application. [This significantly narrows down the things that could have happened](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-502-issues).
- These 502s do not appear to correspond to any other notable events within the application or its infrastructure, and I am unable to reproduce the issue. It definitely happened, I see it in the metrics, but I suspect that there may be a significant non-deterministic aspect to this.
- Nothing seems amiss at all within the NestJS application logs, and healthchecks were passing normally. It appears that the problem was limited to the requests in question.

## Additional Notes
Of all of the possible reasons for the 502 error, a common and subtle offender seems to be issues with the `KeepAlive` timeout differing between the application server and the ELB. There's a little more detail on this [here](https://shuheikagawa.com/blog/2019/04/25/keep-alive-timeout/) and [here](https://medium.com/@liquidgecka/a-tale-of-unexpected-elb-behavior-5281db9e5cb4). We have little direct evidence to support this assumption, but I see enough common factors with Shuhei Kagawa's post (using Express under the hood, putting it behind an ELB, relying on default values) that I think it's possible.

See [this Slack thread](https://azavea.slack.com/archives/C6UD5FG3T/p1599856048300400?thread_ts=1599842809.271200&cid=C6UD5FG3T) for additional context.

PR #441, opened 2020-09-23
Add the ability to query ALB logs with Athena

rbreslow:
## Overview

This PR adds the ability to query ALB logs with Athena by codifying the following AWS guides:

- [Access logs for your Application Load Balancer → Enable access logging](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#enable-access-logging)
- [Querying Application Load Balancer Logs → Create ALB Table](https://docs.aws.amazon.com/athena/latest/ug/application-load-balancer-logs.html#create-alb-table)

Connects #428 

### Checklist

## Testing Instructions

See that the results of this basic query look right (`client_ip` has been redacted).

```sql
SELECT *
FROM stg_alb_logs LIMIT 5;
```

[d2063156-1045-4c1a-86a8-238bef2c1db2.csv.zip](https://github.com/PublicMapping/districtbuilder/files/5270179/d2063156-1045-4c1a-86a8-238bef2c1db2.csv.zip)

### DIFF
--- a/deployment/terraform/app.tf
+++ b/deployment/terraform/app.tf
@@ -30,11 +30,24 @@ resource "aws_lb" "app" {
   security_groups = [aws_security_group.alb.id]
   subnets         = module.vpc.public_subnet_ids
 
+  access_logs {
+    bucket  = aws_s3_bucket.logs.id
+    prefix  = "ALB/APP"
+    enabled = true
+  }
+
   tags = {
     Name        = "alb${var.environment}App"
     Project     = var.project
     Environment = var.environment
   }
+
+  # In order to enable access logging, the ELB service account needs S3 access.
+  # This is a "hidden" dependency that Terraform cannot automatically infer, so
+  # it must be declared explicitly.
+  depends_on = [
+    aws_s3_bucket_policy.alb_access_logging
+  ]
 }
 
 resource "aws_lb_target_group" "app" {
--- a/deployment/terraform/athena.tf
+++ b/deployment/terraform/athena.tf
@@ -1,0 +1,196 @@
+locals {
+  short_environment = var.environment == "Staging" ? "stg" : var.environment == "Production" ? "prd" : lower(var.environment)
+}
+
+# https://docs.aws.amazon.com/athena/latest/ug/application-load-balancer-logs.html
+resource "aws_glue_catalog_table" "alb_logs" {
+  name          = "${local.short_environment}_alb_logs"
+  database_name = "default"
+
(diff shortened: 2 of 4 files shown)


deployment/terraform/iam.tf
```diff
@@ -53,3 +53,27 @@ resource "aws_iam_role_policy" "scoped_email_sending" {
   role   = aws_iam_role.ecs_task_role.name
   policy = data.aws_iam_policy_document.scoped_email_sending.json
 }
+
+data "aws_elb_service_account" "main" {}
```
rbreslow, line 57: This is the same [approach](https://github.com/azavea/raster-foundry-deployment/blob/4ca5f0d0132696b410d5f422e9e98e2e41e06c6f/core/terraform/iam.tf#L270-L279) we take in Raster Foundry. According to the AWS docs, the bucket policy must still reference the `arn:aws:iam::elb-account-id:root` principal.

deployment/terraform/athena.tf
```diff
@@ -0,0 +1,196 @@
+locals {
+  short_environment = var.environment == "Staging" ? "stg" : var.environment == "Production" ? "prd" : lower(var.environment)
+}
```
rbreslow, line 3: Translates to the following pseudocode:

```lua
if environment == "Staging" then
    return "stg"
elseif environment == "Production" then
   return "prd"
else
   return string.lower(environment)
end

File: deployment/terraform/app.tf [hunk 3]
```diff
@@ -30,11 +30,24 @@ resource "aws_lb" "app" {
...
+  }
+
   tags = {
     Name        = "alb${var.environment}App"
     Project     = var.project
     Environment = var.environment
   }
+
+  # In order to enable access logging, the ELB service account needs S3 access.
+  # This is a "hidden" dependedncy that Terraform cannot automatically infer, so
```
hectcastro, line 46:
```suggestion
  # This is a "hidden" dependency that Terraform cannot automatically infer, so
```

deployment/terraform/athena.tf
```diff
@@ -0,0 +1,196 @@
...
+
+# https://docs.aws.amazon.com/athena/latest/ug/application-load-balancer-logs.html
+resource "aws_glue_catalog_table" "alb_logs" {
```
hectcastro, line 6: Nice to see an example of this.

hectcastro (approved):
Table looks good. Nice to have an example of an Athena table defined in Terraform. 👍

PR #442, opened 2020-09-23
Fix problem with highlighted resetting

kshepard:
## Overview

A bug was introduced in PR #420 that caused the set of highlighted geounits to be continuously reset when dragging the rectangle, resulting in the tooltip never showing the aggregated counts of all highlighted geounits.

### Checklist

### Demo

Before:

rectangle-highlight-broken

After: 

rectangle-highlight-fixed

## Testing Instructions

- Interact with the rectangle selection tool and ensure you see aggregated counts while you're dragging


pcaisse:
There are a few other tricky things we should test re: the rectangle selection tool:
* You can switch between selecting counties and blockgroups and it works as before
* Selected blockgroups are correctly "enveloped" when the county containing them are selected

maurizi:
@pcaisse I think you know those details a bit better than me, mind taking this review from me? FWIW the code changes look good to me.

kshepard:
Definitely worth testing those. I'm pretty sure I tested those and more, since I've had this fix on a different branch for the past couple days and haven't run into any issues.

maurizi (approved):

PR #448, opened 2020-09-24
Set keep-alive timeout higher than ALB idle timeout

### LINKED ISSUE #428: Resolve 502 errors generated by ELB

rbreslow:
## Overview

> Keep-alive, when enabled, enables the load balancer to reuse back-end connections until the keep-alive timeout expires. To ensure that the load balancer is responsible for closing the connections to your instance, make sure that the value you set for the HTTP keep-alive time is greater than the idle timeout setting configured for your load balancer.

The [default keep-alive timeout](https://nodejs.org/api/http.html#http_server_keepalivetimeout) for the Node http.Server is 5 seconds. The [default idle timeout](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb#idle_timeout) for the ALB is 60 seconds. 

The ALB is opening connections for reuse, the Node http.Server closes that connection, and then the ALB tries to communicate over the closed connection which triggers a 502.

> **HTTP 502: Bad gateway**
>
> Possible causes:
> . . .
> The target closed the connection with a TCP RST or a TCP FIN while the load balancer had an outstanding request to the target. Check whether the keep-alive duration of the target is shorter than the idle timeout value of the load balancer.

We can see more evidence of this by looking at the ALB access logs with Athena:

<details>   
<summary>Details</summary>    

```sql
SELECT elb_status_code,
         request_processing_time,
         target_processing_time,
         response_processing_time,
         request_url
FROM stg_alb_logs
WHERE elb_status_code LIKE '5%';
```

Athena 2020-09-23 14-35-24

</details>

> `response_processing_time` is set to -1 if the load balancer can't send the request to a target. This can happen if the target closes the connection before the idle timeout or if the client sends a malformed request.

This PR sets the keep-alive timeout to be higher than the ALB timeout so that the ALB is responsible for prompting the closure of individual TCP connections to the server.

See:
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-troubleshooting.html#http-502-issues
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-log-entry-syntax
- https://adamcrowder.net/posts/node-express-api-and-aws-alb-502/ (Excellent deep dive into the problem)

Resolves #428 

### Checklist

## Testing Instructions

I'm not sure how to replicate the failures we saw earlier. I've tried exercising the staging application and wasn't able to generate any 502s today before or after applying the fix (see the [Jenkins deploy](http://urbanappsci.internal.azavea.com/job/PublicMapping/job/districtbuilder/job/test%252Fjrb%252Fadjust-timeouts/2/console)). Additionally, the `prd_alb_logs` table in Athena does not contain any 502s.

I think the evidence makes it clear we should apply this either way, and we can be on the look out for more 502s as we're able to gather more data.

### DIFF
--- a/src/server/src/main.ts
+++ b/src/server/src/main.ts
@@ -17,6 +17,16 @@ async function bootstrap(): Promise<void> {
   );
   app.useGlobalInterceptors(new ClassSerializerInterceptor(app.get(Reflector)));
   app.useGlobalFilters(new BadRequestExceptionFilter());
-  await app.listen(3005);
+
+  // Save the output of 'listen' to a variable, which is a Node http.Server
+  const server = await app.listen(3005);
+
+  // Ensure all inactive connections are terminated by the ALB, by setting this
+  // a few seconds higher than the ALB idle timeout
+  server.keepAliveTimeout = 65000; // eslint-disable-line functional/immutable-data
+
+  // Ensure the headersTimeout is set higher than the keepAliveTimeout due to
+  // this nodejs regression bug: https://github.com/nodejs/node/issues/27363
+  server.headersTimeout = 66000; // eslint-disable-line functional/immutable-data
 }
 bootstrap(); // eslint-disable-line @typescript-eslint/no-floating-promises
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Project page sharing & read-only mode [#449](https://github.com/PublicMapping/districtbuilder/pull/449)
-- Set keep-alive timeout higher than ALB idle timeout[#448](https://github.com/PublicMapping/districtbuilder/pull/448)
+- Set keep-alive timeout higher than ALB idle timeout [#448](https://github.com/PublicMapping/districtbuilder/pull/448)
+- Add basic k6 load test for PA/50 district project [#448](https://github.com/PublicMapping/districtbuilder/pull/448)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Project page sharing & read-only mode [#449](https://github.com/PublicMapping/districtbuilder/pull/449)
+- Set keep-alive timeout higher than ALB idle timeout[#448](https://github.com/PublicMapping/districtbuilder/pull/448)
 
 ### Changed
 
--- a/load-tests/README.md
+++ b/load-tests/README.md
@@ -1,0 +1,58 @@
+# Load Testing
+
+This [k6](https://k6.io/) based load testing project is focused on automating
+user requests that are known to stress parts of the application. All of the
+components necessary to execute a load test are encapsulated in this directory.
+
+## Variables
+
+- `JWT_AUTH_TOKEN`: A DistrictBuilder JWT authentication token
+- `DB_PROJECT_ID`: A DistrictBuilder project UUID (PA; 50 districts)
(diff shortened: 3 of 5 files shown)


src/server/src/main.ts
```diff
@@ -17,6 +17,16 @@ async function bootstrap(): Promise<void> {
...
+  // Save the output of 'listen' to a variable, which is a Node http.Server
+  const server = await app.listen(3005);
+
+  // Ensure all inactive connections are terminated by the ALB, by setting this
+  // a few seconds higher than the ALB idle timeout
+  server.keepAliveTimeout = 65000;
+
+  // Ensure the headersTimeout is set higher than the keepAliveTimeout due to
+  // this nodejs regression bug: https://github.com/nodejs/node/issues/27363
+  server.headersTimeout = 66000;
```
rbreslow, line 30: This was resolved and back ported to Node v12 in https://github.com/nodejs/node/pull/34131. However, it doesn't look like the fix will be accessible until Node v12.18.5 drops (https://github.com/nodejs/node/commits/v12.x. vs https://github.com/nodejs/node/commits/v12.x-staging).

hectcastro (approved):
The application server timeout changes appear to be working as an effective 502 deterrent. I added a simple `k6` load test in 4827e9a32638b1aa6229d81e3387bcc473eaa875 and used it against an instance of the application with and without the changes.

When the timeout changes were not applied, 502s were present in the Athena ALB request logs. When the timeout changes were applied, 502s were not present in the logs. 👍

PR #449, opened 2020-09-25
Share map page

### LINKED ISSUE #126: Publicly shareable map URL
- Generate a URL to the map that can be viewed anonymously, but not edited.

maurizi:
## Overview

This PR makes a few different changes aimed at sharing Maps:
 - The map page can now be accessed for projects that don't belong to the user, or even when not logged in
 - There is a share link in the Map page header to copy the current page URL to the clipboard
 - UX changes were made to the map page for the new read-only mode:
   - The Accept button is hidden and the Cancel button is renamed to Clear
   - The block editing warning modal now appears every time you enter block editing mode, as we can't save to the server that you've already seen it as we do when you're logged in.
     We could store preferences like this in local storage for anonymous users as a follow-on task.
   - Similarly copy changes were made to the block editing modal to reflect that changes can't be persisted

### Checklist

### Demo

Sharing
Peek 2020-09-25 08-12

Read-only mode:
Peek 2020-09-25 08-13

### Notes

I thought about just never showing the block warning modal at all, and automatically clearing changes when switching geounits if you're in read-only mode, but figured we'd want to keep the behavior as close to editing-mode a possible.

## Testing Instructions

- `scripts/server
- Ensure the Map page continues to work normally when editing your own map
- Try going to a different users map while logged in, or going to a map while not logged in at all. It should work but in read-only mode
- Sharing from the link in the header should copy the page URL to the clipboard

Closes #126


kshepard (approved):
Looks great. I agree that a follow-up task would be good to make it so anonymous users only see the block warning once. Will probably be a couple other things as well that I'm wondering about: such as some text/color changes to show we're in read-only mode, and possibly getting rid of the share button when in read-only mode.

PR #452, opened 2020-09-29
Use web workers for real-time demographic updates

### LINKED ISSUE #219: Calculate sidebar updates in real-time as rectangle tool is dragged

kshepard:
## Overview

Calculate sidebar updates in real-time as rectangle tool is dragged, via use of web workers

### Checklist

### Demo

web-workers-demo

### Notes

 * This is based off of #395, and much of the base web worker setup has remained intact
 * Unfortunately, I needed to squash everything here down, since it was a long-running branch that weathered a large amount of PRs, and the constant rebasing and merging wreaked havoc on the commit log. However, I'll add some key notes here for various changes I needed to make.
    * Added short circuit to rectangle selection tool when features are exactly the same
    * ProjectSidebar memoization wasn't working correctly, because `useMemo` wasn't properly able to determine equality of the dependencies. I overrode a comparator here and used lodash's `isEqual`, which worked. But then I ended up instead moving the memoization down to the sidebar row layer, which is what really needed the memoization for performance.
    *  The memoization normalizer for saved selection was looking at the wrong arguments and causing strange behavior
    * The deviation display was broken and was showing all zeros most of the time
    * Increased the set feature throttle delay a bit, since it improves performance and I didn't notice any negative impact from a usage perspective
    * Added a mock for worker to make tests pass 
    * Fixed a bug where a combination of selection, canceling, and re-selecting was causing the map to not update correctly
    * Made a couple changes to eliminate sidebar flickering on save: combined state for the two calculations, and added special handling to not recalculate when no geounits are highlighted or selected
  * We discussed re-architecting the usage of web workers to flow through `redux-loop`. I started digging into it, but became a bit overwhelmed, and wasn't confident that it would even actually fix the flickering problem I was experiencing. I think that may be something we want to explore more down the road, but I'm happy enough with how it's working at the moment that I don't think the additional time is currently warranted.

## Testing Instructions

- `/scripts/update`, `./scripts/server`, and load up a project
- Use the app to build out some districts, and pay attention to the sidebar and tooltip
- Both should be updating in real time as the rectangle tool is dragged
- Verify that the performance while moving the rectangle is acceptable, and is at least as fast as what's currently on staging
- Verify that when saving districts, the sidebar values do not flicker

Closes #219

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,6 +38,7 @@ module.exports = {
           "allowTernary": true,
         }
       ],
+      "react/display-name": "off",
       "@typescript-eslint/ban-ts-comment": "off",
       "@typescript-eslint/explicit-function-return-type": "off",
       "@typescript-eslint/explicit-module-boundary-types": "off",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Project page sharing & read-only mode [#449](https://github.com/PublicMapping/districtbuilder/pull/449)
 - Set keep-alive timeout higher than ALB idle timeout [#448](https://github.com/PublicMapping/districtbuilder/pull/448)
 - Add basic k6 load test for PA/50 district project [#448](https://github.com/PublicMapping/districtbuilder/pull/448)
+- Use web workers for real-time demographic updates [#452](https://github.com/PublicMapping/districtbuilder/pull/452)
 
 ### Changed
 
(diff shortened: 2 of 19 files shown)


maurizi (approved):
🚤

PR #467, opened 2020-10-08
Add microcopy to Create Project fields

### LINKED ISSUE #430: Add microcopy to Create Map page
We got some feedback that the Create Map page is pretty barren, and could use some microcopy to explain what each field means. We should add some to help guide the user.

alexelash:
## Overview
This PR adds some descriptive copy underneath labels on the Create Map page.

### Checklist

### Demo
Screen Shot 2020-10-09 at 8 40 34 AM

### Notes
- I made these descriptions accessible using `aria-describedby` rather than putting them directly in the label and tested with Mac OS's [Voiceover](https://webaim.org/articles/voiceover/) software.
- Fixed an accessibility issue with the radio selects by adding a fieldset container around them and changing their label to `<legend>`.

## Testing Instructions
- `git pull`
- Log in, click "New map" button
- Note new descriptive copy under each label

Closes #430

### DIFF
--- a/src/client/screens/CreateProjectScreen.tsx
+++ b/src/client/screens/CreateProjectScreen.tsx
@@ -147,8 +147,7 @@ const CreateProjectScreen = ({ regionConfigs }: StateProps) => {
     },
     legend: {
       paddingInlineStart: "0",
-      paddingInlineEnd: "0",
-      marginBottom: "0"
+      paddingInlineEnd: "0"
     },
     fieldset: {
       border: "none",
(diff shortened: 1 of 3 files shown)


jfrankl:
@alexelash A couple of things I'm noticing from the screenshot above:

1. The different sections appear to have a different amount of space between the headings and microcopy. I especially see it in the Districts section.
2. Previously, the district options were in two columns, and now they seem to be one per row. Was that intentional?

@tgilcrest could you take a look at the copy for these?

alexelash:
@jfrankl Thanks. Both of these things should be resolved now:

Screen Shot 2020-10-09 at 2 46 22 PM

src/client/screens/CreateProjectScreen.tsx
```diff
@@ -181,6 +221,12 @@ const CreateProjectScreen = ({ regionConfigs }: StateProps) => {
                     Map name
                   </Box>
                 }
+                description={
+                  <Box as="span" sx={style.cardHint}>
+                    e.g. ‘Arizona House of Representatives’. This will distinguish maps from one
+                    another on the landing page.
```
tgilcrest, line 227: The second sentence is a little sterile and refers to the landing page, but they might not know what that is.

I'm imagining something like this, which tries to give friendly advice:
>e.g. ‘Arizona House of Representatives’. Make it specific to help tell your maps apart.

tgilcrest (approved):
Made a suggestion for the Map Name copy. Other than that LGTM!

src/client/screens/CreateProjectScreen.tsx
alexelash: Cool, my latest commit uses your suggestion.

PR #479, opened 2020-10-22
Add support for single county region

### LINKED ISSUE #472: Configure Dane County
We need to prep a data file for Dane County Wisconsin for their custom instance. Same as state-level data, just clipped to Dane County.

AC: 
* Filter out non-Dane County data from Wisconsin
* Configure Dane County custom region
* Attempt to configure without County level (do if minimal, write up further issues if not)
* Figure out what we need to do to support not showing block modal

### LINKED ISSUE #475: The product tour should account for states with different geolevels
Right now, all of the uploaded states have counties, blockgroups, and blocks. When we add a state that does not have these geolevels, we will need to update the product tour's hard-coded references to all three.

For example:
Screen Shot 2020-10-15 at 3 58 09 PM

kshepard:
## Overview

We would like to configure a region for Dane County, WI, and more generally would like to be able to configure single-county regions. This updates the processing script to be able to supply a prefix inclusion filter, to allow only blocks with matching FIPS code prefixes to be processed. Because the region is only a single county, we also don't need to allow selection of the "county" geolevel in the UI (the tour was also updated to reflect this change). Similarly, due to its smaller size, we can safely always display all blocks, so special handling of the base geolevel can be lifted.

### Checklist

### Demo

dane-county

### Notes

I didn't go overboard with logic in attempting to handle every scenario related to when a geolevel needs special handling. The normal case that we were already handling was for when the base geolevel is special (not always visible). I simply added another scenario for when all min zooms are the same, and thus the base geolevel isn't special (it is always visible). Those are the only known cases that we need to support at the moment, so I think it makes sense to maintain that simplicity until we come across a more involved scenario.

## Testing Instructions

- Process Dane County using the new filter flag, e.g.:
    - `mkdir src/manage/data/output-dane-county`
    - `./scripts/manage process-geojson data/input/WI.geojson -o data/output-dane-county -n 4,4 -x 12,12 -l block,blockgroup -f 55025`
- Publish the region (e.g. `./scripts/manage publish-region data/output-dane-county US WI "Dane County"`), create a project with it, and test it out
- Ensure that the text in the tour seems to make sense given the different set of geolevels
- Ensure you can select both blocks and blockgroups at all levels
- Also test another full-state project to make sure nothing has broken (also read through the tour here)
- In case it comes in handy, here's some SQL for resetting the tour and also disabling advanced editing mode
    - `update "user" set "hasSeenTour" = false;`
    - `update "project" set "advancedEditingEnabled" = false;`

Closes #472 
Closes #475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add product tour [#471](https://github.com/PublicMapping/districtbuilder/pull/471)
 - Update data tooling [#468](https://github.com/PublicMapping/districtbuilder/pull/468)
 - Find unassigned menu [#476](https://github.com/PublicMapping/districtbuilder/pull/476)
+- Add support for single county region [#479](https://github.com/PublicMapping/districtbuilder/pull/479)
 
 ### Changed
 
--- a/src/client/components/MapHeader.tsx
+++ b/src/client/components/MapHeader.tsx
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { Flex, Box, Label, Button, jsx, Select, ThemeUIStyleObject } from "theme-ui";
 import { GeoLevelInfo, GeoLevelHierarchy, GeoUnits, IStaticMetadata } from "../../shared/entities";
-import { areAnyGeoUnitsSelected, geoLevelLabel } from "../functions";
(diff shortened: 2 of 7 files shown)


maurizi (approved):
Working great 👍 

BTW - do we need a separate issue to filter Dane County (and other non-states) out from the dropdown?

kshepard:
> BTW - do we need a separate issue to filter Dane County (and other non-states) out from the dropdown?

That's right. We're waiting on some design thoughts on how that will all work before we write up issues.

PR #498, opened 2020-11-04
Allow saving with undo/redo

### LINKED ISSUE #405: Persist Locked Districts
We don't persist locked districts so when you refresh the locked districts disappear. We should persist the locked districts so if you refresh or come back to your map later, we still maintain if they were locked. Save the locked district to the database and show the spinner when we make the call.

pcaisse:
## Overview
This change makes undo/redo work with effects so that we can go backwards and forwards to previously saved map states. 

### Checklist

### Demo
db_undo_redo_saving

### Notes
This was really quite hard to get working... I've added longer comments around the trickier parts of this to try to make things more clear.

The big change in terms of what state we represent is to have both a `state` and optionally `effect` represented to undo/redo saves. We also now need to keep track of `districtsDefinition` in our `UndoableState` to allow for undoing/redoing of saves. Most of the rest of the code is state/effect management in the reducers as you might suspect.

There were several non-obvious things that tripped me up along the way to getting this working, such as how we were always clearing the selected geounits after a save (this worked fine until we wanted to not do that for undo/redo and thus had to be wired up differently for the undo/redo save case). There were several such small-ish changes needed to the reducers to make things shake out correctly, such as having a separate action just for updating `saving` state. Most of these changes are in or around `updateDistrictsDefinition`, so taking a look at that plus the comments I've added there should hopefully make things clear.

I'm a little concerned about the complexity of managing undo/redo state here but I'm not sure how it can really be avoided. Having the notion of a `Cmd` in redux-loop I found helpful as a description of an effect that can be kept lying around in the queue until we really need to apply it. One of the things I did to try to make this less brittle is to modify `pushState` and `replaceState` to work with essentially patches (single discrete changes) and not wholesale replacements (PATCH vs PUT in REST-speak). It seems that in all cases we just want to change one thing at a time so this should be relatively ergonomic (before this change, I was having trouble updating the whole piece of state where I had only meant to update a part of it) 

There is a refactor in this (see 4e307e3) which is to move the map management around undo/redo-ing geounit selection into the map itself. IMHO it's preferable for the reducer to just be responsible for state updates (and effects that relate to state) and then the map should sort itself out so as to look right in response to those state changes.

In closing, I'm glad that this seems to work but would welcome further abstraction/guard rails to make things more ergonomic/less brittle. I'm also open to having horribly over-complicated this so if there are any suggestions to simplify things I'm open to them.

## Testing Instructions
* Select a bunch of geounits
* Save them to a district
* Undo one step and they should be unsaved (the selected geounits should be back assigned to whatever district they were before) and the geounits should still be selected
* Redo and the geounits should be re-saved to the district you originally assigned them to
* Undo all the way back and your changes should be completely undone
* Redo all the way back and your changes should be completely redone
* Undo partially, select some other geounit, and then save again and it should work as expected

Closes #405

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -30,7 +30,8 @@ import {
   levelToLineLayerId,
   onlyUnlockedGeoUnits,
   getChildGeoUnits,
-  DISTRICTS_OUTLINE_LAYER_ID
+  DISTRICTS_OUTLINE_LAYER_ID,
+  setFeaturesSelectedFromGeoUnits
 } from "./index";
 import DefaultSelectionTool from "./DefaultSelectionTool";
 import FindMenu from "./FindMenu";
@@ -244,6 +245,25 @@ const DistrictsMap = ({
       );
     }
   }, [map, staticMetadata, geoLevelIndex]);
+
+  // Keep track of when selected geounits change
+  const prevSelectedGeoUnitsRef = useRef<typeof selectedGeounits | undefined>();
+  useEffect(() => {
+    prevSelectedGeoUnitsRef.current = selectedGeounits; // eslint-disable-line
+  });
+  const prevSelectedGeoUnits = prevSelectedGeoUnitsRef.current;
+
+  // Update map when selected geounits change.
+  // Typically the geounits change as a result of using the selection tools directly -- map is
+  // changed and then that needs to be reflected in state -- but this accounts for undo/redo actions
+  // affecting state which then needs to be reflected in the map.
+  useEffect(() => {
+    // eslint-disable-next-line
+    if (map) {
+      prevSelectedGeoUnits && setFeaturesSelectedFromGeoUnits(map, prevSelectedGeoUnits, false);
+      selectedGeounits && setFeaturesSelectedFromGeoUnits(map, selectedGeounits, true);
+    }
+  }, [map, selectedGeounits, prevSelectedGeoUnits]);
 
   // Keep track of when selected geolevel changes
   const prevGeoLevelIndexRef = useRef<typeof geoLevelIndex | undefined>();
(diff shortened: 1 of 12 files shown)


src/client/components/map/Map.tsx
```diff
@@ -245,6 +246,25 @@ const DistrictsMap = ({
...
+  // Typically the geounits change as a result of using the selection tools directly -- map is
+  // changed and then that needs to be reflected in state -- but this accounts for undo/redo actions
+  // affecting state which then needs to be reflected in the map.
+  useEffect(() => {
+    // eslint-disable-next-line
+    if (map) {
+      prevSelectedGeoUnits && setFeaturesSelectedFromGeoUnits(map, prevSelectedGeoUnits, false);
+      selectedGeounits && setFeaturesSelectedFromGeoUnits(map, selectedGeounits, true);
+    }
+  }, [map, selectedGeounits, prevSelectedGeoUnits]);
```
maurizi, line 266: I like this, this is much cleaner than the previous method of doing this in a Loop side-effect 🎉

maurizi (changes requested):
Overall I think this is a good approach to this problem.

I am however encountering something that seems like it could be an issue, or at least unintended - once an `effect` is in place in the Redux store, `PATCH` calls are issued for _every_ undo/redo, even the ones that are just staging changes not persisting them.

Watch the network tab in this GIF to see what I mean:
d600644e-bbe3-46c7-ac08-60ec5e2afb35

pcaisse:
@maurizi Okay, I think this is ready for another look. The functional change I made to address the issue was to not preserve the effect when calling `pushStateUpdate`.

I also went ahead and broke the undo/redo specific functionality out into its own module which I think should be a nice guardrail/affordance from a developer perspective in terms of always using the helpers to modify undo history to ensure it's updated in the right way.

maurizi (approved):
🎉

PR #500, opened 2020-11-12
Fix region config validation error

### LINKED ISSUE #494: Error creating map in staging
## Description
When creating a new map, an error is displayed: `property regionConfig should not exist`

## Steps To Reproduce
1. Go to the create map page
2. Select a region
3. Click create map button
4. See error: `property regionConfig should not exist`

## Expected behavior
Map should be created

## Actual behavior
An error is displayed

## Screenshots
property-region-config

## Your environment
 - OS: Ubuntu 18.04
 - Browser: Firefox
 - Version: 81

## Additional context
Most likely caused by this commit: https://github.com/PublicMapping/districtbuilder/commit/f9238245471b413b81b724dc38cd35590056226b

kshepard:
## Overview

There was an error when creating a new map: `property regionConfig should not exist`. This was happening, because `forbidNonWhitelisted` validation was turned on in this commit: https://github.com/PublicMapping/districtbuilder/commit/f9238245471b413b81b724dc38cd35590056226b, but the `regionConfig` property didn't have a validation decorator, so wasn't whitelisted. The fix implemented here adds a simple validation decorator to `regionConfig`, just verifying that it's present.

### Checklist

## Testing Instructions

- Attempt to create a new map. This wasn't working beforehand, but should succeed now.

Closes #494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix region config validation error [#500](https://github.com/PublicMapping/districtbuilder/pull/500)
+
 ## [1.1.0] - 2020-10-08
 
 ### Added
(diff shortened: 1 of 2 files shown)


pcaisse (approved):
Works well!

Does this mean that if we remove any of the validation around project creation it will break project creation though?

maurizi:
> Does this mean that if we remove any of the validation around project creation it will break project creation though?

Yes, however, there is a decorator whose only purpose is whitelisting, if we need it: https://github.com/typestack/class-validator/blob/99eec62eb9a80a5e8495fcd151ac636d5812ce90/src/decorator/common/Allow.ts#L10

PR #525, opened 2020-11-27
Persist locking districts

### LINKED ISSUE #405: Persist Locked Districts

maurizi:
## Overview

This persists locking districts to the backend, including as part of undo/redo

### Checklist

### Demo

Saving:
22de1df9-bf61-4176-bd02-aad38136b7cf

Undo/redo
52cf93eb-6498-40ea-b6e1-4173546d83a8

## Testing Instructions

- `scripts/server`

Closes #405

### DIFF
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -154,7 +154,7 @@ export class ProjectsController implements CrudController<Project> {
       return this.service.createOne(req, {
         ...dto,
         districtsDefinition: new Array(geoCollection.hierarchy.length).fill(0),
-        lockedDistricts: new Array(geoCollection.hierarchy.length).fill(false),
+        lockedDistricts: new Array(dto.numberOfDistricts).fill(false),
         user: req.parsed.authPersist.userId
       });
     } catch (error) {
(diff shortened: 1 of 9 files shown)


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -154,6 +154,7 @@ export class ProjectsController implements CrudController<Project> {
       return this.service.createOne(req, {
         ...dto,
         districtsDefinition: new Array(geoCollection.hierarchy.length).fill(0),
+        lockedDistricts: new Array(geoCollection.hierarchy.length).fill(false),
```
pcaisse, line 157: I was getting a 400 when trying to save locked districts and it seems like it's due to us creating projects with the wrong number of locked districts.

After making this change everything seems to test well:
```suggestion
        lockedDistricts: new Array(dto.numberOfDistricts).fill(false),
```

pcaisse (changes requested):
Looks good! Just one small change needed.

pcaisse (approved):

PR #526, opened 2020-12-01
Allow copying maps

### LINKED ISSUE #493: Copy Map as Template
If a user is viewing a map (either a shared map or their own), we want to enable them to copy the map as a template and start a new map from it. This will create a copy of the map in the user's account so they can begin editing the map.

AC:
* There is a button on maps for "Copy Map"
* A modal pops up with a field to enter the new name of the map
* When the user confirms the name, the map is copied to the users account and the user is redirected to begin editing the map
* If the user is not authenticated, the modal instead prompts them to register with Name, Email, Password.
* If the user registers, we display a success message with Toast and the modal to copy map

### LINKED ISSUE #497: Authentication modal to prompt unauthenticated users
Some places in the app are accessible to users without needing to be authenticated. However, it should be possible for the users to then access places that require authentication, so we want to provide a path to authenticate these users. For example, if an unauthenticated user follows the link in #496, they will need to be authenticated to create a map.

Rather than redirect a user to the authentication page and back, we want to use a modal that will authenticate them in place to simplify things. This modal will contain the same forms on the registration and login pages, it will just allow the user to select which one they want based on whether they have an account or not.

Note: In the case of the link in #496 we'll need to send the user to a blank page with the modal first before then sending them into the map page.

AC:
1. Authentication modal that we can present when the user wants to take an action when they are authenticated.
2. The modal will include fields for account registration
3. If the user already has an account, the modal will offer to Login instead, and use login fields
4. When the user successfully registers / logs in, they are routed to the place they intended to go
5. Any unsuccessful login attempts are handled gracefully with an error message and don't interrupt the flow into the authenticated resource

kshepard:
## Overview

If a user is viewing a map (either a shared map or their own), we want to enable them to copy the map as a template and start a new map from it. This will create a copy of the map in the user's account so they can begin editing the map. This also takes into account user login/registration and will ensure a user is logged in before allowing copying.

### Checklist

### Demo

copy-map-01-fixed
copy-map-02
copy-map-03
copy-map-04

### Notes

 * I based this off the wireframes, which had some placeholder copy. There will need to be a design pass to flesh this out, as well as tweak styling as needed.
 * The acceptance criteria on the issue is a little out-of-date. Read all the comments, and see the wireframes for the full picture. I believe I've addressed everything.
 * Jeff requested that we build the auth modal in such a way that it'd be possible to update the header content/etc. independently from that of the login/registration pages. That is why I took the approach I did here: allowing the ability to pass arbitrary child nodes for rendering, and abstracting out only the bare minimum required fields/functionality from the login/registration pages. This allows keeping things flexible while also eliminating the need for copy/pasted code.
 * There was no mention of adding the password reset functionality to the modal, and in the interest of time I didn't do so. This will be straightforward to implement if we'd like to do so, however.

## Testing Instructions

- `./scripts/server`
- Log in, view a map, and ensure you're able to make a copy of it
- Grab the url for the map, log out, and view the read-only version
- Ensure you're able to make a copy of it (along the way, testing both the login and register functionality of the modal)
- Ensure the login and register functionality of the non-modal pages still work

Closes #493 
Closes #497


src/server/src/projects/entities/create-project.dto.ts
```diff
@@ -11,4 +11,6 @@ export class CreateProjectDto implements CreateProjectData {
   readonly numberOfDistricts: number;
   @IsNotEmpty({ message: "Need to supply a region configuration" })
   readonly regionConfig: RegionConfigIdDto;
+  @Allow()
```
maurizi, line 14: The validators on `update-project.dto.ts` should be able to be used here (and due to `@IsOptional` will be ignored if the `districtsDefinition` isn't provided)
```suggestion
  @IsArray()
  @ArrayNotEmpty()
  @IsOptional()
```

kshepard: Nice, I didn't know about `@IsOptional`. Makes sense.

maurizi (approved):
Looking good 🎉

PR #528, opened 2020-12-07
Reduce problem with hidden base geounits

### LINKED ISSUE #492: Hidden/invisible blocks that aren't selectable
## Steps To Reproduce
1. Create a new Dane County map in staging
2. Use the rectangle tool to select all blocks and assign them to a district
3. Notice that there is still a population of 122 in the unassigned district
4. Use the find unassigned tool, and notice that there are 48 unassigned blocks, but none appear to actually be visible  

## Expected behavior
Using the rectangle tool to select all blocks in Dane County should not result in any unassigned population/areas

## Actual behavior
Using the rectangle tool to select all blocks in Dane County does result in unassigned population/(invisible) areas

## Screenshots
dane-county-unassigned

## Your environment
 - OS: Ubuntu 18.04
 - Browser: Firefox
 - Version: 81 

## Additional context
This hasn't show itself in other regions yet, because I think it would be pretty hard to come across in practice. It's easy to see here, because we removed the block zoom restriction in Dane County and are able to select way more blocks than is normally allowed. Since we're allowed to select all of them, we're able to easily expose this issue, where it could go unnoticed otherwise. This does have the potential to cause a problem in all regions though, so we should investigate and see what the underlying problem may be.

kshepard:
## Overview

Very small geounits are not rendered to the map at all zoom levels in MapboxGL. Some geounits can get so small that they aren't rendered even at the maximum zoom level we had in place. The max zoom level has been increased here to allow users to select geounits that would previously have been hidden. 

These small geounits may be present in a region even without performing any simplification. However, they are much more prevalent when simplification is present; particularly around bendy bodies of water. Simplification has been reduced to minimize the presence of these small geounits. As a side effect, the static data file sizes have increased. Quantization has been introduced, which decreases the size of these files, making them even a little smaller than they were before the simplification change. 

This PR also includes a variety of minor fixes that were found while investigating this issue.

### Checklist

### Demo

tiny-blocks

### Notes

Some additional changes:
 * Fixed a problem with the find unassigned functionality that was causing the page to crash after fixing the final unassigned polygon
 * Type information was missing for one field. This is needed by the `manage` commands, and has been added
 * Intermediate files were being uploaded to S3, even though they would never be used (GeoJSON and MB tiles files). They are now being filtered out to reduce S3 storage and speed up publishing times.
 * A little meta, but updated the `manage` README with an easier command to run in order to update the `manage` README.

Other notes:
 * When zooming in as far as is done here, some shape consistency starts to break down. You'll notice in the GIF above that that triangular patch of land doesn't exactly match up between it's GeoJSON and vector tiles. This doesn't appear to be related to either simplification or quantization. To test, I published a region that has no simplification or quantization, and the same mismatch exists. I don't think this will be much of a problem in practice, it just looks a little odd, and I don't think there's any good way to improve it.
 * Just to provide some raw numbers for what changes with the new simplification/quantization in place: for Dane county, we originally were looking at a TopoJSON file that was 7.9MB, and when selecting all blocks from the default zoom, we'd have 48 unassigned areas totaling a population of 122. Decreasing the simplification to the amount here ends up increasing the TopoJSON file size to 9.5MB, but reduces the unassigned areas to 1, with a 0 population. Adding quantization reduces this filesize to 7.3MB. I tried varying levels of simplification/quantization on both Dane County and Pennsylvania, and this combination seemed the most optimal.
 * For Dane County, if we would like to eliminate this problem for good, we can change it so blocks are displayed at zoom level 10, instead of at all zoom levels. This is easy to do, the only downside is the slight friction for users (the enable blocks dialog, and not being able to select them at all zoom levels).
 * To more obviously see the problem of MapboxGL not rendering small geounits, you can manually change the minimium zoom on the client side, and then zoom way out and use the rectangle tool to select all blocks. If you do this, you will see a large amount of blocks that have not been selected.

## Testing Instructions

- Process and publish Dane County:
```
./scripts/manage process-geojson data/input/WI.geojson -o data/output-dane-county -n 4,4 -x 12,12 -l block,blockgroup -f 55025
./scripts/manage publish-region data/output-dane-county US WI "Dane County"
```
- Create a new map with that region, use the rectangle tool to select all blocks and assign them to a district
- Use the find unassigned tool to fix the one unassigned block

Closes #492


maurizi (approved):
Good sleuthing! 👍

PR #531, opened 2020-12-07
Add find non-contiguous option to find menu

### LINKED ISSUE #517: Find Non-Contiguous Areas
Similar to find unassigned, users should be able to find pieces of their districts that are non-contiguous. This will function just as Find Unassigned does. The pop-down menu should be updated to accommodate this additional function. When the user clicks Find, the pop-down menu will have a selector to choose Unassigned or Non-contiguous. The system should switch to finding the other type of area when the user switches the option.

AC:
* When the user clicks the find button, the pop-down menu has an option to choose between Unassigned and Non-contiguous areas
* When find Non-contiguous is selected, the system will highlight any areas of districts that are non-contiguous from other parts of the district
* When the user fixes a non-contiguous part is removed from the count, the way Unassigned works

Note: Since we won't know which part of the district is the "main" part, we'll just need to highlight every part of a multi-part district. This means it will never be possible to find 1 non-contiguous part. There will always be at least two parts of any non-contiguous district.

maurizi:
## Overview

Add find non-contiguous option to find menu

### Checklist

### Demo

localhost_3003_

localhost_3003_ (1)

## Testing Instructions

- `scripts/server`

Closes #517


kshepard (approved):
Nice, looks good to me.

PR #532, opened 2020-12-07
Update email verification copy

### LINKED ISSUE #518: Update Email Verification Copy
Remove mention of 14 days from the email verification header.

kshepard:
## Overview

There was a mention of "14 days" in the email verification copy, but it's not tied to any logic and isn't actually the case, so we're removing it. 

### Checklist

- ~~[ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible~~

### Demo

update-email-configm-copy

## Testing Instructions

- Log in with a user that has an unverified email
- On the home page, notice that the copy no longer refers to the "14 days" time frame.

Closes #518


maurizi (commented):
Looks good to me 👍

maurizi (approved):

PR #538, opened 2020-12-11
Don't show district lock status on read-only maps

### LINKED ISSUE #529: Don't show district lock status on read-only maps
## Overview

The state of a district being locked is probably not relevant to a user viewing a read-only version of the map. Currently the lock icons are showing up, and when clicked will trigger an error toast message.

## Acceptance criteria

Locked district icons are not displayed on the read-only view of maps.

kshepard:
## Overview

District lock status was showing up in the sidebar and the map in read-only mode, where it wasn't desired.

### Checklist

### Demo

Read only:
lock-not-logged-in

Owner view:
lock-logged-in

### Notes

The sidebar was straightforward, but getting the map layer to not show the locked status was a little trickier than expected. The least invasive and future-proof approach seemed to be stripping the data at the source: the Project `getOne` function in the controller. However in order to get this to work, I needed to add a new Auth guard that would allow user information to be available in that function if present. This is because using the standard Auth guard will deny unauthenticated requests, and using no auth guard will not pass along any user information to the function. Once that was worked out, the new logic is simply removing the locking information from the project when it's not being requested by the project owner. As our auth story grows, if it turns out we need to do more things of this nature, I can see it being worth getting fancier with types here: for instance having a separate type for a ReadOnlyProject that doesn't even include certain fields, such as locked districts. This type of change seems overkill at this point though, so I just kept it simple.

## Testing Instructions

- `./scripts/server`
- View a project that you own and ensure you can lock/unlock districts and they show as such on the map
- View the same project in a private tab (not logged in) and ensure lock information doesn't show up on the sidebar/map
- Log in as a second user, view the same map, and ensure you also see the same read-only locking behavior with this user

Closes #529

### DIFF
--- a/src/server/src/projects/entities/project.entity.ts
+++ b/src/server/src/projects/entities/project.entity.ts
@@ -58,4 +58,9 @@ export class Project implements IProject {
     default: "'{}'"
   })
   lockedDistricts: readonly boolean[];
+
+  // Strips out data that we don't want to have available in the read-only view in the UI
+  getReadOnlyView(): Project {
+    return { ...this, lockedDistricts: new Array(this.numberOfDistricts).fill(false) };
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Don't show district lock status on read-only maps [#538](https://github.com/PublicMapping/districtbuilder/pull/538)
+
 ## [1.2.0] - 2020-11-18
 
 ### Added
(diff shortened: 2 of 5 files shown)


src/server/src/projects/entities/project.entity.ts
```diff
@@ -43,4 +44,11 @@ export class Project implements IProject {
     default: "'{}'"
   })
   lockedDistricts: readonly boolean[];
+
+  // Strips out data that we don't want to have available in the read-only view in the UI
+  getReadOnlyView(): Project {
+    const mutableRoProject = clone(this);
+    mutableRoProject.lockedDistricts = [];
+    return mutableRoProject;
```
maurizi, line 52: Using the `...` spread operator can eliminate the use of `lodash` here. Also, consider returning a `lockedDistricts` array w/ a valid number of (false) values.
```suggestion
    return { ...this, lockedDistricts: new Array(this.numberOfDistricts).fill(false) }
```

maurizi (approved):
I left one comment, but I don't feel super strongly about that suggestion, this looks good as-is.

src/server/src/projects/entities/project.entity.ts
kshepard: That is nicer, thanks!

PR #544, opened 2020-12-17
Various minor changes to processing script

kshepard:
## Overview

Made a few small changes while reprocessing all regions for less simplification + quantization:

 * Increased max Node memory by a couple GBs, since Texas hit the limit
 * Fixed quantization parameter short-name
 * Allow bypassing quanitzation by supplying a value of zero
 * Use streaming to write GeoJSON, since Texas hit the JS string length limit

### Checklist

- ~~[ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible~~

## Testing Instructions

- I don't think this needs a whole lot of testing, since I've just processed all regions. I'm mostly looking for a code spot check, and if optionally process a small region.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@ services:
       - POSTGRES_PASSWORD=districtbuilder
       - POSTGRES_DB=districtbuilder
       - NODE_ENV=Development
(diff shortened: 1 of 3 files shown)


maurizi (approved):
Looks good to me, didn't test

PR #545, opened 2020-12-17
Remove "Copy this map" button from owned maps

### LINKED ISSUE #530: Remove "Copy this map" from map view
When editing your own map, we should hide the "Copy this map" button. The "Copy this map" button should only appeal when viewing a shared map of a different user. I want to try to keep the header as simple as possible, focusing on operations that relate to editing or working with a map.

There are wireframes for a "Duplicate" function, which is similar to "Copy this map", but meant for users to duplicate their maps. It is located on the map list view, as opposed to the map editor. Designs on Abstract: https://share.goabstract.com/aafa38b8-1ec7-4c12-a653-29aa85872440

kshepard:
## Overview

We want to remove the "Copy this map" button from owned maps in order to unclutter the UI. We'll eventually have this functionality available from the map list page as a "Duplicate" function.

### Checklist

### Demo

screenshot_12

## Testing Instructions

- Load a map that you own, and ensure you don't see a "Copy this map" button
- Open this same map in a private/unlogged-in window and verify that the button is displayed

Closes #530

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Remove "Copy this map" button from owned maps [#545](https://github.com/PublicMapping/districtbuilder/pull/545)
+
 ### Fixed
 
 - Don't show district lock status on read-only maps [#538](https://github.com/PublicMapping/districtbuilder/pull/538)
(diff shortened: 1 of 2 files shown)


pcaisse (approved):
Works like a dream! :cloud: 

I noticed that there's still placeholder copy in the "Copy this map" modal:
db_lorem_ipsum

I suspect you're aware but I figured I might as well mention in in case it was an oversight.

kshepard:
Thanks for checking, but yeah, there's an open issue to add new copy.

PR #546, opened 2020-12-18
Fix health check when unable to load RegionConfig data

### LINKED ISSUE #515: Ensure all errors are properly handled on startup
## Overview

Note: I haven't attempted to reproduce this problem yet, and don't know the exact cause, but we saw it take place on both staging and production. The error occurred after adding a new field to the `region_config` data model. After a deploy, and before migrations were run, the service seemingly comes up just fine, and responds positively to health checks. However, the memory usage is at close to 0%, and users are unable to load or create maps. In both staging and production, we were able to fix this issue by running migrations and then cycling the service.

I speculate that the problem most likely has to do with error handling during startup. I'm guessing that an error is thrown when the (pre-migrated) data isn't matching up with what's expected, and from the outside, it looks healthy even though the data hasn't been loaded. 

## Acceptance Criteria

The issue described above is able to be reproduced locally and fixed

maurizi:
## Overview

During a previous deployment, we encountered an issue where a change to the `RegionConfig` entity caused the service to be unable to load the topojson layers, but the healthcheck did not report a problem.

By adding a new field to the `RegionConfig` entity, I was able to recreate the original issue:

```diff
diff --git a/src/server/src/region-configs/entities/region-config.entity.ts b/src/server/src/region-configs/entities/region-config.entity.ts
index 8cba8b3..8636751 100644
--- a/src/server/src/region-configs/entities/region-config.entity.ts
+++ b/src/server/src/region-configs/entities/region-config.entity.ts
@@ -32,4 +32,7 @@ export class RegionConfig implements IRegionConfig {
 
   @Column({ type: "boolean", default: false })
   hidden: boolean;
+
+  @Column({ type: "boolean", default: false })
+  test: boolean;
 }
 ```
 
 From what I can tell, because the `_layers` field of TopologyService defaulted to `[]`, when it encountered an error in retrieving the regionConfigs the health check thought it was healthy bc it had no regions configured.
 
 Since the regions are loaded on startup, they didn't throw an error in the healthcheck itself, but are instead an unhandled error that only gets logged.

### Checklist

## Notes

I also updated the topology health check indicator while I was at it, because it is a little unwieldy at this point to report every state we have configured. Instead I only list the states that are still pending, which should be enough to identify any problematic layers that are taking too long to load.

## Testing Instructions

Use the above diff to recreate the health check problem on develop. Verify on this branch that the health check correctly returns 503 if it fails to / has not yet loaded the region configs.

Closes #515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Don't show district lock status on read-only maps [#538](https://github.com/PublicMapping/districtbuilder/pull/538)
+- Fix topology health check to report inability to load region configs as an error [#546](https://github.com/PublicMapping/districtbuilder/pull/546)
 
 ## [1.2.0] - 2020-11-18
 
--- a/src/server/src/healthcheck/topology-loaded.indicator.ts
+++ b/src/server/src/healthcheck/topology-loaded.indicator.ts
@@ -27,9 +27,14 @@ export default class TopologyLoadedIndicator extends HealthIndicator {
         });
       })
     )) as [string, boolean][];
-    const layerStatus = Object.fromEntries(layerEntries);
-    const isHealthy = Object.values(layerStatus).every(status => status);
-    const result = this.getStatus(key, isHealthy, layerStatus);
+
+    const isHealthy = layerEntries.every(([_, status]) => status);
+    const pendingLayers = layerEntries.filter(([_, status]) => !status).map(([layer, _]) => layer);
+    const result = this.getStatus(key, isHealthy, {
+      total: layerEntries.length,
+      complete: layerEntries.length - pendingLayers.length,
+      pendingLayers
+    });
 
     if (isHealthy) {
       return result;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/server/src/healthcheck/topology-loaded.indicator.ts [hunk 1]
```diff
@@ -22,9 +27,14 @@ export default class TopologyLoadedIndicator extends HealthIndicator {
         });
       })
     )) as [string, boolean][];
-    const layerStatus = Object.fromEntries(layerEntries);
-    const isHealthy = Object.values(layerStatus).every(status => status);
-    const result = this.getStatus(key, isHealthy, layerStatus);
+
+    const isHealthy = layerEntries.every(status => status);
```
kshepard, line 31: `isHealthy` was always returning `true`, since it was just checking the truthiness of the tuple itself. This should do the trick:

```suggestion
    const isHealthy = layerEntries.every(([_, status]) => status);
```

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Reduce problem with hidden base geounits [#528](https://github.com/PublicMapping/districtbuilder/pull/528)
 - Find non-contiguous [#531](https://github.com/PublicMapping/districtbuilder/pull/531)
 - Add last updated date to map list [#541](https://github.com/PublicMapping/districtbuilder/pull/541)
+- Fix topology health check to report inability to load region configs as an error [#546](https://github.com/PublicMapping/districtbuilder/pull/546)
```
kshepard, line 17: Very minor, but placing this under the `Fixed` section would probably be more appropriate.

kshepard (approved):
Looks great. I added a suggestion for one problem I noticed with the health check. I love the new succinct health check status message, that will be very helpful.

PR #547, opened 2020-12-21
Delete maps from home screen

### LINKED ISSUE #120: Delete a project
On the project page, we want to give users the ability to clean up previous maps they have made, so we want to give them a button to delete a project. In this case, we won't actually the delete the project but rather just archive it (set an archived flag so it doesn't show up).

The Delete map option [will be presented as an option in the flyout](https://share.goabstract.com/aafa38b8-1ec7-4c12-a653-29aa85872440?collectionLayerId=e9734580-7734-47b4-84a4-bf2cbf9fe13f&mode=design) referenced for Duplicate map in #530. 

AC:
1. For each map on the map list page there is a flyout menu that contains a "Delete map" option. 
2. The "Delete map" function doesn't actually delete the map, but archives the map so it does not show up in the list
3. Since we don't have a way to un-archive the maps, there should be a confirmation modal with:

>#### Delete Map
>Are you sure you want to delete [map name]? This can't be undone.
>[Cancel]      [Delete]

*Delete button is red. Cancel button is neutral

maurizi:
## Overview

Adds option to project list flyout component to delete that map, after confirming with a modal.

Projects are not actually deleted, but marked with an `archived` flag, in anticipation of a potential unarchive feature.

### Checklist

### Demo

localhost_3003_ (2)

## Testing Instructions

- `scripts/update`
- Delete a project from the home screen

Closes #120

### DIFF
--- a/src/client/components/DeleteProjectModal.tsx
+++ b/src/client/components/DeleteProjectModal.tsx
@@ -1,0 +1,104 @@
+/** @jsx jsx */
+import AriaModal from "react-aria-modal";
+import { connect } from "react-redux";
+import { Box, Button, Flex, Heading, jsx, ThemeUIStyleObject } from "theme-ui";
+
+import Icon from "./Icon";
+import { IProject } from "../../shared/entities";
+import { projectArchive, setDeleteProject } from "../actions/projects";
+import { State } from "../reducers";
+import store from "../store";
+
+const style: ThemeUIStyleObject = {
+  modal: {
+    bg: "muted",
+    p: 3,
+    width: "small",
+    maxWidth: "90vw"
+  },
+  header: {
+    bg: "error",
+    padding: "16px 12px",
+    margin: "-12px -12px 24px"
+  },
+  footer: {
+    flex: "auto",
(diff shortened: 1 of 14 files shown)


src/client/components/DeleteProjectModal.tsx
```diff
@@ -0,0 +1,104 @@
...
+          <Button
+            id="cancel-delete-project"
+            onClick={hideModal}
+            sx={{ variant: "buttons.secondary", mr: 2 }}
+          >
+            Cancel
+          </Button>
+          <Button
+            sx={{ variant: "buttons.danger" }}
+            disabled={isPending}
```
kshepard, line 83: `isPending` will never be set to `true`, because there's both a `deleteProjectPending` and an `archiveProjectPending` that seem to mistakenly be used interchangeably. Also, I see `deleteProjectPending` is defined as a `boolean | undefined`, but probably makes more sense as just a boolean, since there should only be two possibly states, I think.

kshepard (approved):
Looks great. I pointed out one minor problem, but good to merge after that's addressed.

PR #552, opened 2021-01-11
Adjust state dropdown copy to direct to newsletter signups

### LINKED ISSUE #549: Change CTA on State Dropdown
We get a bunch of requests for new states in our email. I think this because we ask users to let us know when creating a map:

Rather than send us an email, we should instead have people sign up for the mailing list so they know when new states (and features) will be released. So we should change this text and also update the mailing list sign up to capture additional info:

1. Update the text on the state dropdown to say (something like this--open to suggestions):
>What state do you want to map? If you don’t see it in the list, [sign up for our mailing list](districtbuilder.us1.list-manage.com/subscribe?u=61da999c9897859f1c1fff262&id=70fdf1ae35) to know when new states are available!
2. Add a new field to our [mailing list signup](https://districtbuilder.us1.list-manage.com/subscribe?u=61da999c9897859f1c1fff262&id=70fdf1ae35) to capture additional info (free text):
>What state do you want to map, or what feature is especially important to you?

I'm not sure where this would be captured exactly because I'm not sure how the mailchimp form works, but it would be great to capture this info somewhere as we build our list.

alexelash:
## Overview
We're getting a lot of emails from people who don't see their state in the dropdown (which is helpful for prioritizing). Typically, we direct them to sign up for the DB newsletter in order to get an update when their state is added. This PR places that signup earlier in this process so that people know what to expect.

### Checklist

### Demo
Screen Shot 2021-01-11 at 1 11 01 PM

### Notes
- This is what the Mailchimp form looks like now:
Screen Shot 2021-01-11 at 1 16 29 PM

## Testing Instructions
- `git pull`
- Create a new map
- Note the new helper copy for the "State" field. Click the link. It will open the newsletter signup in another tab, which now has a new text input.

Closes #549


tgilcrest:
LGTM!

jfrankl (commented):
The copy in the app looks good to me. 

I have a concern about the added field on the Mailchimp signup form, "Don't see your state? Tell us which to add." I suspect that will work well if a user is coming from the link in the New Map workflow. However, I think it is potentially confusing if you are entering the form from a different context. I'm picturing a user who maybe isn't making maps, but wants to track the DistrictBuilder project because they just read about it in an article about citizen-made maps. I think the question would be potentially confusing to them.

A few possible directions:
- Can we have a different MailChimp form for this one particular link, which would have this extra question?
- Can we make the question more general, so it would make sense to anyone? For example, "What state(s) are the most important to you around redistricting?"

alexelash:
After some discussion with @jfrankl, we edited the copy in the Mailchimp form to make this question more general:
Screen Shot 2021-01-12 at 8 34 16 PM

jfrankl (approved):
The new sign up copy looks good to me

PR #556, opened 2021-01-14
Rewrite null values in SHP export to prevent them from being dropped

### LINKED ISSUE #554: Compactness Not Exported in SHP Export
### Steps to reproduce:
1. Load a map with districts
2. Download a shapefile of the map
3. Open in GIS software e.g. QGIS, open attributes

### Observed Result
The `compactnes` field correctly is populated with "non-contiguous" for non-contiguous districts, but there otherwise the values is NULL.

### Expected Result
The `compactnes` field is populated with a value for the compactness when the district is contiguous.

maurizi:
~Fixes issue of the compactness field being dropped, which seems to have happened whenever the first district had a null value.~

~This manifested in compactness scores being dropped, but I think would be an issue for any nullable field.~

Edit:
The initial approach didn't work out well, and was reworked in https://github.com/PublicMapping/districtbuilder/pull/556/commits/1e3e09130376c6d889535870000e50b4ee21569f to make the `compactness` field numeric-only and add a new `contiguity` field in the export.

### Checklist

## Testing Instructions

- Load up a map with no unassigned blocks
- On `develop` the Shapefile export should not have a `compactness` column - on this branch it should

Closes #554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Added Shapefile export [#516](https://github.com/PublicMapping/districtbuilder/pull/516)
+- Added Shapefile export [#516](https://github.com/PublicMapping/districtbuilder/pull/516) & [#556](https://github.com/PublicMapping/districtbuilder/pull/556)
 - Allow copying maps [#526](https://github.com/PublicMapping/districtbuilder/pull/526)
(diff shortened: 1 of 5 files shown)


kshepard (changes requested):
It looks like the field names have disappeared (they're just showing up as numbers). And also, I'm seeing the compactness column (field #6), but it's seemingly empty for districts that should have a compactness score. Like my district selected in this screenshot, which has a 36% compactness score on the frontend. 

screenshot_13

kshepard (changes requested):
The field names are showing up now, but I'm still not seeing any compactness values populated for contiguous districts that should have them, e.g.:

compactnes

kshepard (approved):

PR #560, opened 2021-01-25
Private / published maps

### LINKED ISSUE #533: Add private and published maps
We currently have a way for users to share an unlisted link to a map. A common request is to make maps private and then have the user specify them as public when they want to share them.

We will add 2 more options to the map link sharing option:
* Private - visible only to the authenticated author
* Link sharing only - visible to anyone who has the link (the current way it works)
* Published - visible on public DistrictBuilder pages (e.g. the user profile) or anyone with the link

[Wireframes can be found here](https://share.goabstract.com/d1021c24-3f58-4ee3-af2f-b6e5e6b5154e)

AC:
* Maps are private by default. The only user that can see the map by default is the authenticated user
* The Share button drops down a list of options: Private, Link sharing only, Published
* When selected, each option changes the icon on the button. By default, Private is selected and the button shows a 🔒 icon. When Link sharing is selected, the share button shows a 🔗. When Published is selected, the share button shows an 👁.
* The shareable link (which is currently available by clicking on the share button) is only visible when Link sharing or Published are selected
EDIT:
~* When a map is marked published, it is visible on the users profile page (#535)~

maurizi:
## Overview

This PR expands the functionality of the share menu to allow controling the visiblity of a map, which can now be:
 - private (the default for new maps)
 - share with link (the current behavior)
 - published (effectively the same as visible, but in a later feature we may list these publicly somewhere)

### Checklist

### Demo

image

### Notes

I made the default visibility for all existing maps "share with link" since that matches existing behavior and I didn't want to mess up anyone's already shared maps, but the default for _new_ maps is private as requested in the issue.

## Testing Instructions

- `scripts/migration`
- Existing maps should default to a visibility of Shared with link. New maps should default to Private.
- Ensure the backend appropriately returns 404s for private maps w/o authentication or when logged in as another user

Closes #533


kshepard (changes requested):
This is looking good, but it's missing something from the A.C.: there's supposed to be an icon displayed on the share button that changes based on the selected status.

kshepard (approved):

PR #562, opened 2021-01-26
Organization page stub

### LINKED ISSUE #224: Initial Organization Page and Route
We want to support organizations in DistrictBuilder so we'll need a page to serve as a profile for the organization. This task is to create the organization page and route.

Note that there we'll also need a new database model (`organizations`) and associated migration(s) to complete this. This table can be barebones for now: a unique slug name, and a label.

AC:
1. We can create an organization in the database enabling a route to that organization's page
2. The organization page can be found at districtbuilder.org/o/[orgname]
3. [orgname] is a field in the database
4. The organization page can be viewed anonymously

maurizi:
## Overview

This adds a skeleton frontend & backend for displaying an Organization page

### Checklist

### Demo

image

### Notes

@jfrankl I saw in the Abstract wireframes that this had a similar header to the main Home screen, so I made the Home screen header into a reusable `SiteHeader` component (this also brought along the Support option, which _wasn't_ in the wireframes). I can revert this easily enough if you think the headers on those pages are going to end up looking different.

## Testing Instructions

- `scripts/migration`
- Add a new Organization using `scripts/dbshell` (e.g. `INSERT INTO organization (name, slug) VALUES ( 'Mikes Map Competition', 'mikes-maps' );`)
- You should be able to access the Organization page for your organizations slug, whether or not you are logged in.

Closes #224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Share menu now controls map visibility (private/public/shared w link) [#560](https://github.com/PublicMapping/districtbuilder/pull/560)
+- Organization detail screen [#562](https://github.com/PublicMapping/districtbuilder/pull/562)
 
 ### Changed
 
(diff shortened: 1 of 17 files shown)


kshepard (approved):
Nice move breaking out the header :+1:

PR #571, opened 2021-02-08
Add internal-ip to package.json

kevinearldenny:
- Add internal-ip package to package.json
- Add line to README about creating user accounts

## Overview

Adds the `internal-ip` npm module to package.json, and adds a line in the README to answer a question that tripped me up during the setup process around creating user accounts.

### Checklist

- [ x] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

## Testing Instructions

- `./scripts/update`
- `./scripts/server`


maurizi (commented):
I don't think we should merge this PR - see my comment for more details, but TL;DR is I can't recreate the original issue and as far as I can tell this change shouldn't do anything (we had the same version of `internal-ip` already installed).

PR #572, opened 2021-02-08
Allow for duplication of a map project from home screen

### LINKED ISSUE #563: Add option to copy map from Home screen
## Detailed Description
 - This should be an option in the ProjectListFlyout component
 - Should create a copy of the project, with a new name (e.g. "Copy of <Project Name>", "Project Name Copy 1", etc)
 - Should show a success toast

kevinearldenny:
- Add actions for duplicating a project
- Add reducers for duplicating a project
- Fix reducer to reload projects on successful duplication
- Add changelog entry

## Overview

Allows for a user to create a copy of one of their existing projects from the context menu on the project's list record in the homepage view. 

### Checklist

### Demo

image

image

## Testing Instructions

- `./scripts/update`
- `./scripts/server`
- Navigate to the homepage at http://localhost:3005
- Click on the context menu for an existing project, and select 'Duplicate Map'
- Expect: A new project is created, with the name "Copy of {PROJECT}"

Closes #563

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Share menu now controls map visibility (private/public/shared w link) [#560](https://github.com/PublicMapping/districtbuilder/pull/560)
 - Organization detail screen [#562](https://github.com/PublicMapping/districtbuilder/pull/562)
+- Duplicate a map from home screen [#572] (https://github.com/PublicMapping/districtbuilder/pull/572)
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


kevinearldenny:
@jfrankl Semantically, it seems like 'Duplicate Project' might be a more appropriate name for this context menu option than 'Copy Project', especially since we already have a (separate) flow for a user to create a copy of _another user's_ project - I currently have it worded as such, though we could definitely switch it if you have a preference for 'Copy Project' (or something else entirely)

maurizi (changes requested):
I left some comments that need to be addressed, but at a high-level this looks great! 🎉

maurizi (approved):

Issue #564, opened 2021-02-03
Replace Blockgroups with Wards for Dane County


tgilcrest:
We want to replace blockgroups in Dane County with wards. We were provided a file that mapped blocks to ward IDs and we should use that to build the geolevel hierarchy. This is provided in a shapefile I received. The ward ID is in the field WardMuniUp.

AC:
1. For Dane County, the geolevel above blocks is Wards instead of Blockgroups

PR #573, opened 2021-02-09
Fix label for Ward geounits

maurizi:
## Overview

Fix label for new Dane County Ward geounits to say "Wards" instead of "ward" in the geounit selector.
Also makes the labelling function a little more generic to better support adding more types of geounits in the future.

### Checklist

### Demo

localhost_3003_

## Testing Instructions

- Add a new row to your `region_config` table using this for the `s3_uri` column: `s3://global-districtbuilder-dev-us-east-1/regions/US/WI/2021-02-08T15:19:22.145Z/`
- On `develop` the geounit label will be `ward`, on this branch the label will be `Wards`

Connects to #564

### DIFF
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -127,16 +127,10 @@ export function assertNever(x: never): never {
 
 export const geoLevelLabel = (id: string): string => {
   switch (id) {
-    case "block":
-      return "Blocks";
-    case "tract":
-      return "Tracts";
-    case "blockgroup":
-      return "Blockgroups";
     case "county":
       return "Counties";
     default:
-      return id;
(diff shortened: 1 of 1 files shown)


pcaisse (approved):
Tested well.

I think long term it would be good to move towards supplying these labels when we process the GeoJSON since we already supply the levels at that time. On the frontend, we could potentially look for the geolevel hierarchy labels and then fall back to this auto-formatting if it's not found. That way we avoid reprocessing existing states but then also potentially avoid incorrect pluralization going forward (I'm imagining another special case needing to be added in the future if a region config with a `municipality` geolevel id were added, for example).

PR #576, opened 2021-02-11
Add chamber field to project entity

### LINKED ISSUE #559: Store reference to chamber for created plans
## Overview

It will be useful to know which chamber (if any) was used to create a given plan. This will particularly be needed when we implement organization templates.

## Acceptance criteria

All future plans that are created have a reference to the chamber that was selected upon creation.

kevinearldenny:
## Overview

Adds a foreign key onto the Project entity to associate it with an optional Chamber that is selected by the user when a project is created.

### Checklist

- [ x ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

## Testing Instructions

- Ensure that you have your local dev environment set up with a region and a chamber associated with that region
- Create a new project with a region that has an associated chamber
- `./scripts/dbshell`
- `SELECT * FROM project`
- _Expect:_ the last project in the list has a chamber_id value associated with it equal to the ID of the chamber you selected on the frontend

Closes #559

### DIFF
--- a/src/client/screens/CreateProjectScreen.tsx
+++ b/src/client/screens/CreateProjectScreen.tsx
@@ -17,7 +17,7 @@ import {
 import { Link } from "react-router-dom";
 import { ReactComponent as Logo } from "../media/logos/mark-white.svg";
 
-import { IProject, IRegionConfig } from "../../shared/entities";
+import { IProject, IRegionConfig, IChamber } from "../../shared/entities";
 import { regionConfigsFetch } from "../actions/regionConfig";
 import { createProject } from "../api";
 import { InputField, SelectField } from "../components/Field";
@@ -38,6 +38,7 @@ const validate = (form: ProjectForm) =>
 
 interface ProjectForm {
   readonly name: string;
+  readonly chamber: IChamber | null;
   readonly regionConfig: IRegionConfig | null;
   readonly numberOfDistricts: number | null;
   readonly isCustom: boolean;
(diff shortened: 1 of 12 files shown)


src/client/screens/CreateProjectScreen.tsx
```diff
@@ -82,6 +88,7 @@ const CreateProjectScreen = ({ regionConfigs }: StateProps) => {
         ...(chamber
           ? {
               numberOfDistricts: chamber.numberOfDistricts,
+              chamber: chamber || null,
               isCustom: false
             }
           : { numberOfDistricts: null, isCustom: true })
```
maurizi, line 94: You need to explicitly null out `chamber` here or switching to a chamber & back to custom will still set a chamber ID.
```suggestion
          : { numberOfDistricts: null, isCustom: true, chamber: null })
```

src/client/screens/CreateProjectScreen.tsx
```diff
@@ -32,12 +32,16 @@ interface StateProps {
 }
 
 const validate = (form: ProjectForm) =>
-  form.name.trim() !== "" && form.numberOfDistricts !== null && form.regionConfig !== null
+  form.name.trim() !== "" &&
+  form.numberOfDistricts !== null &&
+  form.regionConfig !== null &&
+  form.chamber
```
maurizi, line 38: It is acceptable for `chamber` to be null (custom districts)
```suggestion
  form.regionConfig !== null
```

maurizi (changes requested):
Mostly looking good and I verified this is working for when a chamber is set - however note comments re: support for custom districts where chamber is not set, as well as a few small refactor suggestions / Qs.

maurizi (approved):
🎉

PR #582, opened 2021-02-18
Show error page for 404 errors on project and organization screens

### LINKED ISSUE #561: Add 404 page, use for organization and project pages
Currently, if you view a project that is hidden, or for which the ID is mis-spelled / copied, we end up with a partially loaded page and display an error toast.

This is a situation we should be able to handle more gracefully, either w/ a generic 404 page (which could also be applied to the Organization page) or with a project-specific error page

kevinearldenny:
## Overview

Displays an error screen on the Project or Organization pages if a user attempts to navigate to an entity (project / organization) that does not exist. 

### Checklist

### Demo

image
image

### Notes

I'm not entirely sure about how to handle the current toast notification that appears when a fetch error occurs. It seems like if nothing else, the copy should be updated here - since refreshing will not change the fact that the organization or project does not exist. But, I imagine that we want this to be displayed in the event that there is an unexpected failure when fetching a project that _does_ exist...

Also, we may need a little finessing on the style here @jfrankl 

## Testing Instructions

- `./scripts/update`
- Attempt to navigate to a URL for a project that does not exist, e.g. [http://localhost:3003/projects/777](http://localhost:3003/projects/777)
- Attempt to navigate to a URL for an organization that does not exist, e.g. [http://localhost:3003/o/FAKE](http://localhost:3003/o/FAKE)

Closes #561

### DIFF
--- a/src/client/screens/PageNotFoundScreen.tsx
+++ b/src/client/screens/PageNotFoundScreen.tsx
@@ -1,0 +1,25 @@
+/** @jsx jsx */
+import { Box, Card, Heading, jsx } from "theme-ui";
+
+import CenteredContent from "../components/CenteredContent";
+
+interface StateProps {
+  readonly model: string;
+}
+
+const PageNotFoundScreen = ({ model }: StateProps) => {
+  return (
+    <CenteredContent sx={{ mb: 5 }}>
+      <Card sx={{ variant: "card.floating" }}>
+        <Heading as="h2" sx={{ fontSize: 4, mb: 20, textAlign: "left" }}>
+          Page not found!
+        </Heading>
+        <Box sx={{ fontSize: 1, mt: 3, textAlign: "left" }}>
+          The {model} you are looking for could not be found :(
+        </Box>
+      </Card>
+    </CenteredContent>
+  );
+};
+
+export default PageNotFoundScreen;
--- a/src/client/screens/ProjectScreen.tsx
+++ b/src/client/screens/ProjectScreen.tsx
@@ -32,12 +32,15 @@ import { State } from "../reducers";
 import { Resource } from "../resource";
 import store from "../store";
 import { useBeforeunload } from "react-beforeunload";
+import PageNotFoundScreen from "./PageNotFoundScreen";
+import SiteHeader from "../components/SiteHeader";
 
 interface StateProps {
   readonly project?: IProject;
   readonly geojson?: DistrictsGeoJSON;
   readonly staticMetadata?: IStaticMetadata;
   readonly staticGeoLevels: UintArrays;
+  readonly projectNotFound?: boolean;
   readonly geoUnitHierarchy?: GeoUnitHierarchy;
   readonly districtDrawing: DistrictDrawingState;
   readonly isLoading: boolean;
@@ -62,6 +65,7 @@ const ProjectScreen = ({
   geojson,
   staticMetadata,
   staticGeoLevels,
+  projectNotFound,
   geoUnitHierarchy,
   districtDrawing,
   isLoading,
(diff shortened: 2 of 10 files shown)


jfrankl:
Nice! I have a few design changes:

- Remove the "DistrictBuilder" logo from just above the "Page not found" box, since that information is in the header
- Left align all text
- @tgilcrest could you take a shot at the copy for this?

maurizi (changes requested):
> I'm not entirely sure about how to handle the current toast notification that appears when a fetch error occurs. It seems like if nothing else, the copy should be updated here - since refreshing will not change the fact that the organization or project does not exist. But, I imagine that we want this to be displayed in the event that there is an unexpected failure when fetching a project that does exist...

The way this is implemented, _any_ error on an organization or project page will result in displaying a "not found" page - even an error _updating_ an existing project, like a transient server error.

We should only show this page specifically for 404 errors.

src/client/screens/PageNotFoundScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+const PageNotFoundScreen = ({ organization }: StateProps) => {
+  return (
+    <CenteredContent sx={{ mb: 5}}>
+      <Card sx={{ variant: "card.floating"}}>
+        <Heading as="h2" sx={{ fontSize: 4, mb: 20, textAlign: "left" }}>
+          Page not found!
+        </Heading>
+        <Box sx={{ fontSize: 1, mt: 3, textAlign: "left" }}>
+          The {"errorMessage" in organization ? "organization" : "project"} you are looking for
```
maurizi, line 22: Rather than determining what to show here based on the Redux state, I think it might be less brittle to handle this via a prop when the component is used, e.g.
```
<PageNotFoundScreen model="organization" />
```

kevinearldenny:
> > I'm not entirely sure about how to handle the current toast notification that appears when a fetch error occurs. It seems like if nothing else, the copy should be updated here - since refreshing will not change the fact that the organization or project does not exist. But, I imagine that we want this to be displayed in the event that there is an unexpected failure when fetching a project that does exist...
> 
> The way this is implemented, _any_ error on an organization or project page will result in displaying a "not found" page - even an error _updating_ an existing project, like a transient server error.
> 
> We should only show this page specifically for 404 errors.

Hmmm, so malformed requests are currently returning a 400 error. Is it fair to handle these sorts of errors like:

"errorMessage" in state.project.projectData && state.project.projectData.errorMessage && `state.project.projectData.errorMessage === "Request failed with status code 400"`?

maurizi:
> Hmmm, so malformed requests are currently returning a 400 error. Is it fair to handle these sorts of errors like:
> `"errorMessage" in state.project.projectData && state.project.projectData.errorMessage && state.project.projectData.errorMessage === "Request failed with status code 400"?`

That seems brittle and likely to break if this automatically generated message were to change for some reason (e.g. a browser / library upgrade).

We should refactor the error data passed through to include the status code along with or instead of the existing `errorMessage`. I don't think we actually display the `errorMessage` anywhere right now just check for its presence, so that shouldn't be too hard.

kevinearldenny:
> > Hmmm, so malformed requests are currently returning a 400 error. Is it fair to handle these sorts of errors like:
> > `"errorMessage" in state.project.projectData && state.project.projectData.errorMessage && state.project.projectData.errorMessage === "Request failed with status code 400"?`
> 
> That seems brittle and likely to break if this automatically generated message were to change for some reason (e.g. a browser / library upgrade).
> 
> We should refactor the error data passed through to include the status code along with or instead of the existing `errorMessage`. I don't think we actually display the `errorMessage` anywhere right now just check for its presence, so that shouldn't be too hard.

Okay, I refactored for `errorMessage` to display the status code from the error response.

src/client/screens/ProjectScreen.tsx
```diff
@@ -193,6 +202,10 @@ function mapStateToProps(state: State): StateProps {
     isLoading:
       ("isPending" in state.project.projectData && state.project.projectData.isPending) ||
       ("isPending" in state.project.staticData && state.project.staticData.isPending),
+    projectNotFound: !!(
+      ("statusCode" in state.project.projectData && state.project.projectData.statusCode === 400) ||
+      ("statusCode" in state.project.staticData && state.project.staticData.statusCode === 400)
+    ),
```
kevinearldenny, line 208: @maurizi I decided to leave this as 'working' with the Redux architecture that you proposed, and since there are a few other outstanding things on this page around error handling split up the 400 vs 404 error piece into a separate issue: https://github.com/PublicMapping/districtbuilder/issues/595

maurizi (approved):

PR #592, opened 2021-02-23
Add option to scripts/server to run in debug mode

maurizi:
## Overview

Adds a `--debug` flag to `scripts/server` to run the backend in debug mode, with the standard `9229` Node.js debugging port exposed through Docker to the host.

## Notes

I used VS Code's built-in debugger with the following configuration, but I believe other tools compatible w/ the Node inspector like the Chrome dev tools should also work:
```
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "attach",
      "name": "Attach to Node",
      "localRoot": "${workspaceFolder}/src",
      "remoteRoot": "/home/node/app",
      "port": 9229
    }
  ]
}
```

## Testing Instructions

- `scripts/server --debug`
- Attach a debugger to the backend
- Ensure you can set a breakpoint in backend code

### DIFF
--- a/docker-compose.debug.yml
+++ b/docker-compose.debug.yml
@@ -1,0 +1,11 @@
+version: "2.4"
+services:
+  server:
+    ports:
+      - 3005:3005
+      - 9229:9229
+    command: start:debug
+
+  client:
+    ports:
+      - 3003:3003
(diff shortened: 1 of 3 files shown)


maurizi:
@pcaisse Kevin ran into some issues w/ setting this up, do you think you could give it a go?

pcaisse (approved):
I used VSCode with your config and it worked great

Issue #179, opened 2020-06-10
Add ability to import external dependencies from the shared module


maurizi:
## Detailed Description
In #177 I tried to add types to a module in `shared` that were aliases of `geojson` types.

However, because the `shared` module is outside of the root directory when accessed from the `server`, it does not have a `node_modules` directory and TypeScript cannot find any external modules.

PR #594, opened 2021-02-23
Project districts geometry

### LINKED ISSUE #567: Store District GeoJSON on Project entity
Currently we generate district geojson as-needed from the project's `districtDefinition` column, but for a number of features it would be better to have these cached:
 - Improved load times for the project screen
 - Displaying project maps on the home screen / organization screen
 - Enabling a read-only mode for old regions
 
 AC
  - A new column is added to `Project` to store GeoJSON, w/ appropriate defaults for new projects
  - Updates to the `districtDefinition` field cause the project GeoJSON to be updated
  - The GeoJSON export endpoint is refactored to use this new data

maurizi:
## Overview

Brief description of what this PR does, and why it is needed.

### Checklist

### Demo

Cached `district` field can be seen in debugger:
image

### Notes

Note the `districts` is excluded from the main project detail endpoint and I also left it off of the `IProject` interface - instead access continues to go through the `/export/geojson` endpoint. This solves two issues:
 - We can't import from `geojson` in the shared module yet (see #179) so I can't move the `DistrictsGeoJSON` type there
 - Including the geojson on the list page would unnecessarily slow down things
 
There are definitely downsides to this approach though, and I think we could refactor this further to eliminate the need for the `/export/geojson` endpoint.

## Testing Instructions

- `scripts/update`
- Displaying and updating districts should continue to work as normal

Closes #567

### DIFF
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -291,6 +291,9 @@ export class ProjectsController implements CrudController<Project> {
         ...feature,
         properties: {
           ...feature.properties,
+          // Flatten nested demographics object so it is maintained when converting
+          demographics: undefined,
+          ...feature.properties.demographics,
           // The feature ID doesn't seem to make its way over as part of 'convert' natively
           id: feature.id
         }
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -31,8 +31,8 @@ import { GeometryCollection } from "topojson-specification";
 
 import { MakeDistrictsErrors } from "../../../../shared/constants";
 import {
+  DistrictsDefinition,
   GeoUnitHierarchy,
-  IProjectTemplate,
   ProjectId,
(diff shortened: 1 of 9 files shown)


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -241,27 +268,43 @@ export class ProjectsController implements CrudController<Project> {
...
+  async getGeojson({
+    districtsDefinition,
+    numberOfDistricts,
+    regionConfig
+  }: {
+    readonly districtsDefinition: DistrictsDefinition;
+    readonly numberOfDistricts: number;
+    readonly regionConfig: RegionConfig;
+  }): Promise<DistrictsGeoJSON> {
+    const geoCollection = await this.getGeoUnitTopology(regionConfig.s3URI);
```
pcaisse, line 280: I wonder if it might be a good idea to add some INFO logging to this controller/topology service so it's more clear when we're using cached data vs generating GeoJSON.

pcaisse (approved):
LGTM and tested well! Left a couple of small comments.

PR #611, opened 2021-03-04
Add paintbrush selection tool

### LINKED ISSUE #149: Paintbrush drawing tool
Add a paint brush drawing tool that selects geounits. The interaction we want to emulate can be seen in the prototype ([link here](https://azavea.slack.com/archives/C6UD5FG3T/p1593548105043600)).

AC:
* There is a paintbrush drawing tool that behaves like the prototype, with the following exceptions:
* The paintbrush only highlights the 1 geo-unit the mouse is hovering over. (It should have the smallest "brush size" possible. The prototype highlights a radius of geo-units, and we don't want this larger "brush" size to be the default)
* The label of that blockgroup is shown, like the rectangle select behaves in the production app. (In the prototype, it says "1 [geo-unit]")

maurizi:
## Overview

Adds a paintbrush tool to the project editing options

### Checklist

### Demo

output

## Testing Instructions

- `scripts/server`
- Edit a map and select the 🖌 tool

Closes #149


pcaisse (changes requested):
This is looking really good so far!

I did find a bit of an edge case bug, similar to the types of issues I ran into with the rectangle selection tool, where if you select a blockgroup and then select the county containing that blockgroup you end up with a double selection:
db_paintbrush_double_selection

I think you'd need something like this to filter out geounits that are already selected from _all_ geolevels (see `getAllSelectedFeatures`): https://github.com/PublicMapping/districtbuilder/blob/develop/src/client/components/map/RectangleSelectionTool.ts#L148-L152

pcaisse (approved):
I talked with @maurizi about this offline and it seems that the issue mentioned re: selecting geounits whose children are already selected causing a double selection also exists with the rectangle selection tool. As such, this can be tackled separately and applied to both tools at the same time.

PR #613, opened 2021-03-05
Fix 404 page

### LINKED ISSUE #595: Fix error handling for 404 page
## Description
Currently all 400 errors encountered when accessing a project detail receive the same treatment - redirect to the 'project not found' page created implemented in https://github.com/PublicMapping/districtbuilder/pull/582 

However, we should be handling errors more robustly than this. If an error occurs while attempting to make changes to an existing project, the user should not be redirected. We also should not be displaying the error toast encouraging users to refresh the page on the 404 error screen - this is bad advice.

AC

- Invalid string project IDs return 404 from the server (E.g., `/api/projects/bob`)
- Invalid numeric projects IDs return 404 from the server (`/api/projects/9`)
- ProjectScreen only displays error page on 404 error
- Error toast that reads "Something went wrong, please refresh the page" should not be displayed on 404 error page. 
- Editing errors on ProjectScreen trigger the error toast and do not display the 'project not found' error screen. 

## Steps To Reproduce
I am not entirely sure how to produce a project editing error on the frontend currently. 

## Expected behavior
User should not be redirected to the 404 error page unless they have attempted to access a project page for a project that does not exist. All other errors on the project page should only trigger the error modal.

## Actual behavior
All 400 errors on the Project page redirect to 404 error page 

## Your environment
 - macOS

## Additional context

- This came up as I was pairing with @maurizi on #561 (PR  #582 ) as a hypothetical problem that might come up with my solution. That was only a 1 point issue and I've already spent over 2x the estimated time on it, so it probably makes sense to split this out into a separate issue.

maurizi:
## Overview

 - Fixes the backend to properly return 404 in all cases for invalid project URLS or missing IDs

### Checklist

## Testing Instructions

- Got to a project
  - Changing one character in ID in the URL, so you have a _valid_ UUID which is not in our database, and refreshing the page, should return a 404 from backend and displaying the Not Found component
  - Replacing the ID in the URL, so you have an _invalid_ UUID, and refreshing the page, should return a 404 from backend and displaying the Not Found component
- Attempting to visit an organization for a slug not in our DB should result in displaying the Not Found component
- Neither of the above should display failure toasts

Closes #595

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display error page for 404 errors [#582](https://github.com/PublicMapping/districtbuilder/pull/582)
 - Allow users to join and leave an organization [#226](https://github.com/PublicMapping/districtbuilder/pull/578)
 - Require users to confirm before joining an organization [#593](https://github.com/PublicMapping/districtbuilder/pull/593)
+- Show not found page for missing projects and organizations [#613](https://github.com/PublicMapping/districtbuilder/pull/613)
 
 ### Changed
 
--- a/src/client/api.ts
+++ b/src/client/api.ts
@@ -142,7 +142,7 @@ async function fetchProject(id: ProjectId): Promise<IProject> {
       .get(`/api/projects/${id}`)
       .then(response => resolve(response.data))
       .catch(error =>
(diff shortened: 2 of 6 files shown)


kevinearldenny:
Looks good 👍

kevinearldenny (approved):
Looks good 👍

Issue #610, opened 2021-03-03
Fix preventing joining organizations if user is unverified


maurizi:
The user is supposed to be prevented from joining an organization if their email address is unverified, but we're not fully handling this, merely preventing using the "Join this organization" button:
![localhost_3003_o_mikes-maps_ (9)](https://user-images.githubusercontent.com/4432106/109854942-a9d6e880-7c25-11eb-9a43-6df1ec633983.png)

You can bypass this check in several ways:
 - Using "Use this template" when logged in but not verified
 - Starting logged out, clicking "Join this organization" or "Use this template" and then use the modal to create a new account or lo into an unverified account

AC:
 - The "Use this template" button should also be disabled, and the same tooltip shown for "Join this organization" shown there
 - After logging into an unverified account subsequent to clicking "Join this account" or "Use this template" while logged out, we should show a _new_ modal with copy stating that the user must verify their account to join an organization

PR #615, opened 2021-03-08
Disable "Use this template" button if user email is unverified

maurizi:
## Overview

We were already doing this for the "Join this organization" button, but now we also disable the "Use this template" button.

### Checklist

### Demo

image

### Notes

This is only the first half of #610, opening this now to avoid conflicts w/ @kevinearldenny 

## Testing Instructions

- `scripts/server`

Connects to #610

### DIFF
--- a/src/client/screens/OrganizationScreen.tsx
+++ b/src/client/screens/OrganizationScreen.tsx
@@ -136,6 +136,47 @@ const OrganizationScreen = ({ organization, user }: StateProps) => {
     userInOrg ? createProjectFromTemplate() : store.dispatch(showCopyMapModal(true));
   }
 
+  const joinButton = (
+    <Flex sx={{ flexDirection: "column", flex: "none" }}>
+      <Button sx={style.join} disabled={isLoggedIn && !userIsVerified} onClick={signupAndJoinOrg}>
+        Join organization
+      </Button>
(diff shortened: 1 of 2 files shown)


src/client/screens/OrganizationScreen.tsx
```diff
@@ -88,18 +88,9 @@ const OrganizationScreen = ({ organization, project, user }: StateProps) => {
...
   const userIsVerified = "resource" in user && user.resource && user.resource.isEmailVerified;
 
-  useEffect(() => {
-    !userLoggedIn &&
-      "resource" in user &&
-      user.resource &&
-      project.showCopyMapModal &&
-      store.dispatch(showCopyMapModal(false));
-  }, [user, project.showCopyMapModal]);
-
```
maurizi, line 102: @kevinearldenny As far as I could tell this didn't do anything?

kevinearldenny:
@maurizi This is mostly looking good, but there is a bit of related functionality I included in #616 that I included that helps to make the current 'unverified user' state more apparent here:

image

maurizi:
@kevinearldenny As per our discussion offline lets add those in #616 after this PR gets merged

src/client/screens/OrganizationScreen.tsx
kevinearldenny: Yep, you're correct on this. I removed it in my updates to #616

Issue #609, opened 2021-03-03
Show verify email banner on organization detail screen


maurizi:
On the main home page we show a banner if the user has not yet verified their email address:
![localhost_3003_ (1)](https://user-images.githubusercontent.com/4432106/109854677-52d11380-7c25-11eb-8964-d43e4c17fb20.png)

We should show this same banner on the organization detail screen if needed

PR #616, opened 2021-03-08
Add verify and join org link to verification email

### LINKED ISSUE #586: Add verify and join org link to verification email
When users create an account after trying to join an organization, we should keep track of that and make them automatically join the organization after verifying their email address.

Unlike when normally signing up, users should land on the organization homepage instead of the main districtbuilder homepage.

kevinearldenny:
## Overview

When a new user attempts to sign up for an account and join an organization from an organization's page, they are automatically added to the Organization as an 'unverified user'. If an unverified user views their organization's page before confirming their email, they are prompted to confirm their email.

### Checklist

### Demo

image

### Notes

## Testing Instructions

- Ensure that you are not logged in with an existing user account
- Navigate to an organization page (e.g., http://localhost:3003/o/mikes-maps)
- Click 'Join Organization' and fill in credentials not associated with an existing user

Closes #586 & #609

### DIFF
--- a/src/client/screens/OrganizationScreen.tsx
+++ b/src/client/screens/OrganizationScreen.tsx
@@ -163,12 +163,18 @@ const OrganizationScreen = ({ organization, user }: StateProps) => {
           {template.regionConfig.name} · {template.numberOfDistricts}
         </Text>
         <Text>{template.description}</Text>
-        {!isLoggedIn || userIsVerified ? (
+        {userIsVerified || !isLoggedIn ? (
           useButton
         ) : (
           <Tooltip
             key={template.id}
-            content={<div>You must confirm your email before joining an organization</div>}
+            content={
+              <div>
+                {userInOrg
+                  ? "You must confirm your email before using your organization's template"
+                  : "You must confirm your email before using this organization's template"}
+              </div>
+            }
           >
             <Box>{useButton}</Box>
           </Tooltip>
@@ -213,7 +219,7 @@ const OrganizationScreen = ({ organization, user }: StateProps) => {
                   <Box>{organization.resource.description}</Box>
                 )}
               </Box>
-              {userInOrg ? (
+              {userInOrg && userIsVerified ? (
                 <Flex sx={{ flexDirection: "column", flex: "none" }} onClick={leaveOrg}>
                   <Button sx={style.join}>Leave organization</Button>
                 </Flex>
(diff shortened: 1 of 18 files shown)


src/client/screens/OrganizationScreen.tsx
```diff
@@ -147,6 +151,49 @@ const OrganizationScreen = ({ organization, project, user }: StateProps) => {
...
+                Error resending email. If this error persists, please contact us at{" "}
+                <Styled.a sx={{ color: "muted" }} href="mailto:support@districtbuilder.org">
+                  support@districtbuilder.org
+                </Styled.a>
+                .
+              </Box>
+            )}
+          </Box>
+        </Alert>
+      )}
```
maurizi, line 196: Refactoring this banner was supposed to be tackled by https://github.com/PublicMapping/districtbuilder/issues/609 but in any case, we should just copy/paste the markup here - we're already sharing the `SiteHeader` component between the two pages, we should just _move_ [this code in `HomeScreen`](https://github.com/PublicMapping/districtbuilder/blob/develop/src/client/screens/HomeScreen.tsx#L78-L120) into `SiteHeader` as well and we won't need these changes.

kevinearldenny: Gotcha, okay - guess it probably makes sense to pull #609 into this PR as well

kevinearldenny:
@maurizi Merge conflicts are now resolved here, ready for review

src/client/screens/OrganizationScreen.tsx
```diff
@@ -223,26 +223,49 @@ const OrganizationScreen = ({ organization, user }: StateProps) => {
...
+                          : "You must confirm your email before joining an organization"}
+                      </div>
+                    }
                   >
                     {joinButton}
                   </Tooltip>
                 )
               ) : (
-                joinButton
+                <div></div>
```
maurizi, line 238: I don't think you meant to get rid of the Join button for logged out users?
```suggestion
                joinButton
```

src/client/screens/OrganizationScreen.tsx
```diff
@@ -223,26 +223,49 @@ const OrganizationScreen = ({ organization, user }: StateProps) => {
...
+                            </div>
+                          }
+                        >
+                          <TemplateCard template={template} key={template.id} />
+                        </Tooltip>
+                      ) : (
+                        <TemplateCard template={template} key={template.id} />
+                      )
+                    )}
+                  </Box>
```
maurizi, line 265: This is not correct - the tooltip should be wrapped around the `<Button>`, not the `<TemplateCard>` component created in #615.

maurizi (changes requested):

maurizi (changes requested):
The main functionality of storing intent to join an organization as part of registering is working, and I was able to activate my account and was a part of the organization.

This is still missing the feature to redirect to the organization upon activation, as per the wireframes on Abstract:
image

As we discussed offline that will likely require some kind of new flag on `User` so we don't show that modal again, or thinking about it a bit more - storing a separate FK field on `User` for organization(s)-to-join instead of reusing the existing `organizations` M2M field could serve this purpose as well (I think you originally suggested this).

maurizi (approved):

PR #623, opened 2021-03-11
Build out scaffolding for Project Evaluate view

### LINKED ISSUE #603: Add basic evaluate mode skeleton page
Main evaluate page:
image

Example subpage:
image

AC
 - Add evaluate button to project header
 - Clicking evaluate should switch the map view and swap out the sidebar contents
   - The map should zoom out to the starting region extent
   - Skip the map styling changes
 - Evaluate sidebar should show all 2 required and 4 more options and allow navigating to subpages
   - Skip any subpage content / map layer changes other than copy (skeleton components)
   - Do implement the "Back to summary" link

kevinearldenny:
## Overview

- Add new EvaluateMetric type
- Add ProjectEvaluateView
- Add ProjectEvaluateMetricDetail
- Build out actions for toggling evaluate mode and switching between metrics

### Checklist

### Demo

image

image

### Notes

## Testing Instructions

- Open a project
- Click the evaluate button in the header
- _Expect_: Evaluate view displayed, with metrics listed in sidebar
- Click a metric in the sidebar
- _Expect_: Metric detail view displayed

Closes #603

### DIFF
--- a/src/shared/entities.d.ts
+++ b/src/shared/entities.d.ts
@@ -15,6 +15,14 @@ export interface IUser {
   readonly organizations: readonly OrganizationNest[];
 }
 
+export type EvaluateMetric = {
+  readonly name: string;
+  readonly type?: "fraction" | "percent" | "count";
+  readonly value?: number;
+  readonly status?: boolean;
+  readonly description: string;
+};
+
 export type UpdateUserData = Pick<IUser, "name" | "hasSeenTour">;
 
 export type OrganizationSlug = string;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add verify and join link to user verification email [#616](https://github.com/PublicMapping/districtbuilder/pull/616)
 - Display organizations dropdown in header [#620](https://github.com/PublicMapping/districtbuilder/pull/620)
 - Organization Admin page and featured map workflow [#614](https://github.com/PublicMapping/districtbuilder/pull/614)
+- Build out scaffolding for Project Evaluate view [#623](https://github.com/PublicMapping/districtbuilder/pull/623)
 
 ### Changed
 
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -1,5 +1,5 @@
 import { createAction } from "typesafe-actions";
-import { DistrictId, GeoUnits } from "../../shared/entities";
+import { DistrictId, EvaluateMetric, GeoUnits } from "../../shared/entities";
 import { SavingState } from "../types";
 
 export enum SelectionTool {
@@ -55,6 +55,11 @@ export const redo = createAction("Redo project action")();
 export const toggleFind = createAction("Toggle find menu visibility")<boolean>();
 export const setFindType = createAction("Set find menu search type")<FindTool>();
 export const setFindIndex = createAction("Set find menu polygon index")<number | undefined>();
+
+export const toggleEvaluate = createAction("Toggle evaluate mode")<boolean>();
+export const selectEvaluationMetric = createAction("Select evaluation metric")<
+  EvaluateMetric | undefined
+>();
 
 export const saveDistrictsDefinition = createAction("Save districts definition")();
 
--- a/src/client/components/ProjectEvaluateMetricDetail.tsx
+++ b/src/client/components/ProjectEvaluateMetricDetail.tsx
@@ -1,0 +1,102 @@
+/** @jsx jsx */
+import { Box, Button, Flex, jsx, ThemeUIStyleObject, Heading } from "theme-ui";
+
+import { EvaluateMetric } from "../../shared/entities";
+import Icon from "./Icon";
+import { DistrictsGeoJSON } from "../types";
+import store from "../store";
+import { selectEvaluationMetric } from "../actions/districtDrawing";
+
+const style: ThemeUIStyleObject = {
+  td: {
+    fontWeight: "body",
+    color: "gray.8",
+    fontSize: 1,
+    p: 2,
+    textAlign: "left",
+    verticalAlign: "bottom"
+  },
+  header: {
+    variant: "header.app",
+    borderBottom: "1px solid",
+    borderColor: "gray.2",
+    paddingBottom: "20px",
+    flexDirection: "column",
+    m: "0"
+  },
+  closeBtn: {
+    position: "absolute",
+    right: "20px"
+  },
+  metricRow: {
+    p: 1,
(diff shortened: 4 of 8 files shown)


pcaisse (changes requested):
This is looking pretty good. Tested well.

I noticed this one warning in the console (looks like we need to set a `key` somewhere):
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `ProjectEvaluateMetricDetail`. See https://fb.me/react-warning-keys for more information.
    in EmotionCssPropInternal (at ProjectEvaluateMetricDetail.tsx:89)
    in ProjectEvaluateMetricDetail (at ProjectEvaluateView.tsx:208)
    in div (created by Context.Consumer)
    in Styled(div) (created by ForwardRef)
    in ForwardRef (created by Context.Consumer)
    in EmotionCssPropInternal (at ProjectEvaluateView.tsx:143)
    in div (created by Context.Consumer)
    in Styled(div) (at ProjectEvaluateView.tsx:142)
    in ProjectEvaluateView (at ProjectScreen.tsx:195)
    in div (created by Context.Consumer)
    in Styled(div) (created by Context.Consumer)
```

I did noticed that the map is being reloaded when evaluate mode is toggled. I think we should try to avoid that if we could since it isn't strictly necessary. It looks like `ProjectEvaluateView` renders to `Map` in much the same way that `ProjectScreen` does. I think that instead of having wholly separate code paths for the rendering of `ProjectScreen`, it might make sense to add some conditional logic to hide/show the header if in evaluate mode and also conditionally show the project sidebar/evaluate sidebar depending on which mode we're in. That way we can avoid mounting/unmounting `Map` based on the mode, avoiding the refresh. It also should significantly reduce the number of props being fed into `ProjectEvaluateView` since most of them relate to populating the map correctly and allow for getting rid of the `setMap` callback in `ProjectEvaluateView` which is duplicated from `ProjectScreen`.

Let me know if that makes sense.

maurizi:
> I did notice that the map is being reloaded when evaluate mode is toggled. I think we should try to avoid that if we could since it isn't strictly necessary.

I would like to second this as well - Jeff wants to do an animation/transition on the map when you toggle evaluate mode, which requires not reloading the map.

kevinearldenny:
> This is looking pretty good. Tested well.
> 
> I noticed this one warning in the console (looks like we need to set a `key` somewhere):
> 
> ```
> Warning: Each child in a list should have a unique "key" prop.
> 
> Check the render method of `ProjectEvaluateMetricDetail`. See https://fb.me/react-warning-keys for more information.
>     in EmotionCssPropInternal (at ProjectEvaluateMetricDetail.tsx:89)
>     in ProjectEvaluateMetricDetail (at ProjectEvaluateView.tsx:208)
>     in div (created by Context.Consumer)
>     in Styled(div) (created by ForwardRef)
>     in ForwardRef (created by Context.Consumer)
>     in EmotionCssPropInternal (at ProjectEvaluateView.tsx:143)
>     in div (created by Context.Consumer)
>     in Styled(div) (at ProjectEvaluateView.tsx:142)
>     in ProjectEvaluateView (at ProjectScreen.tsx:195)
>     in div (created by Context.Consumer)
>     in Styled(div) (created by Context.Consumer)
> ```
> 
> I did notice that the map is being reloaded when evaluate mode is toggled. I think we should try to avoid that if we could since it isn't strictly necessary. It looks like `ProjectEvaluateView` renders to `Map` in much the same way that `ProjectScreen` does. I think that instead of having wholly separate code paths for the rendering of `ProjectScreen`, it might make sense to add some conditional logic to hide/show the header if in evaluate mode and also conditionally show the project sidebar/evaluate sidebar depending on which mode we're in. That way we can avoid mounting/unmounting `Map` based on the mode, avoiding the refresh. It also should significantly reduce the number of props being fed into `ProjectEvaluateView` since most of them relate to populating the map correctly and allow for getting rid of the `setMap` callback in `ProjectEvaluateView` which is duplicated from `ProjectScreen`.
> 
> Let me know if that makes sense.

Gotcha, this makes sense. I initially was thinking that we explicitly wanted a separate map view, since the container is of a different size and the data display is totally different, but in light of your note I refactored the `ProjectEvaluateView` component into a view that only lives in the sidebar. 

I also added `id` as a key prop to the list of features displayed in the `ProjectEvaluateMetricView` component.

src/shared/entities.d.ts
```diff
@@ -15,6 +15,14 @@ export interface IUser {
   readonly organizations: readonly OrganizationNest[];
 }
 
+export type EvaluateMetric = {
+  readonly name: string;
+  readonly type?: string;
+  readonly count?: number;
+  readonly status?: boolean;
+  readonly adjective?: string;
```
pcaisse, line 23: I wonder if this should be called `description` or something since it's not really an adjective. It also does seem to be required so maybe could ditch the `?`.

pcaisse: You could go ahead and constrain which types are possible here so that the compiler will call out any typos, like so:
```
export type EvaluateMetric = {
  readonly name: string;
  readonly status: boolean;
  readonly adjective: string;
  readonly type: "fraction" | "percent" | "count";
  readonly value: number; 
};
```
This can then be leveraged for exhaustiveness checking. Instead of as it is now where we return a nonsense value `???` if we make a mistake:
```
                      {metric.type === "fraction"
                        ? `${metric.count} / 18`
                        : metric.type === "percent"
                        ? `${metric.count}%`
                        : metric.type === "count"
                        ? `${metric.count}`
                        : "???"}
```
We can have a format function that will error at compile time if we try to feed it a bad value:
```ts
function formatMetricValue(metric: EvaluateMetric): string {
  switch (metric.type) {
    case "fraction":
      return `${metric.count} / 18`;
    case "percent":
      return `${metric.count}%`;
    case "count":
      return `${metric.count}`;
    default:
      const exhaustive: never = metric;
      return exhaustive;
  }
}
```

pcaisse (approved):
LGTM. Left a couple of TypeScript-ful comments.

src/shared/entities.d.ts
kevinearldenny: @pcaisse This mostly does look like a nice improvement, but I'm getting a TS compile error:

```
src/client/components/ProjectEvaluateSidebar.tsx:132:15 - error TS2322: Type 'EvaluateMetric' is not assignable to type 'never'.

132         const exhaustive: never = metric;
```

Any thoughts on fixing this? I added a `// @ts-ignore` above it in the commit so it would compile, but then after it compiled it was still throwing an error that caused the whole screen to blank out when you attempted to navigate to Evaluate mode. I updated it to:

```
// @ts-ignore
const exhaustive: never = metric;
```

but there might be a better way to handle this?

pcaisse: I think this would be a great thing to pair on if you'd like.

PR #626, opened 2021-03-15
Make org project templates active or inactive

### LINKED ISSUE #597: Make Org Templates Active or Inactive
Currently if there is a template configured for an org, we always show that template on the org page. We want to make it possible to deactivate a template so it is not visible on the org page but the template still exists so any created maps can still reference it.

AC:
* It is possible to activate/deactivate an org's template
* When deactivated, the org template does not show on the org page
* When deactivated, any maps that have been created by the org still reference the template from which they were created

Note:
This doesn't need to be controllable by the org admin; it could just be something we do in the database.

kevinearldenny:
## Overview

- Adds `isActive` field to Project Template entity
- Only display active templates on Org screen

### Checklist

### Notes

## Testing Instructions

- Create an organization with a project template using the `update-organization` command
- Duplicate the organization's project template in the psql shell
- Update the value of `isActive` on the copied record to be false
- Reload the organization screen; expect that the second template is no longer displayed on the screen

Closes #597

### DIFF
--- a/src/server/src/organizations/services/organizations.service.ts
+++ b/src/server/src/organizations/services/organizations.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from "@nestjs/common";
 import { InjectRepository } from "@nestjs/typeorm";
 import { TypeOrmCrudService } from "@nestjsx/crud-typeorm";
 import { Repository } from "typeorm";
+import { IOrganization } from "../../../../shared/entities";
 
 import { Organization } from "../entities/organization.entity";
 
@@ -12,6 +13,46 @@ export class OrganizationsService extends TypeOrmCrudService<Organization> {
   }
 
   save(org: Organization): Promise<Organization> {
+    // @ts-ignore
     return this.repo.save(org);
+  }
+
+  async getOrgAndProjects(slug: string): Promise<Organization | undefined> {
+    const builder = this.repo.createQueryBuilder("organization");
+    const data = await builder
+      .leftJoinAndSelect("organization.users", "users")
+      .leftJoinAndSelect("organization.admin", "admin")
+      .leftJoinAndSelect(
+        "organization.projectTemplates",
+        "projectTemplates",
+        "projectTemplates.isActive = TRUE"
+      )
+      .leftJoinAndSelect("projectTemplates.regionConfig", "regionConfig")
+      .where("organization.slug = :slug", { slug: slug })
+      .select([
+        "organization.id",
+        "organization.slug",
+        "organization.name",
+        "organization.description",
+        "organization.linkUrl",
+        "organization.municipality",
+        "organization.region",
+        "users.id",
+        "users.name",
+        "admin.id",
+        "admin.name",
+        "projectTemplates.id",
+        "projectTemplates.name",
+        "regionConfig.name",
+        "regionConfig.id",
+        "regionConfig.countryCode",
+        "regionConfig.regionCode",
+        "projectTemplates.numberOfDistricts",
+        "projectTemplates.description",
+        "projectTemplates.details",
+        "projectTemplates.isActive"
+      ])
+      .getOne();
+    return data;
   }
 }
--- a/src/server/src/project-templates/services/project-templates.service.ts
+++ b/src/server/src/project-templates/services/project-templates.service.ts
@@ -19,6 +19,7 @@ export class ProjectTemplatesService extends TypeOrmCrudService<ProjectTemplate>
       .innerJoinAndSelect("projectTemplate.projects", "projects")
       .innerJoinAndSelect("projects.user", "user")
       .where("organization.slug = :slug", { slug: slug })
+      .andWhere("projectTemplate.isActive = TRUE")
       .select([
         "projectTemplate.name",
         "projectTemplate.numberOfDistricts",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display organizations dropdown in header [#620](https://github.com/PublicMapping/districtbuilder/pull/620)
 - Organization Admin page and featured map workflow [#614](https://github.com/PublicMapping/districtbuilder/pull/614)
 - Build out scaffolding for Project Evaluate view [#623](https://github.com/PublicMapping/districtbuilder/pull/623)
+- Add flag on project templates to mark as active or inactive [#626](https://github.com/PublicMapping/districtbuilder/pull/626)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add verify and join link to user verification email [#616](https://github.com/PublicMapping/districtbuilder/pull/616)
 - Display organizations dropdown in header [#620](https://github.com/PublicMapping/districtbuilder/pull/620)
 - Organization Admin page and featured map workflow [#614](https://github.com/PublicMapping/districtbuilder/pull/614)
+- Build out scaffolding for Project Evaluate view [#623](https://github.com/PublicMapping/districtbuilder/pull/623)
 
 ### Changed
 
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -1,5 +1,5 @@
 import { createAction } from "typesafe-actions";
-import { DistrictId, GeoUnits } from "../../shared/entities";
+import { DistrictId, EvaluateMetric, GeoUnits } from "../../shared/entities";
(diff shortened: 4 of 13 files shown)


src/server/src/organizations/services/organizations.service.ts
```diff
@@ -12,6 +13,43 @@ export class OrganizationsService extends TypeOrmCrudService<Organization> {
...
+        "regionConfig.id",
+        "regionConfig.countryCode",
+        "regionConfig.regionCode",
+        "projectTemplates.numberOfDistricts",
+        "projectTemplates.description",
+        "projectTemplates.details",
+        "projectTemplates.isActive"
+      ])
+      .getOne();
+    return data;
```
maurizi, line 53: @kevinearldenny I don't understand why this function needed to be introduced for this change, or why it selects all of the fields that it does.

Is this intended to replace the options in the `@Crud` decorator? If so why are we still using the decorator?

kevinearldenny: Yes, the intention was to replace the `@Crud` decorator since I don't believe we are able to add the `isActive = TRUE` filter on the existing `projectTemplate`  join within the Crud schema. I tried removing the CRUD schema but it appeared that the endpoint stopped functioning after it was removed.

maurizi: Alright - I did some brief investigation and it does indeed seems like `nestjsx-crud` literally doesn't have the ability to set custom `ON` clauses and also it can't fetch the the `ProjectTemplate` items in a separate query (like Django's `prefetch_related`) so it seems worthwhile to just ditch it entirely for this relatively simple controller, as you've done.

Additionally, the query is not quite right. We should be doing `LEFT JOIN`s  not `INNER JOIN`s, the `isActive` filter needs to be in the `ON` clause not the `WHERE` clause (or it will return 0 rows for organizations with no active templates), and while it was already there, I can't see why we'd need to expose the list of users on an organization or the organizations admin on a public endpoint - those fields don't get used by the `OrganizationScreen` component anyways as far as I can tell.

```diff
diff --git a/src/server/src/organizations/controllers/organizations.controller.ts b/src/server/src/organizations/controllers/organizations.controller.ts
index ba586a5..4c98f6c 100644
--- a/src/server/src/organizations/controllers/organizations.controller.ts
+++ b/src/server/src/organizations/controllers/organizations.controller.ts
@@ -1,8 +1,6 @@
-import { Controller, UseGuards, Param, Post, NotFoundException, Body } from "@nestjs/common";
-import { Crud, CrudController, Override } from "@nestjsx/crud";
-import { IsNotEmpty } from "class-validator";
+import { Controller, UseGuards, Param, Post, NotFoundException, Body, Get } from "@nestjs/common";
 
-import { OrganizationSlug, PublicUserProperties, UserId } from "../../../../shared/entities";
+import { OrganizationSlug, UserId } from "../../../../shared/entities";
 import { JoinOrganizationErrors } from "../../../../shared/constants";
 
 import { JwtAuthGuard, OptionalJwtAuthGuard } from "../../auth/guards/jwt-auth.guard";
@@ -13,41 +11,8 @@ import { Organization } from "../entities/organization.entity";
 import { OrganizationUserDto } from "../entities/organizationUser.dto";
 import { OrganizationsService } from "../services/organizations.service";
 
-@Crud({
-  model: {
-    type: Organization
-  },
-  routes: {
-    only: ["getOneBase"]
-  },
-  params: {
-    slug: {
-      field: "slug",
-      type: "string",
-      primary: true
-    }
-  },
-  query: {
-    join: {
-      projectTemplates: {
-        eager: true
-      },
-      "projectTemplates.regionConfig": {
-        eager: true
-      },
-      users: {
-        allow: ["id", "name"] as PublicUserProperties[],
-        eager: true
-      },
-      admin: {
-        eager: true
-      }
-    }
-  }
-})
 @Controller("api/organization")
-// @ts-ignore
-export class OrganizationsController implements CrudController<Organization> {
+export class OrganizationsController {
   constructor(public service: OrganizationsService, private readonly usersService: UsersService) {}
 
   async getOrg(organizationSlug: OrganizationSlug): Promise<Organization> {
@@ -69,8 +34,8 @@ export class OrganizationsController implements CrudController<Organization> {
     return user;
   }
 
-  @Override()
   @UseGuards(OptionalJwtAuthGuard)
+  @Get(":slug/")
   async getOne(@Param("slug") slug: OrganizationSlug): Promise<Organization> {
     return this.getOrg(slug);
   }
diff --git a/src/server/src/organizations/services/organizations.service.ts b/src/server/src/organizations/services/organizations.service.ts
index 85b2985..d6f7204 100644
--- a/src/server/src/organizations/services/organizations.service.ts
+++ b/src/server/src/organizations/services/organizations.service.ts
@@ -20,12 +20,13 @@ export class OrganizationsService extends TypeOrmCrudService<Organization> {
   async getOrgAndProjects(slug: string): Promise<Organization | undefined> {
     const builder = this.repo.createQueryBuilder("organization");
     const data = await builder
-      .innerJoinAndSelect("organization.users", "users")
-      .innerJoinAndSelect("organization.projectTemplates", "projectTemplates")
-      .innerJoinAndSelect("organization.admin", "admin")
-      .innerJoinAndSelect("projectTemplates.regionConfig", "regionConfig")
-      .where("projectTemplates.isActive = TRUE")
-      .andWhere("organization.slug = :slug", { slug: slug })
+      .leftJoinAndSelect(
+        "organization.projectTemplates",
+        "projectTemplates",
+        "projectTemplates.isActive = TRUE"
+      )
+      .leftJoinAndSelect("projectTemplates.regionConfig", "regionConfig")
+      .where("organization.slug = :slug", { slug: slug })
       .select([
         "organization.id",
         "organization.slug",
@@ -34,10 +35,6 @@ export class OrganizationsService extends TypeOrmCrudService<Organization> {
         "organization.linkUrl",
         "organization.municipality",
         "organization.region",
-        "users.id",
-        "users.name",
-        "admin.id",
-        "admin.name",
         "projectTemplates.id",
         "projectTemplates.name",
         "regionConfig.name",

```

kevinearldenny: @maurizi The list of users for the organization is used on the Organization Screen - it is [how the frontend identifies ](https://github.com/PublicMapping/districtbuilder/blob/1008e781a2c583481f8b0a88f9641c94d2a0e890/src/client/screens/OrganizationScreen.tsx#L112)whether the current user is a member of the organization or not. 

Keep in mind that this endpoint is also used on the OrganizationAdminScreen, and the nested admin value in `organization.resource` is [used to gate ](https://github.com/PublicMapping/districtbuilder/blob/1008e781a2c583481f8b0a88f9641c94d2a0e890/src/client/screens/OrganizationAdminScreen.tsx#L120) access to this screen for only Org Admins. 

That being said - I do agree that it makes sense to strip out the `nestjsx-crud` stuff here, and the other query improvements look good.

src/server/src/organizations/services/organizations.service.ts
```diff
@@ -12,6 +13,46 @@ export class OrganizationsService extends TypeOrmCrudService<Organization> {
...
   save(org: Organization): Promise<Organization> {
+    // @ts-ignore
     return this.repo.save(org);
   }
+
+  async getOrgAndProjects(slug: string): Promise<Organization | undefined> {
+    const builder = this.repo.createQueryBuilder("organization");
+    const data = await builder
+      .innerJoinAndSelect("organization.users", "users")
+      .innerJoinAndSelect("organization.admin", "admin")
```
maurizi, line 24: These should both be `leftJoinAndSelect` as well.
```suggestion
      .leftJoinAndSelect("organization.users", "users")
      .leftJoinAndSelect("organization.admin", "admin")
```

src/server/src/project-templates/services/project-templates.service.ts
```diff
@@ -19,6 +19,7 @@ export class ProjectTemplatesService extends TypeOrmCrudService<ProjectTemplate>
       .innerJoinAndSelect("projectTemplate.projects", "projects")
       .innerJoinAndSelect("projects.user", "user")
       .where("organization.slug = :slug", { slug: slug })
+      .where("projectTemplate.isActive = :true", { true: true })
```
maurizi, line 22: From the documentation for `.where(..)`:
> If you had previously WHERE expression defined, calling this function will override previously set WHERE conditions.

You can get around this by using `.andWhere(...)` instead, or combining this into the above `.where(...)`. Also an additional cleanup - no need to pass in `true` as a parameter - just use the sql keyword `TRUE`.

```suggestion
      .where("organization.slug = :slug", { slug: slug })
      .andWhere("projectTemplate.isActive = TRUE")
```

or maybe 

```suggestion
      .where("organization.slug = :slug AND projectTemplate.isActive = TRUE", { slug: slug })
```

maurizi (changes requested):
Almost there! I left 1 comment that I'd like to see addressed but everything else is looking good.

maurizi (approved):
🎉

PR #634, opened 2021-03-18
Fix showing join organization for logged in users

maurizi:
## Overview

Removes seemingly unneeded useEffect that would hide the modal immediately after it was shown.

This was introduced in https://github.com/PublicMapping/districtbuilder/commit/32ffcb6cf0e8fdfd6391941503b49cc90fcdee57 though I can't figure out what it was intended to do.

## Testing Instructions

- Log in as a verified user
- Go to an organization you are not a member of
- Try and join - on `develop` the modal will open and then immediately close, on this branch it will open and stay open


kevinearldenny (approved):
Looks good 👍

PR #642, opened 2021-03-23
Refactor evaluate mode

kevinearldenny:
## Overview

- Move files for Project Evaluate mode into the `evaluate` folder
- Splits the `ProjectEvaluateSidebar` component into a wrapper component, and puts info display from sidebar into new `ProjectEvaluateSummary` component
- Displays computed County Split metric in `ProjectSummary` to show an example of the computation / flow of data within these components

### Checklist

## Testing Instructions

- ./scripts/server
- Prefer bulleted description


src/client/components/evaluate/ProjectEvaluateSummary.tsx
```diff
@@ -0,0 +1,174 @@
...
+};
+
+const ProjectEvaluateView = ({
+  projectMetricValues,
+  requiredMetrics,
+  optionalMetrics
+}: {
+  // Still working out what the type should be here
+  // eslint-ignore
+  readonly projectMetricValues: any;
```
kevinearldenny, line 79: @pcaisse Not entirely sure what the type should be here, since it's a keyed object with a fairly flexible structure

src/client/screens/ProjectScreen.tsx
```diff
@@ -34,7 +34,7 @@ import store from "../store";
 import { useBeforeunload } from "react-beforeunload";
 import PageNotFoundScreen from "./PageNotFoundScreen";
 import SiteHeader from "../components/SiteHeader";
-import ProjectEvaluateSidebar from "../components/ProjectEvaluateSidebar";
+import ProjectEvaluateView from "../components/evaluate/ProjectEvaluateSidebar";
```
pcaisse, line 37: Should this still be called `ProjectEvaluateSidebar`?

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -0,0 +1,152 @@
...
+    }
+  };
+  return (
+    <Container sx={style.sidebar} className="evaluate-sidebar">
+      {!metric ? (
+        <ProjectEvaluateSummary
+          projectMetricValues={projectMetricValues}
+          requiredMetrics={requiredMetrics}
+          optionalMetrics={optionalMetrics}
+        ></ProjectEvaluateSummary>
```
pcaisse, line 144:
```suggestion
        />
```

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -0,0 +1,152 @@
...
+  metric,
+  project
+}: {
+  readonly geojson?: DistrictsGeoJSON;
+  readonly metric: EvaluateMetric | undefined;
+  readonly project?: IProject;
+}) => {
+  const countySplits = project?.districtsDefinition.filter(x => !Array.isArray(x)).length;
+  const totalCounties = project ? project.districtsDefinition : 0;
+  const projectMetricValues = {
```
pcaisse, line 131: I think `projectMetricValues` could serve as our source of truth now (possibly with dummy data still included for the metrics for which do not yet have real values). That way we can avoid doing the key lookup which may return undefined if we make a mistake and then we have to deal with the undefined case. (In practice this may actually shake out as getting rid of `projectMetricValues` and keeping `requiredMetrics` and `optionalMetrics` with real data in them since we care about the required/optional distinction for display purposes.) The overall point is I think we should avoid doing the lookup by key and instead have the data and metadata together in one object so that if we have a metric key we'll also have the data associated with it. It seems to me that should be workable since this component is both deriving the real data from the project and also currently injecting the dummy data so we should have everything we need.

src/client/components/evaluate/ProjectEvaluateSummary.tsx
pcaisse: See above comment

pcaisse (changes requested):

src/shared/entities.d.ts
```diff
@@ -16,11 +16,14 @@ export interface IUser {
...
 
 export type EvaluateMetric = {
+  readonly key: string;
   readonly name: string;
   readonly type?: "fraction" | "percent" | "count";
   readonly value?: number;
+  readonly total?: number;
   readonly status?: boolean;
   readonly description: string;
+  readonly splitCounties?: readonly (GeoUnitCollection | undefined)[] | undefined;
```
kevinearldenny, line 26: @pcaisse Does this seem sustainable to imagine all of the different permutations for the 6 evaluation metrics to be coerced into `EvaluateMetric` like this? Or is there another way you would recommend structuring this to be more extensible?

pcaisse: So with the placeholder data it's a little tricky to imagine what is actually required for the summary and list views. It seems like `EvaluateMetric` is pulling double duty for both of those views here. It's definitely possible in TS to say "allow any other stuff you want" for an interface like so where the "other stuff" would be what's required for the detail view:
```ts
export type EvaluateMetric = {
  readonly key: string;
  readonly name: string;
  readonly type?: "fraction" | "percent" | "count";
  readonly value?: number;
  readonly status?: boolean;
  readonly description: string;
  [key: string]: any;
};
```
That way the fields that are explicitly called out in `EvaluateMetric` are required for summary and we know they'll be there, so that covers our bases for the summary. For the detail view, it seems like each metric will require different pieces of data, or possibly different components that could be stuffed into `EvaluateMetric`, but we can figure that out when we come to it.

kevinearldenny: Hmmm, this appears to be giving me two lint errors:

```
  28:3   error    Only the same kind of members allowed in types  functional/no-mixed-type
  28:27  warning  Unexpected any. Specify a different type        @typescript-eslint/no-explicit-any
```

pcaisse: You can just throw an `// eslint-disable-next-line` on the line before it for now.

pcaisse (approved):

PR #644, opened 2021-03-24
Display Evaluate Mode metric view for County Splits

### LINKED ISSUE #608: Evaluate mode county split
image

Notes:
We should try and be generic here instead of hardcoding county - so this can be "Ward split" mode for Dane County.

AC:
 - Update evaluate mode summary sidebar to display county splits. This should be possible to determine by seeing if the `districtsDefinition` has a `number` or `array` type at the top-most (county) level.
 - Show county split layer, map legend, and sidebar legend when switching to county split detail

kevinearldenny:
## Overview

- Displays the number of counties / top-level geounits that are split across multiple districts in `ProjectEvaluateSummary`
- Adds a new sidebar display, `CountySplit`, with a list of all counties / geo-units* by name and a marker identifying the counties that are split across multiple districts
- Updates the map view when the County Split metric is selected to display a layer of the counties that are split across multiple districts
- Adds a legend to the map

### Checklist

### Demo

image

### Notes

* For Dane County, do the blockgroups have a `name` property?
~** I'm encountering a slight bug here where the county names are lost if you close the metric and re-open it... I know that there is a React Hook-based solution I could employ here, but figured I'd get a draft PR up now for you to take a look~ Fixed 💯 , now ready for review

## Testing Instructions

- `./scripts/server`
- Select a project that has counties split across multiple districts, and click 'Evaluate'
- Expect: County Split metric is displayed with a real computed value
- Click the 'County Split' metric
- Expect: County Split view is displayed as above

Closes #608

### DIFF
--- a/src/client/components/evaluate/detail/CountySplit.tsx
+++ b/src/client/components/evaluate/detail/CountySplit.tsx
@@ -1,0 +1,102 @@
+/** @jsx jsx */
+import { Box, Flex, jsx, ThemeUIStyleObject, Heading } from "theme-ui";
+import {
+  EvaluateMetricWithValue,
+  IProject,
+  IStaticMetadata,
+  RegionLookupProperties
+} from "../../../../shared/entities";
+import { useState, useEffect } from "react";
+import { getLabelLookup } from "../../map/labels";
+import { Resource } from "../../../resource";
+import { geoLevelLabel } from "../../../functions";
+
+const style: ThemeUIStyleObject = {
+  td: {
+    fontWeight: "body",
+    color: "gray.8",
+    fontSize: 1,
+    p: 2,
+    textAlign: "left",
+    verticalAlign: "bottom",
+    minWidth: "20px",
+    maxWidth: "200px",
+    mr: "10px"
+  },
+  fillBox: {
+    height: "10px",
+    width: "10px",
+    background: "#fed8b1",
+    opacity: "0.5",
+    outline: "1px solid black"
+  },
+  unfilledBox: {
+    height: "10px",
+    width: "10px",
+    background: "none",
+    outline: "1px solid black"
+  }
+};
+
+type CountyLookup = ReadonlyArray<string | undefined>;
+
+const CountySplitMetricDetail = ({
+  metric,
+  project,
+  staticMetadata,
+  regionProperties,
+  geoLevel
+}: {
+  readonly metric: EvaluateMetricWithValue;
+  readonly project?: IProject;
+  readonly geoLevel: string;
+  readonly regionProperties: Resource<readonly RegionLookupProperties[]>;
+  readonly staticMetadata?: IStaticMetadata;
+}) => {
+  const [countyLookup, setCountyLookup] = useState<CountyLookup | undefined>(undefined);
+
+  useEffect(() => {
+    if ("resource" in regionProperties && !countyLookup) {
+      setCountyLookup(
+        regionProperties.resource.map(c => (typeof c.name === "string" ? c.name : undefined))
+      );
+    }
+  }, [regionProperties, countyLookup]);
+
+  return (
+    <Box sx={{ ml: "20px", overflowY: "scroll" }}>
+      <Heading as="h4" sx={{ variant: "text.h4", display: "block" }}>
+        {metric.value} / {metric.total} {geoLevelLabel(geoLevel)} {metric.description}
+      </Heading>
+      <Flex sx={{ flexDirection: "column", width: "60%" }}>
+        <table>
+          <tbody>
+            {countyLookup ? (
+              project?.districtsDefinition.map((d, id) => (
+                <tr key={id}>
+                  <td sx={style.tableElement}>
+                    {countyLookup && id in countyLookup && staticMetadata
+                      ? getLabelLookup(geoLevel, countyLookup[id])
+                      : staticMetadata
+                      ? getLabelLookup(geoLevel, undefined, id)
+                      : ""}
+                  </td>
+                  <td sx={style.tableElement}>
+                    <Box sx={Array.isArray(d) ? style.fillBox : style.unfilledBox}></Box>
+                  </td>
+                  <td sx={style.tableElement}>{Array.isArray(d) ? "Split" : "Not split"}</td>
+                </tr>
+              ))
+            ) : (
+              <tr>
+                <td>Loading...</td>
+              </tr>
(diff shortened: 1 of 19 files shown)


src/client/components/evaluate/detail/CountySplit.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <Box sx={{ ml: "20px", overflowY: "scroll" }}>
+      <Heading as="h4" sx={{ variant: "text.h4", display: "block" }}>
+        {metric.value} / {metric.total} counties {metric.description}
+      </Heading>
+      <Flex sx={{ flexDirection: "column", width: "60%" }}>
+        <table>
+          <tbody>
+            {project?.districtsDefinition.map((d, id) => (
+              <tr key={id}>
+                <td sx={style.tableElement}>{id in countyLookup ? countyLookup[id] : id}</td>
```
maurizi, line 84: We probably want to follow the same naming scheme we use for the tooltip (and ideally refactor this to a shared function): https://github.com/PublicMapping/districtbuilder/blob/49572073b521cf436cc028061a7c5dc522c4d2a4/src/client/components/map/MapTooltip.tsx#L149-L159

maurizi (changes requested):
As per the issue, we should gracefully handle cases like Dane County where the top-most geo-level is not "County" - so this should say "Ward splits" not "County" splits, etc.:

localhost_3003_projects_071d8fa3-b459-4bc8-85bb-159a5007cf1f

maurizi:
> @maurizi How would you recommend parsing this out from the available data on project?

You can look at the `id`  field of the objects in the `geoLevels` or `geoLevelHierarchy` properties of `statisMetadata` - on most projects the "top-most" level will be `county`, but for Dane it would be `ward`: https://github.com/PublicMapping/districtbuilder/blob/e448364cedfb2087d8ee5f6fe4d3be2765e8aff5/src/shared/entities.d.ts#L115-L120

There is a shared function you can use for pluralization: https://github.com/PublicMapping/districtbuilder/blob/e448364cedfb2087d8ee5f6fe4d3be2765e8aff5/src/client/functions.ts#L130-L137

src/client/components/evaluate/detail/CountySplit.tsx
```diff
@@ -0,0 +1,112 @@
...
+    height: "10px",
+    width: "10px",
+    background: "none",
+    outline: "1px solid black"
+  }
+};
+
+interface CountyLookup {
+  readonly [key: number]: string;
+}
```
maurizi, line 43: We can make this quite a bit simpler by using an array type here instead:
```diff
diff --git a/src/client/components/evaluate/detail/CountySplit.tsx b/src/client/components/evaluate/detail/CountySplit.tsx
index b71bdae..dadf89e 100644
--- a/src/client/components/evaluate/detail/CountySplit.tsx
+++ b/src/client/components/evaluate/detail/CountySplit.tsx
@@ -38,9 +38,7 @@ const style: ThemeUIStyleObject = {
   }
 };
 
-interface CountyLookup {
-  readonly [key: number]: string;
-}
+type CountyLookup = ReadonlyArray<string | undefined>;
 
 const CountySplitMetricDetail = ({
   metric,
@@ -59,15 +57,7 @@ const CountySplitMetricDetail = ({
 
   useEffect(() => {
     if ("resource" in regionProperties && !countyLookup) {
-      // eslint-disable-next-line
-      let lookup: any = {};
-      regionProperties.resource.forEach((c, i) => {
-        if (c.name) {
-          // eslint-disable-next-line
-          lookup[i] = c.name;
-        }
-      });
-      setCountyLookup(lookup);
+      setCountyLookup(regionProperties.resource.map(c => c.name));
     }
   }, [regionProperties, countyLookup]);
 
@@ -84,11 +74,7 @@ const CountySplitMetricDetail = ({
                 <tr key={id}>
                   <td sx={style.tableElement}>
                     {countyLookup && id in countyLookup && staticMetadata
-                      ? getLabelLookup(
-                          geoLevel,
-                          // @ts-ignore
-                          countyLookup[id]
-                        )
+                      ? getLabelLookup(geoLevel, countyLookup[id])
                       : staticMetadata
                       ? getLabelLookup(geoLevel, undefined, id)
                       : ""}
```

src/shared/entities.d.ts
```diff
@@ -266,3 +266,6 @@ export type LockedDistricts = readonly boolean[];
 
 export type UintArray = Uint8Array | Uint16Array | Uint32Array;
 export type UintArrays = ReadonlyArray<UintArray>;
+export interface RegionLookupProperties {
+  readonly name?: string;
+}
```
maurizi, line 271: We should make this match the type being used in the frontend (which is also more accurate as to what the API can return), and make use of it in `src/client/api.ts` :

```suggestion
export type RegionLookupProperties = Record<string, unknown>;
```

maurizi (changes requested):
Functioning well, just a few small cleanups left 😄

src/client/components/evaluate/detail/CountySplit.tsx
```diff
@@ -0,0 +1,101 @@
...
+  readonly geoLevel: string;
+  readonly regionProperties: Resource<readonly RegionLookupProperties[]>;
+  readonly staticMetadata?: IStaticMetadata;
+}) => {
+  const [countyLookup, setCountyLookup] = useState<CountyLookup | undefined>(undefined);
+
+  useEffect(() => {
+    if ("resource" in regionProperties && !countyLookup) {
+      // @ts-ignore
+      setCountyLookup(regionProperties.resource.map(c => c.name));
```
maurizi, line 61: We can appease the compiler here w/ some runtime type checking:
```suggestion
      setCountyLookup(regionProperties.resource.map(c => typeof c.name === "string" ? c.name : undefined));
```

src/client/components/map/Map.tsx
```diff
@@ -230,6 +248,35 @@ const DistrictsMap = ({
...
+      return c;
+    } else {
+      return undefined;
+    }
+  });
+
+  // Set features from topmost geolayer that are split across districts to be selected
+  // eslint-disable-next-line
+  const counties =
+    map &&
```
maurizi, line 263: We never use counties?
```suggestion
  map &&
```

maurizi (approved):
👍 I left a few small comments but they're really tiny.

Great job, there were a lot of moving pieces in this 🎉

PR #648, opened 2021-03-25
Add Project Evaluate view for contiguity

### LINKED ISSUE #605: Evaluate mode contiguity
image

AC
 - Update evaluate mode summary sidebar to display contiguity value and ✅ or ❌
 - Show contiguity layer, map legend, and sidebar legend when switching to contiguity detail

kevinearldenny:
## Overview

Adds project evaluate mode view for contiguity.

- Add Contiguity detail
- Display contiguity on map
- Add map legend for contiguity

### Checklist

### Demo

image

## Testing Instructions

- ./scripts/server
- Open a project and click evaluate
- Click 'Contiguity'
- Expect: Contiguity view is rendered as pictured

Closes #605


src/client/reducers/districtDrawing.ts
```diff
@@ -266,6 +266,7 @@ const districtDrawingReducer: LoopReducer<ProjectState, Action> = (
     case getType(toggleEvaluate):
       return {
         ...state,
+        evaluateMetric: undefined,
```
kevinearldenny, line 269: I added in this logic here to clear the selected metric when evaluate mode is toggled off, but I'm not totally sure this is how it's supposed to be. @maurizi , any thoughts here?

maurizi:
@kevinearldenny it looks like this has a lot of conflicts both with `develop` and likely also quite a few w/ #644 - I'm going to hold off on review for now. Let me know when its ready to review.

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -103,14 +103,18 @@ const ProjectEvaluateSidebar = ({
     {
       key: "contiguity",
       name: "Contiguity",
-      status: true,
+      status:
+        geojson?.features.filter(f => f.properties.contiguity === "non-contiguous").length === 0
+          ? true
+          : false,
```
maurizi, line 109: In general, any code of the type `let my_var = some_boolean_expression ? true : false` can **always** be simplified to just `let my_var = some_boolean_expression`.

This is a similar code smell to things like:
```
if (some_boolean_expression) {
  return true;
} else {
  return false;
}
```

Which can always be simplified to simply: `return some_boolean_expression;`.


So in this specific case:
```suggestion
        geojson?.features.filter(f => f.properties.contiguity === "non-contiguous").length === 0,
```

src/client/components/map/Map.tsx
```diff
@@ -92,13 +97,21 @@ const style: ThemeUIStyleObject = {
...
     display: "inline-block",
     outline: "1px solid gray",
     padding: "20px"
   }
+  // legendTitle: {
+  //     display: "inline-block",
+  //     width: "120px",
+  //     fontWeight: "600",
+  //     mr: "40px"
+  // }
```
maurizi, line 114: ?

```suggestion
```

src/client/reducers/districtDrawing.ts
maurizi: I'm honestly not sure - @jfrankl What do you think?

If I go into Evaluate mode, go into a detail screen like Contiguity, and then I toggle Evaluate Mode off and then on again what do I see?
 - The Evaluate Summary view
 - The Evaluate Contiguity detail view

src/client/components/evaluate/detail/Contiguity.tsx
```diff
@@ -0,0 +1,97 @@
...
+}) => {
+  function computeRowFill(row: DistrictProperties) {
+    return row.contiguity === "contiguous" ? CONTIGUITY_FILL_COLOR : EVALUATE_GRAY_FILL_COLOR;
+  }
+  return (
+    <Box sx={{ ml: "20px", overflowY: "scroll" }}>
+      <Heading as="h4" sx={{ variant: "text.h4", display: "block", width: "100%" }}>
+        {metric.value} of {metric.total} districts are contiguous
+      </Heading>
+      <Flex sx={{ flexDirection: "column", width: "60%" }}>
```
maurizi, line 56: This looked better for me (why would we want 40% of the sidebar to be blank?):
```suggestion
      <Flex sx={{ flexDirection: "column", width: "100%" }}>
```

src/client/components/map/Map.tsx
```diff
@@ -507,6 +527,37 @@ const DistrictsMap = ({
...
+          <Box sx={style.legendItem}>
+            <Box
+              sx={{
+                display: "inline-block",
+                mr: "10px",
+                width: "20px",
+                height: "20px",
+                opacity: "0.9",
+                backgroundColor: CONTIGUITY_FILL_COLOR
+              }}
```
maurizi, line 542: You should extract these shared styles to make things a little more DRY:
```typescript
              sx={{
                ...style.legendColorSwatch,
                backgroundColor: CONTIGUITY_FILL_COLOR
              }}
```

src/client/components/evaluate/detail/Contiguity.tsx
```diff
@@ -0,0 +1,97 @@
...
+                          height: "20px",
+                          opacity: "0.9",
+                          backgroundColor: computeRowFill(feature.properties)
+                        }}
+                      ></Box>
+                      <Box sx={style.compactnessItem}>
+                        {feature.properties.contiguity === "contiguous"
+                          ? "Contiguous"
+                          : "Non-contiguous"}
+                      </Box>
```
maurizi, line 85: We should handle empty districts as well:
```suggestion
                      {feature.properties.contiguity ? (
                        <React.Fragment>
                          <Box
                            sx={{
                              display: "inline-block",
                              mr: "40px",
                              width: "20px",
                              height: "20px",
                              opacity: "0.9",
                              backgroundColor: computeRowFill(feature.properties)
                            }}
                          ></Box>
                          <Box sx={style.compactnessItem}>
                            {feature.properties.contiguity === "contiguous"
                              ? "Contiguous"
                              : "Non-contiguous"}
                          </Box>
                        </React.Fragment>
                      ) : (
                        <Box sx={style.compactnessItem}>Empty</Box>
                      )}
```

maurizi (changes requested):
Looking pretty good, just a few small cleanups needed 👍

src/client/reducers/districtDrawing.ts
jfrankl: Thank you for checking in, this is a great question. I think we should have the user get dropped back in the section they were in. So, in your case, they would see the "The Evaluate Contiguity detail view" when they re-enter evaluate mode. My thinking is that this would allow a user to toggle back and forth between the two modes more seamlessly; you could see a particular metric, go back to editing your maps, and then quickly get back to the same metric.

maurizi: @kevinearldenny it sounds like answer is we don't want to include this change:
```suggestion
```

maurizi (approved):
_edit_ nevermind

src/client/components/map/Map.tsx
maurizi: @kevinearldenny Why did you mark this as resolved?

maurizi (changes requested):

src/client/components/map/Map.tsx
kevinearldenny: Whoops... I think I made this refactor on the other branch but not here. Should be fixed now.

maurizi (approved):

PR #669, opened 2021-04-01
Add button to link to admin screen on org page

### LINKED ISSUE #653: Add button to navigate to admin maps list
Currently we have an /admin page for an organization that lists all the maps that are associated with the Organization. However, there's no way to navigate there from the Organization page. We should add a button that is only visible to admins on the Organization page so they can navigate to the /admin page.

AC:
* There is a button that says "View All" on the organization page, next to the Featured Maps section, that is visible to the organization admin, that navigates to the /admin page.

kevinearldenny:
## Overview

- Adds a 'View all' button to Org Screen visible only for org admins to link to the Org Admin screen
- Adds a link on the Org Admin screen to navigate back to the public Org page

### Checklist

### Demo

### Notes

## Testing Instructions

- Login as an org admin user and navigate to org page
- Expect: Button appears enabling you to navigate to admin screen
- Login as a non-admin user and navigate to org page
- Expect: Button is not visible

Closes #653

### DIFF
--- a/src/client/screens/OrganizationAdminScreen.tsx
+++ b/src/client/screens/OrganizationAdminScreen.tsx
@@ -1,7 +1,7 @@
 /** @jsx jsx */
 import { useEffect } from "react";
 import { connect } from "react-redux";
-import { useParams } from "react-router-dom";
+import { useParams, Link } from "react-router-dom";
 import { Box, Flex, Heading, jsx } from "theme-ui";
 import { organizationFetch } from "../actions/organization";
 import { organizationProjectsFetch } from "../actions/organizationProjects";
@@ -126,7 +126,9 @@ const OrganizationAdminScreen = ({ organization, user, organizationProjects }: S
           <Box>
             <Flex sx={style.header}>
               <Box>
-                <Heading as="h3">{organization.resource.name}</Heading>
+                <Heading as="h3">
+                  <Link to={`/o/${organizationSlug}`}>{organization.resource.name}</Link>
+                </Heading>
                 <Heading>Maps</Heading>
                 <Box>
                   Published maps that were created by members of your organization. You can select
--- a/src/client/screens/OrganizationScreen.tsx
+++ b/src/client/screens/OrganizationScreen.tsx
@@ -1,8 +1,8 @@
 /** @jsx jsx */
 import { useEffect, useState } from "react";
 import { connect } from "react-redux";
-import { useParams } from "react-router-dom";
-import { Box, Button, Flex, Heading, Image, jsx, Link, Spinner, Text } from "theme-ui";
+import { useParams, Link as RouterLink } from "react-router-dom";
+import { Box, Button, Flex, Heading, Image, jsx, Spinner, Text, Link } from "theme-ui";
 import { formatDate } from "../functions";
 
 import { showCopyMapModal } from "../actions/districtDrawing";
@@ -115,6 +115,10 @@ const style = {
     width: "100%",
     background: "lightgray",
     color: "black"
+  },
+  viewAllBtn: {
+    fontSize: "14pt",
+    ml: 4
   }
 } as const;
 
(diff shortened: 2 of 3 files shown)


maurizi (changes requested):
@kevinearldenny Can you please rework this to use the `<Link>` component from `theme-ui`? See the `ProjectHeader` component for an example of it in use:
https://github.com/PublicMapping/districtbuilder/blob/4cd51f15966cfdc273b451d4e00547e5efb1b299/src/client/components/ProjectHeader.tsx#L73-L82

In general it is fine to use programmatic manipulation of `history` when required (say after an AJAX call, or because you want to disable the link in certain application states and so would prefer to use a `<button>` over an `<a>`), but we should strive to use basic HTML components like anchor tags when possible - it is better for things like screen readers and enables built-in web behaviors like CTRL-clicking to open in a new tab.

src/client/screens/OrganizationAdminScreen.tsx
```diff
@@ -126,7 +132,13 @@ const OrganizationAdminScreen = ({ organization, user, organizationProjects }: S
...
             <Flex sx={style.header}>
               <Box>
-                <Heading as="h3">{organization.resource.name}</Heading>
+                <Heading
+                  as="h3"
+                  onClick={() => history.push(`/o/${organizationSlug}`)}
+                  sx={style.orgName}
+                >
+                  {organization.resource.name}
+                </Heading>
```
maurizi, line 141: We can use a `<Link>` here too, and it will eliminate the need for specialized styling as a bonus:
```suggestion
                <Heading
                  as="h3"
                >
                  <Link to={`/o/${organizationSlug}`}>{organization.resource.name}</Link>
                </Heading>
```

src/client/screens/OrganizationScreen.tsx
```diff
@@ -294,7 +307,22 @@ const OrganizationScreen = ({ organization, organizationProjects, user }: StateP
...
+                        "resource" in organization &&
+                        organization.resource &&
+                        userIsOrgAdmin(organization.resource, user.resource) && (
+                          <RouterLink
+                            to={orgAdminUrl}
+                            sx={{
+                              "&:focus": { outline: "none", boxShadow: "focus" }
+                            }}
+                          >
+                            <Button sx={style.viewAllBtn}>View All</Button>
```
maurizi, line 322: It is invalid HTML to place a `<button>` inside an `<a>` tag (see MDN for what content is valid: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties).

I'm guessing you did this for styling purposes? We have an existing variant in our Theme for making links look like buttons:
```diff
diff --git a/src/client/screens/OrganizationScreen.tsx b/src/client/screens/OrganizationScreen.tsx
index 615233b..0c4a4d6 100644
--- a/src/client/screens/OrganizationScreen.tsx
+++ b/src/client/screens/OrganizationScreen.tsx
@@ -118,9 +118,7 @@ const style = {
   },
   viewAllBtn: {
     fontSize: "14pt",
-    position: "absolute",
-    right: "40%",
-    width: "120px"
+    ml: 4
   }
 } as const;
 
@@ -315,11 +313,9 @@ const OrganizationScreen = ({ organization, organizationProjects, user }: StateP
                         userIsOrgAdmin(organization.resource, user.resource) && (
                           <RouterLink
                             to={orgAdminUrl}
-                            sx={{
-                              "&:focus": { outline: "none", boxShadow: "focus" }
-                            }}
+                            sx={{ ...style.viewAllBtn, variant: "links.button" }}
                           >
-                            <Button sx={style.viewAllBtn}>View All</Button>
+                            View all
                           </RouterLink>
                         )}
                     </span>
```

maurizi (changes requested):

src/client/screens/OrganizationAdminScreen.tsx
```diff
@@ -72,6 +72,11 @@ const style = {
   templates: {
     p: 5
   },
+  orgName: {
+    "&:hover": {
+      cursor: "pointer"
+    }
+  },
```
maurizi, line 79: This is unused now.
```suggestion
```

maurizi (approved):
One last little cleanup left but that's it 🧹

PR #671, opened 2021-04-02
Dont allow featuring unpublished maps

### LINKED ISSUE #635: Organization administrators should not be able to set unpublished maps as featured
They should be visible in the list of maps, but the feature button should not be shown:
image

maurizi:
## Overview

Prevents organization admins from featuring unpublished maps.

Also refactors to remove the client-only `OrgProject` type in favor of the shared (and more accurate) `ProjectNest` type.

### Checklist

### Demo

localhost_3003_o_azavea_admin

## Testing Instructions

- `scripts/server`
- Go to the org admin page - all maps listed there previously should still be there, but the Feature/Unfeature button should only be visible if the map is published

Closes #635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add paintbrush drawing tool [#611](https://github.com/PublicMapping/districtbuilder/pull/611)
 - Add verify and join link to user verification email [#616](https://github.com/PublicMapping/districtbuilder/pull/616)
 - Display organizations dropdown in header [#620](https://github.com/PublicMapping/districtbuilder/pull/620)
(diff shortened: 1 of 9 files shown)


kevinearldenny:
Looks good to me, once you resolve the merge conflict 👍 

One note that probably isn't that big of a deal, but worth noting... if you had an unpublished project featured before this change, there is now no longer a way to unfeature it. It might be overkill, but do you think it might make sense to filter on visibility in `OrganizationScreen`?

kevinearldenny (approved):

maurizi:
> One note that probably isn't that big of a deal, but worth noting... if you had an unpublished project featured before this change, there is now no longer a way to un-feature it. It might be overkill, but do you think it might make sense to filter on project visibility for featured projects in OrganizationScreen?

I thought about it, but we haven't released the feature yet to production so I figured it's fine 🤷‍♂️

PR #672, opened 2021-04-02
Allow admins to view projects created from organization template

### LINKED ISSUE #665: Make Org Admins able to view all maps tied to templates
We want to enable org admins to review the maps their organization's users are drawing. Since Private is the default option, a vast majority of the maps will stay private and the admin will not be able to view them. As an initial step, we want to make it possible to for the org admin to view all maps in their org regardless of "Share" setting. This may become more nuanced as we implement submission of maps.

AC:
* The organization admin can view all maps that have been drawn using templates associated with their org, regardless of "Share" setting

pcaisse:
## Overview
Allow admins to view projects created from project template for organizations they administer.

### Checklist

### Demo
db_admin_view_private_map

## Notes
I took a half-hearted stab at writing an integration test to exercise the filtering login within project controller `@CrudAuth` but it seemed like it would take a while to get it working so I bailed.

## Testing Instructions
* Make sure an organization is created with a project template
* Log in as a different user or create a new user
* Join the organization
* Create a project from the template
* Log back in as the org admin and go to `/o/<org_slug>/admin`
* You should see the private project created by the other user in the list
* Ensure you can view it but cannot edit the map

Closes #665

### DIFF
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -119,14 +119,22 @@ import { Errors } from "../../../../shared/types";
       };
     } else {
       // Unauthenticated access is allowed for individual projects if they are
-      // visible or published, and not archived
+      // visible or published, and not archived. Admins can also see projects created from templates
+      // for organizations that they administer.
+      const publicallyVisible = [
+        { visibility: ProjectVisibility.Published },
+        { visibility: ProjectVisibility.Visible }
+      ];
       const visibleFilter = user
         ? [
+            // User created project
             { user_id: user.id },
-            { visibility: ProjectVisibility.Published },
-            { visibility: ProjectVisibility.Visible }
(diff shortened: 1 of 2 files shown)


maurizi (approved):

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -119,14 +119,22 @@ import { Errors } from "../../../../shared/types";
       };
     } else {
       // Unauthenticated access is allowed for individual projects if they are
-      // visible or published, and not archived
+      // visible or published, and not archived. Admins can also see projects created from templates
+      // for organizations that they administer.
+      const publicallyVisible = [
+        { visibility: ProjectVisibility.Published },
+        { visibility: ProjectVisibility.Visible }
+      ];
```
maurizi, line 127: Thanks for sneaking in this refactor 🧹

PR #674, opened 2021-04-05
Display import map button on home screen when no maps have been created

### LINKED ISSUE #643: Add import button to map list empty state
We have an import button on the map list page when there are maps present. We should also add a Import Map button to the empty state of the page.

AC:
* There is an Import Map button on the empty state of the map list page that starts the process of importing a map.
* The import button is the same as when there are maps present. It is below the Create Map button and centered.

kevinearldenny:
## Overview

Displays the import button on the home screen, even when a user has not created any maps yet.

### Checklist

### Demo

image

### Notes

@jfrankl @tgilcrest  Two things I'm a little uncertain about here:

1. What color should this button be? I have it as gray / `links.secondaryButton` currently, but this wasn't explicitly noted in the issue
2. Do we want any icon here? It looks a little weird with an icon on the 'Create a map' button, but not on this button.

## Testing Instructions

- Create a new user account and login
- Expect: Import button displayed on home screen underneath 'Create a map button', which opens the import map dialog when clicked

Closes #643


jfrankl (approved):
Looks good!

PR #676, opened 2021-04-05
Whitelist isFeatured field from triggering change to Project.updatedDt

### LINKED ISSUE #668: "Updated On" is updated by setting featured/unfeatured
## Description

In the admin list for an organization's maps, there is an "Updated on" column that indicates when the map was last updated. This date/time is affected by setting the map to featured/unfeatured

## Steps To Reproduce

1. Go to the Azavea test organization map list as the org admin
2. Click feature/unfeature on any map

## Expected behavior

The Updated on value for the map should not be affected by feature/unfeature.

## Actual behavior

The Updated on value for the map is updated when the map is featured/unfeatured.

maurizi:
## Overview

We want to track the most recently updated date for users projects, but organization admins setting the `isFeatured` flag should not be considered an update.

### Checklist

## Testing Instructions

- On `develop`, toggling the Featured/Unfeatured flag will also update the `Project.updatedDt` field. On this branch it should not.
- Editing other project fields (name, districts, visibility, etc.) should continue to update the `updatedDt` field

Closes #668

### DIFF
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -217,12 +217,13 @@ export class ProjectsController implements CrudController<Project> {
           }
         : dto;
 
-    // Update updatedDt field when non-whitelisted fields have changed
-    const whitelistedFields = ["isFeatured", "districts", "updatedDt"];
-    const data = _.isEqual(
-      _.omit(dataWithDefinitions, whitelistedFields),
-      _.omit(existingProject, whitelistedFields)
-    )
+    // Only change updatedDt field when whitelisted fields have changed
+    const whitelistedFields: ReadonlyArray<keyof UpdateProjectDto> = [
+      "districtsDefinition",
+      "name"
+    ];
+    const fields = whitelistedFields.filter(field => field in dto);
+    const data = _.isEqual(_.pick(dataWithDefinitions, fields), _.pick(existingProject, fields))
       ? dataWithDefinitions
       : { ...dataWithDefinitions, updatedDt: new Date() };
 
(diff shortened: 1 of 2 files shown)


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -217,6 +217,15 @@ export class ProjectsController implements CrudController<Project> {
           }
         : dto;
 
+    // Update updatedDt field when non-whitelisted fields have changed
+    const whitelistedFields = ["isFeatured", "districts", "updatedDt"];
```
maurizi, line 221: @tgilcrest are there any other fields we want to whitelist here? `visibility` (updating the Share menu option)?

tgilcrest: Yeah, visibility should not update it. I don't know all the other fields, but I'm mainly thinking just district definition and name are the relevant ones. I don't know if this would be updated by lock/unlock also, but I'm thinking we don't care about that action either.

kevinearldenny (changes requested):
The core fix looks good here, but there is a little bit of lint to clean up:

```
index 7af8000..0bc5a48 100644
@@ -4,8 +4,7 @@ import {
   Entity,
   JoinColumn,
   ManyToOne,
-  PrimaryGeneratedColumn,
-  UpdateDateColumn
+  PrimaryGeneratedColumn
 } from "typeorm";
```

kevinearldenny:
Also while testing this, I encountered two bugs around table sorting that are now more visible as a result of this fix and wrote them up in #678

src/server/src/projects/controllers/projects.controller.ts
maurizi: Updated to explicitly only change `updatedDt` on changes to districts or name in https://github.com/PublicMapping/districtbuilder/pull/676/commits/928665ee2796e91d3b9a2a769f9cfc8db540f2e2

kevinearldenny (approved):
👍

PR #677, opened 2021-04-05
Update index of locked districts to make the last district lockable

### LINKED ISSUE #666: Last district cannot be locked
## Description
It's not possible to lock the last district on a map.

## Steps To Reproduce

1. Load any map
2. Try to lock the last district

## Expected behavior

The last district should lock.

## Actual behavior

An error message appears. "Something went wrong."

The PATCH request to lock the district returns a 400 with this error message:
```json
message: {lockedDistricts: ["Length of array does not match "number_of_districts""]}
```

## Additional context

This is probably happening because we're counting the unassigned district as a lockable district, but it's not counted in "number of districts", so the length of lockedDistricts is 1 more than the number of districts.

kevinearldenny:
## Overview

Fixes a bug identified in #666 where the last district in the list cannot be locked by updating the IDs of districts in the list. Previously, the ø / unassigned district was associated with the first value in the locked districts list, when in reality it should never be lockable. By updating the indices of the locked districts in the sidebar and map to be n-1, all districts are now able to be locked. 

Also, the ability to lock the unassigned district was removed.

### Checklist

### Demo

image

### Notes

## Testing Instructions

- Create a map with a full set of districts, and attempt to lock all of the districts individually. All of the districts should be lockable, and their lock state should be reflected in the map as well as the sidebar

Closes #666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix bug where projects are created after closing Join Organization modal triggered from an organization's template [#673](https://github.com/PublicMapping/districtbuilder/pull/673)
 
 ### Fixed
+- Allow last district to be locked [#677](https://github.com/PublicMapping/districtbuilder/pull/677)
+
 
 ## [1.3.0] - 2021-01-26
 
--- a/src/server/migrations/1617656641073-locked-districts.ts
+++ b/src/server/migrations/1617656641073-locked-districts.ts
@@ -1,0 +1,15 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+'use strict';
+
+export class lockedDistricts1617656641073 implements MigrationInterface {
+    name = 'lockedDistricts1617656641073'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE project SET locked_districts = locked_districts[2:] || false;`, undefined);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE project SET locked_districts = false || locked_districts[:cardinality(locked_districts) - 1];`, undefined);
+    }
+
+}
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add project evaluate view for contiguity [#648](https://github.com/PublicMapping/districtbuilder/pull/648)
 - Add project evaluate view for County Splits [#644](https://github.com/PublicMapping/districtbuilder/pull/644)
 - Display Import Map button on home screen when a user has not created maps yet [#674](https://github.com/PublicMapping/districtbuilder/pull/674)
+- Allow last district to be locked [#677](https://github.com/PublicMapping/districtbuilder/pull/677)
```
maurizi, line 36: I think this belongs under `### Fixed`?

maurizi (changes requested):
We should add a data migration to this PR, to adjust the existing `locked_districts` column - otherwise this change will effectively change the locked districts of every project.

The following SQL command should do it I think: ` UPDATE project SET locked_districts = locked_districts[2:] || false;`

src/server/migrations/1617656641073-locked-districts.ts
```diff
@@ -0,0 +1,15 @@
...
+export class lockedDistricts1617656641073 implements MigrationInterface {
+    name = 'lockedDistricts1617656641073'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`UPDATE project SET locked_districts = locked_districts[2:] || false;`, undefined);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        
+    }
```
kevinearldenny, line 13: @maurizi Do we need to add anything in the `down` section to potentially reverse this migration?

maurizi: I think this would work?
```suggestion
    public async down(queryRunner: QueryRunner): Promise<void> {
        await queryRunner.query(`UPDATE project SET locked_districts = false || locked_districts[:cardinality(locked_districts) - 1];`, undefined);
    }
```

maurizi (approved):
👍

maurizi (approved):

PR #679, opened 2021-04-06
Fix evaluate mode toggle to no longer break project editing

### LINKED ISSUE #660: Toggling Evaluate mode breaks project editing for all but top-level (e.g. county) geounits
GIF shows this w/ Dane County & blocks, but this also happened w/ PA & block groups / blocks.

output

kevinearldenny:
## Overview

- Fix bug identified in #660 around toggling evaluate mode breaking project editing for lower-level geounits
- Hides editing tools when in evaluate mode
- Hides legend when not in evaluate mode

### Checklist

## Testing Instructions

- Open a project, and select 'Block group' editing level
- Toggle editing mode and select a view (e.g., compactness)
- Expect: Cursor becomes the map pan tool when in evaluate mode
- Turn evaluate mode off
- Expect: You are able to edit the project at the Block group level after opening evaluate mode

Closes #660

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -297,6 +297,9 @@ const DistrictsMap = ({
           map.setLayoutProperty(DISTRICTS_EVALUATE_LAYER_ID, "visibility", "visible");
           map.setLayoutProperty(DISTRICTS_CONTIGUITY_CHLOROPLETH_LAYER_ID, "visibility", "visible");
         }
+        DefaultSelectionTool.disable(map);
+        RectangleSelectionTool.disable(map);
+        PaintBrushSelectionTool.disable(map);
       } else {
         // Reset map state to default
         map.setLayoutProperty(DISTRICTS_LAYER_ID, "visibility", "visible");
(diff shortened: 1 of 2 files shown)


src/client/components/map/Map.tsx
```diff
@@ -305,15 +308,24 @@ const DistrictsMap = ({
...
+            map.setLayoutProperty(
+              levelToSelectionLayerId(geoLevel.id),
+              "visibility",
+              idx >= invertedGeoLevelIndex ? "visible" : "none"
+            );
           }
+          map.setLayoutProperty(
+            levelToLineLayerId(geoLevel.id),
+            "visibility",
+            idx >= invertedGeoLevelIndex ? "visible" : "none"
```
pcaisse, line 323: Since this is duplicated I'd break it out as something like:
```ts
const geoLevelVisibility = idx >= invertedGeoLevelIndex ? "visible" : "none"
```
and use it in both places.

src/client/components/map/Map.tsx
```diff
@@ -305,15 +308,24 @@ const DistrictsMap = ({
         map.setLayoutProperty(TOPMOST_GEOLEVEL_EVALUATE_SPLIT_ID, "visibility", "none");
         map.setLayoutProperty(TOPMOST_GEOLEVEL_EVALUATE_FILL_SPLIT_ID, "visibility", "none");
         map.setLayoutProperty(DISTRICTS_CONTIGUITY_CHLOROPLETH_LAYER_ID, "visibility", "none");
+        const invertedGeoLevelIndex = staticMetadata.geoLevelHierarchy.length - geoLevelIndex - 1;
         staticMetadata.geoLevelHierarchy.forEach((geoLevel, idx) => {
-          if (idx === staticMetadata.geoLevelHierarchy.length - 1) {
-            map.setLayoutProperty(levelToLineLayerId(geoLevel.id), "visibility", "visible");
-            map.setLayoutProperty(levelToSelectionLayerId(geoLevel.id), "visibility", "visible");
+          if (!evaluateMode) {
```
pcaisse, line 313: I'm not sure I fully understand this conditional. Why do we want to avoid toggling geolevel visibility for only selection layers when not in evaluate mode?

pcaisse (approved):
Tested well. Left a couple of comments.

src/client/components/map/Map.tsx
kevinearldenny: I'm not entirely sure, tbh. Seems to work fine after removing

PR #681, opened 2021-04-07
Display export button on shared projects

### LINKED ISSUE #661: Add export to shared map view
When viewing a shared map, we allow a user to copy the map to their account. We should also allow them to export the map to any of our existing formats. This should hopefully just involve adding the same export actions to the shared view as are available in the map editing view.

AC:
* On the shared (non-creator) view of a map (projects/[proj ID]/), there is an export button that allows a user to export to SHP, geojson or CSV.

kevinearldenny:
## Overview

Displays the export button on shared projects, like we had previously been including on an individual's created projects

### Checklist

### Demo

image

## Testing Instructions

- `./scripts/server`
- Create a project from an org template with a non-admin user and publish it
- Login as the org admin user, and go to the org admin page by clicking View All on the Org Page
- Select a project from another user
- Expect: Export menu is displayed in header and is fully functional

Closes #661


pcaisse (approved):
LGTM!

PR #685, opened 2021-04-09
Add Evaluate Mode view for Equal Population

### LINKED ISSUE #606: Evaluate mode equal population
image

Notes:
If https://github.com/PublicMapping/districtbuilder/issues/600 is not completed you can use a default value for population threshold of `5%`

AC:
 - Update evaluate mode summary sidebar to display equal population value and ✅ or ❌
 - Show population layer, map legend, and sidebar legend when switching to equal population detail

kevinearldenny:
## Overview

Adds Evaluate Mode view for Equal Population

- Add new screen
- Compute population deviation on backend and add to GeoJSON so we can style the map based on it
- Add diverging scale on map to display the deviation from the the expected population for a district

### Checklist

### Demo

Scenario A - incomplete project

image
image

Holes in the map are no longer displayed 👍 
Each choropleth step is displayed properly 👍 

Scenario B - 'completed' project

image
image

### Notes

~1) All areas which are not assigned to a district should ideally be no fill... any idea how we can do this?~

~2) It seems like the choropleth stops are getting a little messed up in some instances - but _only_ in the map, not the sidebar view. Namely, districts which are within the +/- 5% bucket appear to be getting knocked down a step?~

## Testing Instructions

- `./scripts/server`
- Open a project, click Evaluate -> Equal Population

Closes #606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Add project evaluate view for Equal Population [#685](https://github.com/PublicMapping/districtbuilder/pull/685)
+
 
 ### Changed
 
@@ -48,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add project evaluate view for County Splits [#644](https://github.com/PublicMapping/districtbuilder/pull/644)
 - Display Import Map button on home screen when a user has not created maps yet [#674](https://github.com/PublicMapping/districtbuilder/pull/674)
 - Display export button in project header for published projects from other users [#681](https://github.com/PublicMapping/districtbuilder/pull/681)
+
 
 ### Changed
 
--- a/src/client/components/ProjectSidebar.tsx
+++ b/src/client/components/ProjectSidebar.tsx
@@ -15,8 +15,8 @@ import { DistrictGeoJSON, DistrictsGeoJSON, SavingState } from "../types";
 import {
   areAnyGeoUnitsSelected,
   assertNever,
-  getTargetPopulation,
-  mergeGeoUnits
+  mergeGeoUnits,
+  getTargetPopulation
 } from "../functions";
 import {
   getSavedDistrictSelectedDemographics,
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -21,7 +21,7 @@ import {
   EvaluateMetric
 } from "../../../shared/entities";
 import { DistrictsGeoJSON } from "../../types";
-import { areAnyGeoUnitsSelected, getSelectedGeoLevel } from "../../functions";
+import { areAnyGeoUnitsSelected, getSelectedGeoLevel, getTargetPopulation } from "../../functions";
 import {
   GEOLEVELS_SOURCE_ID,
   DISTRICTS_SOURCE_ID,
@@ -45,7 +45,8 @@ import {
   DISTRICTS_CONTIGUITY_CHLOROPLETH_LAYER_ID,
   CONTIGUITY_FILL_COLOR,
   COUNTY_SPLIT_FILL_COLOR,
-  EVALUATE_GRAY_FILL_COLOR
+  EVALUATE_GRAY_FILL_COLOR,
+  DISTRICTS_EQUAL_POPULATION_CHOROPLETH_LAYER_ID
 } from "./index";
 import DefaultSelectionTool from "./DefaultSelectionTool";
 import FindMenu from "./FindMenu";
@@ -86,7 +87,7 @@ const style: ThemeUIStyleObject = {
   },
   legendItem: {
     display: "inline-block",
-    minWidth: "150px",
+    minWidth: "120px",
     maxWidth: "170px",
     mr: "20px"
   },
@@ -94,16 +95,16 @@ const style: ThemeUIStyleObject = {
     display: "inline-block",
     width: "120px",
     fontWeight: "600",
-    mr: "50px"
+    mr: "35px"
   },
   legendBox: {
     position: "absolute",
     bottom: "20px",
     left: "100px",
-    right: "200px",
+    right: "40px",
     height: "60px",
     minWidth: "600px",
-    maxWidth: "1100px",
+    maxWidth: "1300px",
     fontSize: "14pt",
     display: "inline-block",
     outline: "1px solid gray",
@@ -194,7 +195,8 @@ const DistrictsMap = ({
         minZoom,
         maxZoom,
         map,
-        geojson
+        geojson,
+        evaluateMetric && "avgPopulation" in evaluateMetric ? evaluateMetric : undefined
       );
 
       setMap(map);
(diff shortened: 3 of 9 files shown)


kevinearldenny:
Worked with @jfrankl to update the color ramp:

image

src/server/src/districts/entities/geo-unit-topology.entity.ts
```diff
@@ -246,17 +246,34 @@ export class GeoUnitTopology {
       type: "GeometryCollection",
       geometries: merged
     });
+    const avgPopulation: number = Math.floor(
+      featureCollection.features.reduce(function(accumulator, feature) {
+        // @ts-ignore eslint-ignore-next-line
+        return accumulator + Number(feature.properties.population);
+      }, 0) /
+        (featureCollection.features.length - 1)
+    );
```
maurizi, line 255: Why are we adding these properties to the backend districts geojson when we _also_ calculate this on the frontend?

I had assumed we would continue to calculate these variables on the frontend (bonus to this approach - it avoids the question of how to migrate existing cached `districts` geojson for now if we don't make changes to it).

If there is some reason we need to calculate these on the backend now though, we should migrate our existing frontend code to rely on that and stop redundantly recalculating this for the normal sidebar.

https://github.com/PublicMapping/districtbuilder/blob/e39dc0fff9dee832539684a26ea5422325064d84/src/client/components/ProjectSidebar.tsx#L458

https://github.com/PublicMapping/districtbuilder/blob/e39dc0fff9dee832539684a26ea5422325064d84/src/client/components/ProjectSidebar.tsx#L474-L479

src/server/src/districts/entities/geo-unit-topology.entity.ts
```diff
@@ -246,17 +246,34 @@ export class GeoUnitTopology {
...
-            demographics: feature.properties,
+            demographics: {
+              ...feature.properties,
+              popDeviation: populationDeviation
+            },
+            percentDeviation:
+              // @ts-ignore
+              feature.properties.population > 0 && feature.id !== 0
+                ? populationDeviation / avgPopulation
+                : undefined,
```
maurizi, line 276: This may be rendered moot by the above comment, but out of curiousity why put `popDeviation` under `properties.demographics` but `percentDeviation` under `properties`?

maurizi (changes requested):

maurizi:
@kevinearldenny my color ramp doesn't look like your screenshot - did you forget to push your changes?
image

src/server/src/districts/entities/geo-unit-topology.entity.ts
kevinearldenny: We need to access `percentDeviation` in the mapbox style expression, and from what I could tell it seems like we are not able to access nested elements in the GeoJSON for styling purposes.

src/client/components/ProjectSidebar.tsx
```diff
@@ -476,7 +469,7 @@ const SidebarRows = ({
         const deviation =
           districtId === 0
             ? feature.properties.demographics.population
-            : feature.properties.demographics.population - averagePopulation;
+            : feature.properties.demographics.popDeviation;
```
kevinearldenny, line 472: Now using the computed `popDeviation` field from the geojson

src/server/src/districts/entities/geo-unit-topology.entity.ts
maurizi: >[...] the question of how to migrate existing cached districts geojson for now if we don't make changes to it

@kevinearldenny I saw you went with the approach of consolidating this to be calculated on the server. Reminder that we also need to handle the above scenario in that case (i.e. what happens if I have a map with `districts` cached from before this code change was introduced, and I enter evaluate mode w/o making any changes? what triggers us to recalculate such that we get `popDeviation` added to our cached data?)

src/client/components/map/Map.tsx
```diff
@@ -231,11 +234,23 @@ const DistrictsMap = ({
...
+      // @ts-ignore
+      // eslint-disable-next-line
+      feature.properties.percentDeviation =
+        // @ts-ignore
+        feature.properties.demographics.population > 0 && feature.id !== 0
+          ? populationDeviation / avgPopulation
+          : undefined;
+      // @ts-ignore
+      // eslint-disable-next-line
+      feature.properties.populationDeviation = populationDeviation;
```
maurizi, line 248: You could clean up this a bit by disabling eslint for this section and explicitly telling TypeScript we're going to treat `feature.properties` as `any`:
```diff
diff --git a/src/client/components/map/Map.tsx b/src/client/components/map/Map.tsx
index b90ed84..e94bb5c 100644
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -222,30 +222,23 @@ const DistrictsMap = ({
     const avgPopulation = getTargetPopulation(geojson, project);
     // Add a color property to the geojson, so it can be used for styling
     geojson.features.forEach((feature, id) => {
-      // @ts-ignore
-      // eslint-disable-next-line
-      feature.properties.outlineColor =
+      /* eslint-disable */
+      const properties: any = feature.properties;
+      properties.outlineColor =
         (findTool === FindTool.Unassigned && id === 0) ||
         (findTool === FindTool.NonContiguous &&
           id !== 0 &&
           feature.geometry.coordinates.length >= 2)
           ? "#F25DFE"
           : "transparent";
-      // @ts-ignore
-      // eslint-disable-next-line
-      feature.properties.color = getDistrictColor(id);
-      // @ts-ignore
+      properties.color = getDistrictColor(id);
       const populationDeviation = feature.properties.demographics.population - avgPopulation;
-      // @ts-ignore
-      // eslint-disable-next-line
-      feature.properties.percentDeviation =
-        // @ts-ignore
+      properties.percentDeviation =
         feature.properties.demographics.population > 0 && feature.id !== 0
           ? populationDeviation / avgPopulation
           : undefined;
-      // @ts-ignore
-      // eslint-disable-next-line
-      feature.properties.populationDeviation = populationDeviation;
+      properties.populationDeviation = populationDeviation;
+      /* eslint-enable */
     });
 
     const districtsSource = map && map.getSource(DISTRICTS_SOURCE_ID);
```

maurizi (approved):
👍 Working well

src/client/components/map/Map.tsx
maurizi: Actually on second thought in light of the changes in https://github.com/PublicMapping/districtbuilder/pull/688 you might want to just hold off on this

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add project evaluate view for County Splits [#644](https://github.com/PublicMapping/districtbuilder/pull/644)
 - Display Import Map button on home screen when a user has not created maps yet [#674](https://github.com/PublicMapping/districtbuilder/pull/674)
 - Display export button in project header for published projects from other users [#681](https://github.com/PublicMapping/districtbuilder/pull/681)
+- Add project evaluate view for Equal Population [#685](https://github.com/PublicMapping/districtbuilder/pull/685)
```
maurizi, line 51: This should go under the `Unreleased` section

PR #686, opened 2021-04-09
Set map as unfeatured if private

### LINKED ISSUE #684: Reset map to unfeatured if user marks it private
It's possible that once a map is selected as featured a user can mark the map private again. The map however remains on the Organization page in the Featured maps section. So we should make sure the map no longer appears there if it's changed to private.

AC:
* If a user changes a map back to private after it has been featured on an Organization, the map is set to Unfeatured on any Organizations it is a linked to.

kevinearldenny:
## Overview

If a map is currently featured and then set to private, `isFeatured` is set to false and the map is no longer displayed on the Organization Screen

### Checklist

## Testing Instructions

- Login as an org admin and create a project
- Set a project as 'Published' from the Project screen
- Set that project to be featured from the Org admin page
- Go back to the project, and set it's visibility to 'Private'
- Go back to Org screen - expect that the project is no longer displayed there

Closes #684


maurizi (approved):
👍

PR #687, opened 2021-04-13
Toggle map pan tool with spacebar

### LINKED ISSUE #429: Use key to pan with rectangle and paintbrush tool
To make it easier to pan the map when using the rectangle tool, we should set up a key press that toggles a panning control.

AC:
* When the rectangle or paintbrush tool is active, when the user presses the space key, the controls change the cursor to map panning.
* When the space key is pressed, the cursor changes to `grab` (see comment below)

A note on the complexity of the interaction of clicking and pressing space:
* Do nothing if the spacebar is pressed while the cursor is held. That means, wait until the user lets go of the cursor and then activate the grab hand. Until they let go of mouse button, the rectangle (and paintbrush for that matter) continue to function as they did before the spacebar is pressed.

Check out the comments below for more detail.

kevinearldenny:
## Overview

Toggles to the map pan ('grab') tool from the rectangle or paintbrush select tool when the spacebar is held.

- Prevent toggle when the user is actively selecting new elements with the rectangle tool

### Checklist

## Testing Instructions

- `./scripts/server`
- Start selecting elements with the rectangle selection or paintbrush
- Hit spacebar - expect cursor to become map pan tool
- Hit spacebar again - expect cursor to return to rectangle select
- Continue selecting more elements, and hit the spacebar as you are selecting more elements. Expect - nothing happens

Closes #429

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-
 ### Added
 
 - Add project evaluate view for Equal Population [#685](https://github.com/PublicMapping/districtbuilder/pull/685)
-
+- Toggle map pan tool when holding down spacebar in rectangle / paintbrush select mode [#687](https://github.com/PublicMapping/districtbuilder/pull/687)
 
 ### Changed
 
-
 ### Fixed
-
 
 ## [1.4.0] - 2021-04-12
 
@@ -66,11 +63,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix bug where projects are created after closing Join Organization modal triggered from an organization's template [#673](https://github.com/PublicMapping/districtbuilder/pull/673)
 
 ### Fixed
+
 - Allow last district to be locked [#677](https://github.com/PublicMapping/districtbuilder/pull/677)
 - Allow editing at lower geolevels after toggling evaluate mode [#679](https://github.com/PublicMapping/districtbuilder/pull/679)
 - Fix off screen geounits deselected when using rectangle tool [#680](https://github.com/PublicMapping/districtbuilder/pull/680)
 - Set project as unfeatured if private [#686](https://github.com/PublicMapping/districtbuilder/pull/686)
-
 
 ## [1.3.0] - 2021-01-26
 
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { useEffect, useRef } from "react";
+import { useEffect, useRef, useState } from "react";
 import { Box, jsx, ThemeUIStyleObject } from "theme-ui";
 
 import MapboxGL from "mapbox-gl";
@@ -140,6 +140,13 @@ const DistrictsMap = ({
   setMap
 }: Props) => {
   const mapRef = useRef<HTMLDivElement>(null);
+  const [selectionInProgress, setSelectionInProgress] = useState<boolean>();
+  const [panToggled, setTogglePan] = useState<boolean>(false);
+
+  const isPanning =
+    panToggled &&
+    !selectionInProgress &&
+    (selectionTool === SelectionTool.Rectangle || selectionTool === SelectionTool.PaintBrush);
 
   // Conversion from readonly -> mutable to match Mapbox interface
   const [b0, b1, b2, b3] = staticMetadata.bbox;
@@ -216,6 +223,28 @@ const DistrictsMap = ({
     // Everything in this effect should only happen on component load
     // eslint-disable-next-line
   }, [mapRef]);
+
+  function downHandler({ key }: KeyboardEvent) {
+    if (key === "Spacebar" || key === " ") {
+      setTogglePan(true);
+    }
+  }
+
+  const upHandler = ({ key }: KeyboardEvent) => {
+    if (key === "Spacebar" || key === " ") {
+      setTogglePan(false);
+    }
+  };
+  // Add event listeners
+  useEffect(() => {
+    window.addEventListener("keydown", downHandler);
+    window.addEventListener("keyup", upHandler);
+    // Remove event listeners on cleanup
+    return () => {
+      window.removeEventListener("keydown", downHandler);
+      window.removeEventListener("keyup", upHandler);
+    };
+  }, []);
 
   // Update districts source when geojson is fetched or find type is changed
   useEffect(() => {
@@ -523,15 +552,16 @@ const DistrictsMap = ({
   ]);
 
   useEffect(() => {
+    // Handle enable and disable of selection tools in the map
     /* eslint-disable */
-    if (map) {
+    if (map && !isReadOnly) {
       // Disable any existing selection tools
       DefaultSelectionTool.disable(map);
       RectangleSelectionTool.disable(map);
       PaintBrushSelectionTool.disable(map);
-      if (!evaluateMode) {
+      if (!evaluateMode && !isPanning) {
         // Enable appropriate tool
-        if (!isReadOnly && selectionTool === SelectionTool.Default) {
+        if (selectionTool === SelectionTool.Default) {
           DefaultSelectionTool.enable(
             map,
             selectedGeolevel.id,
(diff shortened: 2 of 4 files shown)


tgilcrest:
Based on the description, it sounds like hitting spacebar toggles this on and off. I think it should be that holding spacebar activates the pan tool, so it's only active while spacebar is pressed, but goes back to the drawing tool when released. The issue description is ambiguous, sorry about that.

@jfrankl can you confirm which is the intended action?

jfrankl:
Thank you for asking this clarifying question. Holding the spacebar should allow the user to pan. Releasing the spacebar should allow them to resume drawing.

kevinearldenny:
Okay, the pan toggle should now only occur when the user is holding down the spacebar. Ready for review

src/client/components/map/Map.tsx
```diff
@@ -139,6 +139,9 @@ const DistrictsMap = ({
   setMap
 }: Props) => {
   const mapRef = useRef<HTMLDivElement>(null);
+  const [currentTool, setCurrentTool] = useState<SelectionTool | undefined>();
```
pcaisse, line 142: I'm not sure I understand why `currentTool` and `setCurrentTool` are needed at the component level since we have `selectionTool` in Redux. Can we avoid adding to the component state here?

src/client/components/map/Map.tsx
```diff
@@ -139,6 +139,9 @@ const DistrictsMap = ({
...
+  const [rectangleActive, setRectangleActive] = useState<boolean>();
```
pcaisse, line 143: It seems like the paintbrush tool suffers from the same panning-when-in-use-messes-up-cursor issue as the rectangle selection tool:
db_paint_pan

As such, it might make sense to generalize this component state and update function to something like:
 ```suggestion
   const [selectionInProgress, setSelectionInProgress] = useState<boolean>();
 ```

And then have the paintbrush tool also call `setActive` similarly to the rectangle selection tool.

src/client/components/map/RectangleSelectionTool.ts
```diff
@@ -47,7 +47,9 @@ const RectangleSelectionTool: ISelectionTool = {
     staticMetadata: IStaticMetadata,
     districtsDefinition: DistrictsDefinition,
     lockedDistricts: LockedDistricts,
-    staticGeoLevels: UintArrays
+    staticGeoLevels: UintArrays,
+    // eslint-disable-next-line
+    setActive: (isActive: boolean) => {}
```
pcaisse, line 52: We want to [return void](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#return-types-of-callbacks) here:
```suggestion
    setActive: (isActive: boolean) => void
```

pcaisse (changes requested):

src/client/components/map/Map.tsx
```diff
@@ -215,6 +218,82 @@ const DistrictsMap = ({
...
+  // @ts-ignore
+  // eslint-disable-next-line
+  function downHandler({ key }) {
+    if (key === "Spacebar" || key === " ") {
+      setTogglePan(true);
+    }
+  }
+  // @ts-ignore
+  // eslint-disable-next-line
+  const upHandler = ({ key }) => {
```
pcaisse, line 230: See previous comment

src/client/components/map/Map.tsx
```diff
@@ -500,14 +526,14 @@ const DistrictsMap = ({
 
   useEffect(() => {
     /* eslint-disable */
-    if (map) {
+    if (map && !isReadOnly) {
       // Disable any existing selection tools
       DefaultSelectionTool.disable(map);
       RectangleSelectionTool.disable(map);
       PaintBrushSelectionTool.disable(map);
```
kevinearldenny, line 533: @pcaisse I figured it out! We needed to first _disable everything_ before re-enabling the correct tool

src/client/components/map/Map.tsx
```diff
@@ -238,6 +263,7 @@ const DistrictsMap = ({
   }, [map, geojson, findTool]);
 
   const removeSelectedFeatures = (map: MapboxGL.Map, staticMetadata: IStaticMetadata) => {
+    // setSelectionInProgress(false);
```
pcaisse, line 266: Should this be commented out?

src/client/components/map/Map.tsx
```diff
@@ -550,6 +578,61 @@ const DistrictsMap = ({
...
+    if (map && !isReadOnly) {
+      if (
+        panToggled &&
+        !selectionInProgress &&
+        (selectionTool === SelectionTool.Rectangle || selectionTool === SelectionTool.PaintBrush)
+      ) {
+        DefaultSelectionTool.disable(map);
+        RectangleSelectionTool.disable(map);
+        PaintBrushSelectionTool.disable(map);
+        setCurrentTool(selectionTool);
```
pcaisse, line 591: I'm still having trouble understanding what `currentTool` means exactly... Is it the currently-selected-but-not-in-use pannable tool? Maybe renaming it or adding comments would help.

src/client/components/map/Map.tsx
```diff
@@ -550,6 +578,61 @@ const DistrictsMap = ({
...
+      } else {
+        if (currentTool === SelectionTool.Rectangle) {
+          DefaultSelectionTool.disable(map);
```
pcaisse, line 594: Does the default selection tool need to be disabled here?

src/client/components/map/Map.tsx
```diff
@@ -550,6 +578,61 @@ const DistrictsMap = ({
...
+          RectangleSelectionTool.enable(
+            map,
+            selectedGeolevel.id,
+            staticMetadata,
+            project.districtsDefinition,
+            lockedDistricts,
+            staticGeoLevels,
+            setSelectionInProgress
+          );
+          setCurrentTool(undefined);
```
pcaisse, line 604: Does this need to happen for only the rectangle and paint tools and not the default selection tool? If we could avoid  disabling the default selection tool within each if block and could instead do it outside of the if blocks it would be DRY-er.

src/client/components/map/Map.tsx
pcaisse: So here we're setting `currentTool` which is a dependency for this effect. Setting `currentTool` then causes the effect function to be revaluated which then affects the `else` block below... I'm having trouble following the flow here.

I don't fully understand what's happening here, but I think in the interest of keeping our state as simple as we can, only enabling/disabling the selection tools in one place is desirable and we're already doing that in the effect function on line 527. Is that possible? The more places we enable/disable the more places we could forget to do that properly. I'm not sure if that's possible here but may be something to consider.

src/client/components/map/Map.tsx
kevinearldenny: My understanding is that we explicitly do not want this effect to run on the default selection tool. Since it's cursor already has a pan functionality associated with it, the only thing this would do is remove the ability to select - so, the spacebar toggle should have no impact when the default selection tool is enabled. 

@tgilcrest - can you confirm here?

src/client/components/map/Map.tsx
kevinearldenny: Yes, as we discussed when we were pairing on Friday - `currentTool` is preserving the state of the tool that is currently selected with the toggle buttons. If we were to try to use `selectionTool` in lines 595 and 607 instead, this effect would run every time `selectionInProgress` changes making the rectangle selection tool appear while the spacebar is held down.

src/client/components/map/Map.tsx
kevinearldenny: Nope - removed now.

src/client/components/map/Map.tsx
kevinearldenny: I renamed this to `preserveTool`, which I personally think is a little bit _less_ explanatory... but I'm at a loss for a better name here. Open to any suggestions that you have for a better name

pcaisse:
Thanks for the comments. They helped me understand what was going on better.

I think I have a suggestion that will allow us to simplify things. It seems to be functionally equivalent based on my testing but is less code. Here's the commit which you can cherry pick onto your branch if you like: https://github.com/PublicMapping/districtbuilder/commit/882605117f28ac6a2580d4caf00b68e7ebd580fd

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add project evaluate view for County Splits [#644](https://github.com/PublicMapping/districtbuilder/pull/644)
 - Display Import Map button on home screen when a user has not created maps yet [#674](https://github.com/PublicMapping/districtbuilder/pull/674)
 - Display export button in project header for published projects from other users [#681](https://github.com/PublicMapping/districtbuilder/pull/681)
+- Toggle map pan tool when holding down spacebar in rectangle / paintbrush select mode [#687](https://github.com/PublicMapping/districtbuilder/pull/687)
```
maurizi, line 38: This needs to be moved to the `Unreleased` section

maurizi, line 38: This needs to be moved to the `Unreleased` section

kevinearldenny:
> Thanks for the comments. They helped me understand what was going on better.
> 
> I think I have a suggestion that will allow us to simplify things. It seems to be functionally equivalent based on my testing but is less code. Here's the commit which you can cherry pick onto your branch if you like: [8826051](https://github.com/PublicMapping/districtbuilder/commit/882605117f28ac6a2580d4caf00b68e7ebd580fd)

I like this solution a lot, definitely much cleaner 👍  I should have released this sooner, especially after noting that the two effects were conflicting with each other...

pcaisse (approved):

PR #688, opened 2021-04-13
Highlight selected/hovered districts and zoom to district

### LINKED ISSUE #612: Highlight district when selected and on hover
When the user is looking at a district in their sidebar, they want to quickly figure out which district on the map it corresponds to. The active district (district being edited) should have a thick border to make it stand out from the rest. When the user hovers a district in the sidebar, that district should temporarily receive the same treatment. An overflow menu provides the option to "Zoom to district", further connecting the sidebar to the map.

AC:
* The selected district has a thicker border than other districts
* When a user hovers on a district in the side bar, the district hovered on also has a thicker border (in addition to the selected district)
* When hovering on a district, an overflow menu appears on the edge of the district sidebar
* Expanding the overflow menu provides the option to "Zoom to district" (same zoom as Find Unassigned)

Mockups (Ignore "Rename" and "Change color" options in overflow menu):

pcaisse:
## Overview
Highlight selected/hovered districts and zoom to district.

### Checklist

### Demo
db_district_borders_zoom

## Notes
There are two distinct pieces of functionality here:
1. District border added on select/hover
2. Zoom to district

For the border, I ended up reusing the `outlineColor` property which is already used for the "Find Unassigned" feature. By always setting the outline color via logic in TS (and to "transparent" if nothing should be shown) we no longer need the effect to conditionally hide that layer based on `findOpen`. Another change related to borders is moving the state capturing if a sidebar row is hovered up into Redux state since that is now relevant to the map as well, which allowed the component-level state to be deleted from `SidebarRow`.

For the zoom to district functionality, that reuses existing menu styles, borrow heavily from `ProjectListFlyout`, which overriding a few things to make the menu appear to the right of the sidebar. One other thing to note there is that there is currently only one item in the list but in the future there will be multiple items (as we add more items `MenuKeys` is meant to become a union type). 

@jfrankl The styles on the district options flyout look pretty close to me but maybe not to you :sweat_smile: Let me know if you think it's good enough for now.

## Testing Instructions
* Make sure selecting a district adds a border around the district of the same color as the district
* Make sure hovering a district row in the sidebar adds a border around the district of the same color as the district
* Make sure you can zoom to a district that isn't empty
* Make sure that zooming to an empty district (no geounits assigned) does nothing

Closes #612


kevinearldenny (approved):
Both highlighting and zoom to district functionalities work as documented 👍 

However, I do have some UX thoughts here:

1. I'm not sure that I like the reuse of the district fill color here for the outline of selected / hovered districts. In the below map, it's really tough to see which districts are selected:
image
My thought is that red or black would work well to _highlight_ the selected districts across the board

2. The 'Zoom to district' button seems a little clunky to be hidden under a menu. Two possibly ideas here:
a) Could we possibly add a button with a magnifying glass icon here to zoom?
b) Would it be unreasonable to just zoom to a district when it was selected? If the user is going to edit the district, this would be helpful for them... and it would also make the highlight on hover functionality a bit more distinct from the district selection behavior.

Curious to hear @jfrankl 's thoughts here

pcaisse:
> In the below map, it's really tough to see which districts are selected

Yeah, I think it might make sense to make the borders thicker. I left it the same thickness as what's used for outlining districts for "Find unassigned" but we could make the thickness, say, double that for selected/hovered districts pretty easily.

> The 'Zoom to district' button seems a little clunky to be hidden under a menu

I agree it's a bit strange presently but judging by the wireframes there will be "Rename" and "Change color" options added to that menu which I think will make it make more sense. I'll let Jeff weigh in on that.

pcaisse:
Here's how it looks with 5c2ce89:
db_thicker_borders

src/client/components/DistrictOptionsFlyout.tsx
```diff
@@ -0,0 +1,49 @@
...
+      }}
+      style={{ visibility: isDistrictHovered ? "visible" : "hidden" }}
+    >
+      <MenuButton
+        sx={{
+          ...{ variant: "icon", fontWeight: "light", color: "black", px: 1, py: 1 },
+          ...invertStyles({ invert: false })
+        }}
+      >
+        ...
```
maurizi, line 34: Using the middle dot character here will improve the styling just a little:
```suggestion
        ···
```

src/shared/entities.d.ts
```diff
@@ -96,6 +96,12 @@ export type DistrictProperties = {
   readonly contiguity: "contiguous" | "non-contiguous" | "";
   readonly compactness: number;
   readonly demographics: DemographicCounts;
+  /* eslint-disable */
+  // NOTE: These properties are set for styling purposes
+  color?: string;
+  outlineColor?: string;
+  outlineWidthScaleFactor?: number;
+  /* eslint-enable */
```
maurizi, line 104: Nice, this is a much nicer way to deal with these 🧹

jfrankl:
Nice work on this @pcaisse, the functionality looks really good.

I few small details that should get cleaned up. Let me know if you need help with these.

The "Lock" button shows a tooltip on hover, and has a circular background, while the "Ellipsis" does not have a tooltip and has a square background that is a different color. Could you please make the "Ellipsis" button more like the "Lock" in those ways? (Add tooltip, match style):
Screen Shot 2021-04-14 at 11 56 02 AM
Screen Shot 2021-04-14 at 11 57 49 AM

There appears to be a missing border in the sidebar table heading. I suspect you need to add a new empty `<th>` in the header row to accommodate the new column:
Screen Shot 2021-04-14 at 11 58 11 AM

pcaisse:
I made a few changes that make district hovering/selection feel _much_ snappier:
db_selected_hovered_districts_fast

However, I'm noticing a couple of styling/usability issues with the district options flyout. In Chrome, there's a dark border around the icon that's only there after clicking. Also there's a "gap" between the sidebar and the flyout that makes it hard/impossible to click in the flyout since it disappears before you can mouse over it which will likely drive users bananas (I noticed this happening when there's vertical scrolling in the sidebar in Firefox but seems to be more of an ever-present issue in Chrome). 
db_district_options_flyout_issue

@jfrankl Would you be able to help out with that?

jfrankl:
I opened a style PR: https://github.com/PublicMapping/districtbuilder/pull/703

@pcaisse in working on this feature, a small detail I noticed is that opening the menu switches the selected district to that district. This isn't ideal: a user might want to zoom to an area without necessarily starting to draw that area. Is this something you could address?

PR #702, opened 2021-04-21
Fix import region detection

### LINKED ISSUE #683: Filter out custom regions from import detection
When we look up a region based on the FIPS code on import, we don't do anything to distinguish between a state region and a region that falls within a state. Therefore, I can upload a Wisconsin state plan and it will detect a Dane County plan (presumably because that is region it finds that matches the state FIPS code).

In the basic import workflow, we should filter out non-state/custom regions from the detection logic.

AC:
* When a user uploads a CSV to import, we only search for matching regions that are states (we filter out custom/non-state regions from the matching).

maurizi:
## Overview

This fixes the region detection in the `ImportProjectScreen` component, and also adds an index to ensure uniqueness on `region_code` / `ccountry_code` for _visible_ regions, which should ensure that no more than 1 region config could ever be found by the new more comprehensive filter we're using on the frontend.

### Checklist

### Demo

localhost_3003_create-project

## Testing Instructions

- `scripts/migration`
- Ensure there is both a state `RegionConfig` that is not `hidden` and another `RegionConfig` with the same `region_code` that is `hidden` in your database (e.g. Wisconsin w/ `region_code` of WI and a hidden Dane County region with a `region_code` of WI, like on staging)
- Verify that you cannot insert a row into the `region_config` table with duplicate country / region code to an existing row unless it is `hidden`
- When exporting a WI map to CSV and importing it should detect the state as Wisconsin (not Dane County)
- When exporting a Dane County map to CSV and importing it should _also_ detect the state as Wisconsin (not Dane County)

Closes #683


src/client/screens/ImportProjectScreen.tsx
```diff
@@ -174,7 +174,12 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
           const importNumber = importNumberRef.current;
 
           const regionConfig =
-            regionConfigs.resource.find(config => config.regionCode === stateAbbrev) || null;
+            regionConfigs.resource.find(
+              config =>
+                config.hidden !== true &&
```
maurizi, line 179:
```suggestion
                !config.hidden &&
```

pcaisse (approved):

PR #705, opened 2021-04-22
Style cleanup for limiting drawing to county

jfrankl:
## Overview

Style cleanup for the "limit drawing to county" feature. Previously, toggling this feature would close the menu. Now, the menu persists, which allows the user to be more clear about the state. I also corrected the positioning by adding a wrapper with `position: relative` instead of using the more brittle `position: fixed`, which was dependent on the screen size and the positioning of several other components.

### Demo

Screen Shot 2021-04-22 at 4 53 38 PM


kevinearldenny (approved):
LGTM 👍  definitely a much better look and feel here, thanks for helping out here!

PR #706, opened 2021-04-23
Make password validation visible in modal

### LINKED ISSUE #579: Password validation popup not visible in registration modal
## Description
It goes to the right and is cut-off by the edge of the modal:
image

jfrankl:
## Overview

The password validation panel was being cut off by the modal's `overflow: hidden` property. I
removed `overflow: hidden` from the modal. I also reformatted the validation panel to take up less
space, which should help us avoid situations where it would expand to the edge of the browser
window.

### Checklist

### Demo

Screen Shot 2021-04-23 at 9 36 17 AM

### Notes

I experimented with rendering this panel above and below the field, but in both cases it blocked and
prevented interaction with other important elements on the page. I think this solution will work
well in a majority of cases.

## Testing Instructions

- If logged in, log out of DistrictBuilder
- Navigate to a map page
- Click on "Share" and then "Create an account"
- Click in the password input; a panel should appear explaining password requirements
- Confirm that the password requirements are a.) visible; b.) formatted like in the demo screenshot

Closes #579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Display unassigned areas as transparent fill in Evaluate mode map [#694](https://github.com/PublicMapping/districtbuilder/pull/699)
-
+- Make password validation visible in modal [#706](https://github.com/PublicMapping/districtbuilder/pull/706)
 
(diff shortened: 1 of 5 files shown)


maurizi (approved):
@jfrankl I pushed a commit to also fix this on the Organization page, but otherwise this is looking good 👍

jfrankl:
Thanks @maurizi!

Issue #652, opened 2021-03-29
Create map button doesn't activate for import CSV


tgilcrest:
When I go to import a CSV that I know is valid, the create map button doesn't activate after filling in all the fields. It only activates after a successful response the `/csv/` endpoint and then the # of districts field is edited. 

We should change the activation of the button so it activates when all the fields are filled in (and valid, e.g., no letters in the # of districts). Then, if we are still waiting on a response from the `/csv/` endpoint we should handle that separately.

AC:
* When importing a CSV, the "Create Map" button is deactivated until the csv file is uploaded and the # of districts is filled in. When these are completed, the button activates.
* When the user clicks the button, we take them to the map page if we have a successful response from `/csv/`. If are still waiting for a response, we show a spinner where the button was until we receive the response, at which point the map page should be loaded.

PR #708, opened 2021-04-23
Import plan error warnings

### LINKED ISSUE #629: Error/warning validation modals for import plan
# Wireframes
image
image

# Notes
We're very likely to need the modal with row-level details for any warnings/flags, but as per the first wireframe, it might not make sense for all or even any errors (they might be more top-level things like an invalid file format). You'll likely need to work with @jfrankl to clean up the details on this piece once the full set of conditions we're validating against is better known.

TG Edit: Like the note says, we should not create a modal for any errors, since we should just fail on the first one. Show something like this displaying the first error that we throw:
image

# AC
 - Add row-level validation for the following warnings/flags:
   - unknown FIPS codes
   - duplicate FIPS codes
   - FIPS codes from more than 1 state in the same file
    - invalid district values
 - Add any file-level validations for errors that make sense (invalid file format, file too large, empty file, etc.)
 - If there are any errors, display them and allow the user to submit a different file
 - If there are flags, display the # of them in a link that will open a modal with details for each flag.

kevinearldenny:
## Overview

Displays file errors and row flags when importing a block equivalency CSV in `ImportProjectScreen`

File errors
- File too large (based on a newly added constant, `MaxUploadFileSize`
- Invalid file format (although I don't think this is really possible)
- File with no rows

Row-level flags
- Invalid BLOCKID
- Blocks from multiple different states
- Duplicate BLOCKIDs
- Invalid district IDs (currently assuming that any numeric value is valid, and any non-numeric value is invalid)

### Checklist

### Demo

image

image

image

### Notes

This definitely needs some styling work on the modal from @jfrankl 😬 

## Testing Instructions

- Test the attached CSVs at http://localhost:3003/import-project
- Expect: Errors are displayed as indicated in screenshots

[Bad_FIPS.csv](https://github.com/PublicMapping/districtbuilder/files/6366874/Bad_FIPS.csv)
[Blanky.csv](https://github.com/PublicMapping/districtbuilder/files/6366875/Blanky.csv)

Closes #629


CHANGELOG.md
```diff
@@ -10,11 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add project evaluate view for Equal Population [#685](https://github.com/PublicMapping/districtbuilder/pull/685)
 - Toggle map pan tool when holding down spacebar in rectangle / paintbrush select mode [#687](https://github.com/PublicMapping/districtbuilder/pull/687)
-<<<<<<< HEAD
 - Highlight selected/hovered districts and zoom to district [#688](https://github.com/PublicMapping/districtbuilder/pull/688)
-=======
 - Added unique index on region config table for country / region code [#702](https://github.com/PublicMapping/districtbuilder/pull/702)
->>>>>>> Update changelog
```
maurizi, line 17: @kevinearldenny looks like you had a bad merge/rebase

kevinearldenny: Hmmm, weird... seems like this may have something to do with the GitHub issues you mentioned earlier?

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -156,37 +181,51 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
...
   const formData = createProjectResource.data;
 
   const setFile = useCallback(
     (file: Blob) => {
       async function setConfigFromFile() {
         if (file && "resource" in regionConfigs) {
+          // Check file size (must be less than MaxUploadFileSize)
+          file.size > MaxUploadFileSize && setFileError("File must be less than 25mb");
+          // @ts-ignore I don't think that there is any way this can happen, but it was mentioned in the issue
+          file.type !== "text/csv" && setFileError("File must be a .csv");
```
maurizi, line 201: By continuing on afterwards, these more specific errors are getting swallowed by the generic "File must have at least one record".

```suggestion
          if (file.size > MaxUploadFileSize) {
            setFileError("File must be less than 25mb");
            return;
          }
```

Separately, not sure why you had the `// @ts-ignore` there - I was able to remove it w/ no compilation issues.

As for how you could have a `file.type` that isn't "text/csv"
 - In Chrome at least I can switch the pre-selected `.csv` file extension to "Any files"
 
_However_ from my testing relying on `file.type` isn't fool-proof - the CSV file I downloaded from Dave's Redistricting App had an empty string for it's `file.type` when I uploaded it to our site (it seems to be a property of the file?? not sure on this). I think we could still use `file.type` if it is present, but we'd maybe want to fall back to checking if `file.name` ends with `.csv` otherwise.

maurizi (changes requested):
This is looking great so far 👍 

I left a few inline comments.

In addition, aone other error case I'd like to check for:
 - If your file has a row with something that _looks like_ a valid blockid, but isn't actually (for example I sorted my file by block id and then incremented the highest number block id by 1), we won't flag that row, but also it won't get used. That seems like 💯 something we should flag

kevinearldenny:
> In addition, one other error case I'd like to check for:
> 
> * If your file has a row with something that _looks like_ a valid blockid, but isn't actually (for example I sorted my file by block id and then incremented the highest number block id by 1), we won't flag that row, but also it won't get used. That seems like 💯 something we should flag

@maurizi good catch here 👍  I added in logic to check for invalid block IDs and return them as flags.

src/server/test/districts.controller.e2e-spec.ts
```diff
@@ -49,7 +49,7 @@ describe("DistrictsController", () => {
         .post("/api/districts/import/csv")
         .attach("file", `${__dirname}/data/de.csv`)
         .expect(200)
-        .expect([2, 1, 3]);
+        .expect({ districtsDefinition: [2, 1, 3] });
```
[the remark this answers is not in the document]
kevinearldenny: I added a new test, and it looks like it _should_ be returning as a success... but for some reason is being displayed as a failure. Any idea what's going on here?

```
DistrictsController › import plan from block equivalency CSV with invalid block id in first row › should return a flag

    expected {
      districtsDefinition: [ 2, 1, 3 ],
      rowFlags: [
        {
          rowNumber: 0,
          errorText: 'Invalid block ID',
          rowValue: [Array],
          field: 'BLOCKID'
        }
      ]
    } response body, got {
      districtsDefinition: [ 2, 1, 3 ],
      rowFlags: [
        {
          rowNumber: 0,
          errorText: 'Invalid block ID',
          rowValue: [Array],
          field: 'BLOCKID'
        }
      ]
    }
```

src/server/test/districts.controller.e2e-spec.ts
```diff
@@ -49,7 +49,27 @@ describe("DistrictsController", () => {
...
+        .post("/api/districts/import/csv")
+        .attach("file", `${__dirname}/data/bad_blockid.csv`)
+        .expect(200)
+        .expect({
+          districtsDefinition: [2, 1, 3],
+          rowFlags: [
+            {
+              rowNumber: 0,
+              errorText: "Invalid block ID",
+              rowValue: [100059900000024, 3],
```
maurizi, line 68: This should fix the tests:
```suggestion
              rowValue: ["100059900000024", "3"],
```

src/server/test/districts.controller.e2e-spec.ts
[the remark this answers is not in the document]
maurizi: Left a comment on [the relevant line](https://github.com/PublicMapping/districtbuilder/pull/708/files#r621265820)

src/server/src/districts/controllers/districts.controller.ts
```diff
@@ -27,21 +36,72 @@ export class DistrictsController {
...
-  async exportCsv(@UploadedFile() file: Express.Multer.File): Promise<DistrictsDefinition> {
+  async importCsv(@UploadedFile() file: Express.Multer.File): Promise<DistrictsImportApiResponse> {
     const parser = csvParse(file.buffer, { fromLine: 2 });
     // Seemingly the simplest way of getting all the records into an array is to iterate in a for-loop :(
     /* eslint-disable */
     let records = [];
+    // Store object of row flags by row ID
+    let rowFlags: FlaggedRows = {};
+    // Array of flagged rows to be returned
+    let flaggedRows: ImportRowFlag[] = [];
```
maurizi, line 47: You can clean things up in this method by consolidating these two variables together and switching from an object type to an array (don't worry about the performance implications of an array w/ lots of empty elements, most modern JS implementations handle sparse arrays fairly well).

This also eliminates the need for the `FlaggedRows` and `MatchingRows` types in `shared/entities` (which probably shouldn't have been there to begin with, as they were exclusively used by this method, and were also **wrong** in using `string` as the type for `rowNumber`).

```diff
diff --git a/src/server/src/districts/controllers/districts.controller.ts b/src/server/src/districts/controllers/districts.controller.ts
index 925122f..5a2aff7 100644
--- a/src/server/src/districts/controllers/districts.controller.ts
+++ b/src/server/src/districts/controllers/districts.controller.ts
@@ -15,10 +15,8 @@ import {
   DistrictsDefinition,
   GeoUnitHierarchy,
   ImportRowFlag,
-  FlaggedRows,
   DistrictsImportApiResponse,
-  DistrictImportField,
-  MatchingRows
+  DistrictImportField
 } from "../../../../shared/entities";
 import { FIPS } from "../../../../shared/constants";
 import {} from "../../../../shared/entities";
@@ -41,11 +39,9 @@ export class DistrictsController {
     // Seemingly the simplest way of getting all the records into an array is to iterate in a for-loop :(
     /* eslint-disable */
     let records = [];
-    // Store object of row flags by row ID
-    let rowFlags: FlaggedRows = {};
     // Array of flagged rows to be returned
     let flaggedRows: ImportRowFlag[] = [];
-    let matchingRows: MatchingRows = {};
+    let matchingRows: boolean[] = [];
 
     function setFlag(
       row: string[],
@@ -54,9 +50,7 @@ export class DistrictsController {
       errorText: string
     ) {
       const flag = { rowNumber: rowNumber, errorText: errorText, rowValue: row, field: field };
-      flaggedRows.push(flag);
-      // @ts-ignore
-      rowFlags[rowNumber] = flag;
+      flaggedRows[rowNumber] = flag;
     }
 
     for await (const record of parser) {
@@ -84,21 +78,21 @@ export class DistrictsController {
       }
 
       // Check to see if rows match state fips
-      if (rowFips !== stateFips && !rowFlags[i]) {
+      if (rowFips !== stateFips && !flaggedRows[i]) {
         setFlag(record, i, "BLOCKID", "All geounits in an import must be within the same state");
       }
 
       // Check for duplicate block IDs
-      if (blockIdCounts[blockId] > 1 && !rowFlags[i]) {
+      if (blockIdCounts[blockId] > 1 && !flaggedRows[i]) {
         setFlag(record, i, "BLOCKID", "Duplicate BLOCKID included in import");
       }
 
-      if (isNaN(record[1]) && !rowFlags[i]) {
+      if (isNaN(record[1]) && !flaggedRows[i]) {
         setFlag(record, i, "DISTRICT", "Invalid district ID, must be numeric");
       }
     });
 
-    const unflaggedRows = records.filter((record, i) => !rowFlags[i]);
+    const unflaggedRows = records.filter((record, i) => !flaggedRows[i]);
 
     const blockToDistricts = Object.fromEntries(unflaggedRows);
 
@@ -140,21 +134,16 @@ export class DistrictsController {
     const districtsDefinition = mapToDefinition(geoCollection.getGeoUnitHierarchy());
     // Find unmatched records
     records.forEach((record, i) => {
-      if (!matchingRows[record[0]] && !rowFlags[i]) {
+      if (!matchingRows[record[0]] && !flaggedRows[i]) {
         setFlag(record, i, "BLOCKID", "Invalid block ID");
       }
     });
 
+    const rowFlags = flaggedRows.filter(r => !!r);
+
     return {
       districtsDefinition: districtsDefinition,
-      // Sort row flags by row number ascending
-      rowFlags:
-        flaggedRows.length > 0
-          ? // eslint-disable-next-line
-            flaggedRows.sort((a, b) => {
-              return a.rowNumber - b.rowNumber;
-            })
-          : undefined
+      rowFlags: rowFlags.length > 0 ? rowFlags : undefined
     };
   }
 }
diff --git a/src/shared/entities.d.ts b/src/shared/entities.d.ts
index 120789f..90605be 100644
--- a/src/shared/entities.d.ts
+++ b/src/shared/entities.d.ts
@@ -287,15 +287,6 @@ export interface ImportRowFlag {
   readonly rowValue: readonly string[];
 }
 
-export interface FlaggedRows {
-  readonly [rowNumber: string]: ImportRowFlag;
-}
-
-export interface MatchingRows {
-  // eslint-disable-next-line
-  [rowNumber: string]: boolean;
-}
-
 export interface DistrictsImportApiResponse {
   readonly districtsDefinition?: DistrictsDefinition;
   readonly rowFlags?: readonly ImportRowFlag[];
```

src/server/src/districts/controllers/districts.controller.ts
```diff
@@ -62,7 +122,13 @@ export class DistrictsController {
           // TODO (GH#629) error handling could be more comprehensive here
```
maurizi, line 122: error handling now more comprehensive ✔️  
```suggestion
```

maurizi (changes requested):
Some fairly simple cleanups left here, but functionally this is looking good 👍

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -156,37 +181,61 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
...
-          // eslint-disable-next-line
-          importNumberRef.current = importNumberRef.current + 1;
-          const importNumber = importNumberRef.current;
+          // Check file size (must be less than MaxUploadFileSize)
+          if (file.size > MaxUploadFileSize) {
+            setFileError("File must be less than 25mb");
+            return;
+          }
+
+          // @ts-ignore
```
maurizi, line 204: You can remove this `// @ts-ignore` by changing the type of the function from `Blob` to `File` (a subclass of `Blob`)

maurizi (changes requested):
Just realized there is one more error we need to check for - we need to validate the specified number of districts is not less than the max district # in the supplied file (it's ok to be larger).

We should probably handle this error visually in a manner similar to how we do other validations on the number of districts field:
image

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -334,11 +414,26 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
...
+                                  numberOfDistricts < maxDistrictId
+                                    ? {
+                                        error: "Invalid number of districts",
+                                        message: {
+                                          numberOfDistricts: [
+                                            `Number of districts must be at least the maximum district ID in the CSV, ${maxDistrictId} `
+                                          ]
+                                        }
+                                      }
+                                    : undefined
```
maurizi, line 436: Having this in the `onChange` event caused it to be susceptible to a race condition similar to https://github.com/PublicMapping/districtbuilder/issues/652 when you change the number of districts before the API call is complete 

Moving this to a `useEffect` that takes place when _either_ `numberOfDistricts` or `maxDistrictId` changes fixes the issue:
```diff
diff --git a/src/client/screens/ImportProjectScreen.tsx b/src/client/screens/ImportProjectScreen.tsx
index 0354153..92bc976 100644
--- a/src/client/screens/ImportProjectScreen.tsx
+++ b/src/client/screens/ImportProjectScreen.tsx
@@ -253,6 +253,27 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
     [regionConfigs, importNumberRef]
   );
 
+  useEffect(() => {
+    // Set error if number of districts less than max district ID
+    if (
+      formData.numberOfDistricts !== null &&
+      maxDistrictId !== undefined &&
+      formData.numberOfDistricts < maxDistrictId
+    ) {
+      setCreateProjectResource({
+        data: formData,
+        errors: {
+          error: "Invalid number of districts",
+          message: {
+            numberOfDistricts: [
+              `Number of districts must be at least the maximum district ID in the CSV, ${maxDistrictId} `
+            ]
+          }
+        }
+      });
+    }
+  }, [formData, maxDistrictId]);
+
   function resetForm() {
     // eslint-disable-next-line
     importNumberRef.current += 1;
@@ -420,20 +441,7 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
                                 data: {
                                   ...formData,
                                   numberOfDistricts
-                                },
-                                errors:
-                                  maxDistrictId &&
-                                  numberOfDistricts !== null &&
-                                  numberOfDistricts < maxDistrictId
-                                    ? {
-                                        error: "Invalid number of districts",
-                                        message: {
-                                          numberOfDistricts: [
-                                            `Number of districts must be at least the maximum district ID in the CSV, ${maxDistrictId} `
-                                          ]
-                                        }
-                                      }
-                                    : undefined
+                                }
                               });
                             }
                           }}
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -195,6 +253,22 @@ const ImportProjectScreen = ({ regionConfigs }: StateProps) => {
...
+      // eslint-disable-next-line
+      fileInputRef.current.value = "";
+    }
+    setImportResource({
+      data: null
+    });
+    setImportFlagsModal(false);
+    setRowFlags(undefined);
+    setFileError(undefined);
+    setCreateProjectResource({ data: blankForm });
```
maurizi, line 269:
```suggestion
    setCreateProjectResource({ data: blankForm });
    setMaxDistrictId(undefined);
```

maurizi (changes requested):
Looking pretty good - just a few small cleanups to the most recent changes

maurizi (approved):
Looks good 👍

PR #709, opened 2021-04-23
Style improvements for organization page and dropdown

jfrankl:
## Overview

Removes logos from organizations dropdown; I don't think these are useful at such a small size, and
if we allow organizations to have non-square dimension logos, those wouldn't fit in this space

Adds a max-width to the location, url, and # builders indicators on the organization page, to
prevent this content from wrapping to the next line.

Makes icons for location, url, and # builders more visually consistent. I also replaced the "tools"
icon for Builders with a "user" icon, which I think will make the meaning more clear.

### Checklist

### Demo

Screen Shot 2021-04-23 at 12 50 46 PM

Screen Shot 2021-04-23 at 12 56 28 PM

## Testing Instructions

1.)
- View an organization page with a long URL, or manipulate the url user developer tools to make it longer
- You should see ellipses after 220px

2.) 
- Join an organization
- Reload page
- You should see an "Organizations" dropdown
- Open the dropdown, and you should see your organizations with no logos


maurizi (approved):
👍

jfrankl:
Thanks for your review @maurizi.

PR #717, opened 2021-04-27
Move leave organization click handler to correct location

### LINKED ISSUE #716: Clicking around Join/leave button is same as clicking join/leave button
### Steps to reproduce
1. Go to an org page
2. Click near the join/leave button, but not on it

### Expected behavior
No response

### Observed behavior
You join/leave the organization

### Notes
Mike says the is because the `onClick` is in the wrong place - the element containing the "Leave" button, rather than the button itself, so clicking the empty space around it is the same as the button

maurizi:
## Overview

Moves click handler of Leave Organization to actually be on the `Button` element

### Checklist

### Demo

### Notes

## Testing Instructions

- How to test this PR
- Prefer bulleted description

Closes #716


maurizi:
Trivial, merging when CI passes

PR #721, opened 2021-04-28
Fix Organization bugs

### LINKED ISSUE #714: Deactivating Org Template Prevents Joining/Leaving Org
In the Dane County org I made one of their templates `is_active=false` to hide it from the org page. However, after I did this, it prevents everyone from joining or leaving the Org and throws a 500 error. We should be able to deactivate templates without causing this.

### Steps to Reproduce
1. Make org template Dane County - Blockgroups `is_active=false`
2. Try to join or leave the Dane County Org
3. See error
4. Make the template `is_active=true`
5. Try to join or leave the Dane County Org

### Expected behavior
It should be possible to deactivate an org template without breaking join/leave.

### LINKED ISSUE #715: Make is_active=false only hide templates from public org page
For org templates, we have a flag, `is_active` that we can toggle to hide or show them. However, when the template is set to `is_active=false`, not only is the template hidden from the org page, but it hides any maps in the admin list view that were made with that template.

We still want to show these maps, so we should make `is_active` only affect a template's visibility on the org page, and not hide every map in the list that was made from it.

maurizi:
## Overview

Fixes two bugs 
- you couldn't leave an organization with an inactive project template, which bizarrely had a root cause of missing some fields from the custom SQL select we were using
- shows maps for project templates which are not active

### Checklist

### Demo

Note the featured map despite no active project templates, and also we can actually leave the organization:
output

## Testing Instructions

- Set up an organization with at least one template
- Create a map for that template, and publish it
- Using `scripts/dbshell` set `is_active` to `FALSE` on the project template assoicated with your map
- It should still appear on the organization admin screen.
- If you feature your map, it should still appear on the organization home screen
- Even with an inactive project template, you should be able to leave the organization

Closes #714 
Closes #715

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minor style cleanup for organizations page and dropdown [#709](https://github.com/PublicMapping/districtbuilder/pull/709)
 - Make password validation visible in modal [#706](https://github.com/PublicMapping/districtbuilder/pull/706)
 - Sorting on organization admin table [#711](https://github.com/PublicMapping/districtbuilder/pull/711)
+- Fix joining/leaving organization with an inactive project template [#721](https://github.com/PublicMapping/districtbuilder/pull/721)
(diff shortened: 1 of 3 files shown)


kevinearldenny (approved):
LGTM 👍

PR #725, opened 2021-04-30
Log non-HTTP errors to Rollbar

### LINKED ISSUE #712: Server-side Rollbar error not reported
There is currently a bug on production where attempting to leave an organization causes the server to respond with a 500 error:

image

Despite this, there are [no reported errors in Rollbar](https://rollbar.com/Azavea/DistrictBuilder/).

pcaisse:
## Overview

Our logic was a little flawed in that we required errors to be
`HttpException`s to be logged to Rollbar. However, we also want to know
about other types of unhandled exceptions.

This is slightly tricky in terms of types since in JavaScript, bizarrely, you can
actually throw _any_ type (does not need to be an `Error` or extend
`Error`). However, that seems to be enough of an edge case that we don't
have to worry about it.

In the specific case of the leaving-an-organization bug, that generated
a `QueryFailedError` which is why it was not sent to Rollbar.

### Checklist

## Testing Instructions
I didn't super fully test this manually since it seems to be a very straightforward change. If you know of a way to get staging to throw an unhandled, non-whitelisted error, that would be a good test. To test in development, you could simply modify some route to throw an error of some kind and then add some logging within the if block, though you may have to fiddle with the service to make the error be caught (not sure).

Another way to test on staging would be intentionally revert one of the recent bug fixes and deploy to staging but that didn't seem to be worth the trouble (though I can do that if we think it's worthwhile).

Closes #712

### DIFF
--- a/src/server/src/rollbar/rollbar-exception.filter.ts
+++ b/src/server/src/rollbar/rollbar-exception.filter.ts
@@ -30,7 +30,7 @@ export class RollbarExceptionFilter extends BaseExceptionFilter {
   catch(exception: unknown, host: ArgumentsHost): void {
     if (
       (exception instanceof HttpException && !isWhitelisted(exception)) ||
-      exception instanceof Error
+      (exception instanceof Error && !(exception instanceof HttpException))
     ) {
       const ctx = host.switchToHttp();
       const request = ctx.getRequest<Request>();
--- a/src/server/src/rollbar/rollbar-exception.filter.ts
+++ b/src/server/src/rollbar/rollbar-exception.filter.ts
@@ -28,7 +28,10 @@ export class RollbarExceptionFilter extends BaseExceptionFilter {
   }
 
   catch(exception: unknown, host: ArgumentsHost): void {
-    if (exception instanceof HttpException && !isWhitelisted(exception)) {
+    if (
+      (exception instanceof HttpException && !isWhitelisted(exception)) ||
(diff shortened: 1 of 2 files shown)


src/server/src/rollbar/rollbar-exception.filter.ts
```diff
@@ -28,7 +28,10 @@ export class RollbarExceptionFilter extends BaseExceptionFilter {
   }
 
   catch(exception: unknown, host: ArgumentsHost): void {
-    if (exception instanceof HttpException && !isWhitelisted(exception)) {
+    if (
+      (exception instanceof HttpException && !isWhitelisted(exception)) ||
+      exception instanceof Error
+    ) {
```
maurizi, line 34: Since `HttpException` is a subclass of `Error` this meant that our whitelist wasn't doing anything.

To check this, I set a breakpoint here and stepped through execution here after checking the `/healthcheck/` endpoint while the TopoJSON data was still loading.

This fixed it:
```suggestion
    if (
      (exception instanceof HttpException && !isWhitelisted(exception)) ||
      (exception instanceof Error && !(exception instanceof HttpException))
    ) {
```

maurizi (changes requested):
This has reverted the fix introduced in https://github.com/PublicMapping/districtbuilder/pull/398 to make Rollbar less-noisy for things like failed health checks during startup. I added a small suggestion that should fix it.

PR #727, opened 2021-05-03
Update map view for evaluate mode

### LINKED ISSUE #604: Evaluate page map styling changes
image

AC
 - A new district numbers layer should be added and shown when in evaluate mode
 - Map styling matches wireframes, including all acceptance criteria in the note on the left (zoom, opacity, no labels)
 - If there are currently selected geounits they should not be styled differently
 - Leaving evaluate mode should restore map state (reshow selected geounits, labels, reset opacity, etc.)

kevinearldenny:
## Overview

- Add polylabel library
- Add geojson flatten
- Remove highlighted / selected district in evaluate mode

### Checklist

### Demo

image

## Notes

I'm struggling a bit with how to make the place labels in the tiles not displayed in this view. Is this a hard requirement here?

## Testing Instructions

- Open a project and click Evaluate
- Expect: Evaluate mode looks like the above screenshot
- Close evaluate mode
- Expect: Project map is rendered in the same state before toggling evaluate mode

Closes #604


src/client/components/map/Map.tsx
```diff
@@ -360,6 +368,57 @@ const DistrictsMap = ({
...
+      .map((feature: DistrictGeoJSON) => {
+        // eslint-disable-next-line
+        return flatten(feature).reduce((prev, current) => {
+          // If a district contains multiple polygons, label the polygon with the fewest vertices
+          // eslint-disable-next-line
+          return prev.geometry.coordinates[0].length > current.geometry.coordinates[0].length
+            ? prev
+            : current;
+        });
+      })
```
kevinearldenny, line 389: This is throwing a lint error, ` Unsafe return of an any typed value` - struggling to fix this

src/client/components/map/Map.tsx
```diff
@@ -263,6 +267,8 @@ const DistrictsMap = ({
     // Add a color property to the geojson, so it can be used for styling
     geojson.features.forEach((feature, id) => {
       const districtColor = getDistrictColor(id);
+      // @ts-ignore
+      feature.properties.id = id;
```
kevinearldenny, line 271: Struggling with lint errors here as well

src/client/components/map/index.ts
```diff
@@ -345,6 +357,27 @@ export function generateMapLayers(
...
   );
 
+  map.addLayer({
+    id: DISTRICTS_EVALUATE_LABELS_LAYER_ID,
+    type: "symbol",
+    source: DISTRICTS_LABELS_SOURCE_ID,
+    layout: {
+      "text-size": 12,
+      "text-padding": 3,
+      "text-field": ["concat", regionCode, "-", ["get", "id"]],
```
kevinearldenny, line 367: @jfrankl Do we want this field to exist like this? Or to just display `id`?

src/client/components/map/Map.tsx
```diff
@@ -390,28 +451,58 @@ const DistrictsMap = ({
...
+        if (evaluateMode) {
+          // Display district labels in evaluate mode when no metric selected
+          map.setPaintProperty(DISTRICTS_LAYER_ID, "fill-opacity", 1);
+          map.setLayoutProperty(DISTRICTS_EVALUATE_LAYER_ID, "visibility", "visible");
+          map.setLayoutProperty(DISTRICTS_EVALUATE_LABELS_LAYER_ID, "visibility", "visible");
+          // Remove label layers from view
+          filteredLabelLayers.forEach(layer => {
+            console.log(`Hid ${layer}`);
+            map.setLayoutProperty(layer, "visibility", "none");
+          });
```
kevinearldenny, line 468: Based on the contents of `filteredLabelLayers`, it _seems_ like this should be hiding the place name label layers in the map... but it isn't working as expected. Any idea what's going on here?

kevinearldenny:
Feedback from Jeff

Definitely:
- Hide lock layer in Evaluate Mode

Maybe (if time)
- Change the district outline layers so that it is not a new line being added

maurizi (changes requested):
@kevinearldenny FYI the problems you were running into were more type issues than linting issues.

I removed the `geojon-flatten` library (which is really [too simple](https://github.com/tmcw/geojson-flatten/blob/master/index.js) to take on the pain of adding an untyped library) as well, which wasn't really buying us much:
```diff
diff --git a/package.json b/package.json
index 30622eb..6e26d70 100644
--- a/package.json
+++ b/package.json
@@ -26,7 +26,6 @@
     "axios": "0.21.1",
     "date-fns": "^2.19.0",
     "file-saver": "2.0.1",
-    "geojson-flatten": "^1.0.4",
     "json-stable-stringify": "^1.0.1",
     "jwt-decode": "2.2.0",
     "lodash": "^4.17.21",
diff --git a/src/client/components/map/Map.tsx b/src/client/components/map/Map.tsx
index 2ce502a..69af99e 100644
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -6,7 +6,7 @@ import { BBox2d } from "@turf/helpers/lib/geojson";
 
 import MapboxGL from "mapbox-gl";
 import "mapbox-gl/dist/mapbox-gl.css";
-import flatten from "geojson-flatten";
+import { Feature, FeatureCollection, Point, Position } from "geojson";
 import polylabel from "polylabel";
 
 import {
@@ -97,6 +97,12 @@ interface Props {
   readonly setMap: (map: MapboxGL.Map) => void;
 }
 
+interface LabelId {
+  readonly id?: string | number;
+}
+type Label = Feature<Point, LabelId>;
+type Labels = FeatureCollection<Point, LabelId>;
+
 const style: ThemeUIStyleObject = {
   legendBox: {
     position: "absolute",
@@ -316,7 +322,7 @@ const DistrictsMap = ({
     // Add a color property to the geojson, so it can be used for styling
     geojson.features.forEach((feature, id) => {
       const districtColor = getDistrictColor(id);
-      // @ts-ignore
+      // eslint-disable-next-line
       feature.properties.id = id;
       // eslint-disable-next-line
       feature.properties.outlineColor =
@@ -417,41 +423,37 @@ const DistrictsMap = ({
       }
     });
 
-  // @ts-ignore
-  const generateLabelsGeojson = geojson => {
-    const labels = geojson.features
-      // @ts-ignore
+  const generateLabelsGeojson = (geojson: DistrictsGeoJSON): Labels => {
+    // eslint-disable-next-line
+    const labels: Label[] = geojson.features
       .filter((feature: DistrictGeoJSON) => {
-        // @ts-ignore
         return feature.geometry.coordinates.length > 0 && feature.id !== 0;
       })
       .map((feature: DistrictGeoJSON) => {
-        // @ts-ignore
-        return flatten(feature).reduce((prev, current) => {
-          // If a district contains multiple polygons, label the polygon with the fewest vertices
-          // eslint-disable-next-line
-          return prev.geometry.coordinates[0].length > current.geometry.coordinates[0].length
-            ? prev
-            : current;
-        });
+        return {
+          id: feature.id,
+          coords: feature.geometry.coordinates.flat(1).reduce(
+            // eslint-disable-next-line
+            (prev: Position[], current: Position[]) => {
+              // If a district contains multiple polygons, label the polygon with the most vertices
+              return prev.length > current.length ? prev : current;
+            }
+          )
+        };
       })
-      // @ts-ignore
-      .map(feature => {
+      .map(({ id, coords }) => {
         return {
           type: "Feature",
-          // @ts-ignore
-          properties: { id: feature.id },
+          properties: { id },
           geometry: {
             type: "Point",
-            // @ts-ignore
-            coordinates: polylabel(feature.geometry.coordinates, 0.5)
+            coordinates: polylabel([coords], 0.5)
           }
         };
       });
 
     return {
       type: "FeatureCollection",
-      //@ts-ignore
       features: labels
     };
   };
@@ -463,7 +465,6 @@ const DistrictsMap = ({
     const districtsLabelsSource = map && map.getSource(DISTRICTS_LABELS_SOURCE_ID);
     districtsLabelsSource &&
       districtsLabelsSource.type === "geojson" &&
-      //@ts-ignore
       districtsLabelsSource.setData(generateLabelsGeojson(geojson));
   }, [map, geojson]);
 
diff --git a/src/client/types/geojson-flatten.d.ts b/src/client/types/geojson-flatten.d.ts
deleted file mode 100644
index 77ea14d..0000000
--- a/src/client/types/geojson-flatten.d.ts
+++ /dev/null
@@ -1 +0,0 @@
-declare module "geojson-flatten";
diff --git a/src/shared/entities.d.ts b/src/shared/entities.d.ts
index 459443b..1fb00b9 100644
--- a/src/shared/entities.d.ts
+++ b/src/shared/entities.d.ts
@@ -99,9 +99,9 @@ export type DistrictProperties = {
   readonly compactness: number;
   readonly demographics: DemographicCounts;
   readonly voting?: DemographicCounts;
-  readonly id?: number;
   /* eslint-disable */
   // NOTE: These properties are set for styling purposes
+  id?: number;
   color?: string;
   outlineColor?: string;
   percentDeviation?: number;
diff --git a/yarn.lock b/yarn.lock
index 4493543..33a1a6f 100644
--- a/yarn.lock
+++ b/yarn.lock
@@ -5531,14 +5531,6 @@ gensync@^1.0.0-beta.1:
   resolved "https://registry.yarnpkg.com/gensync/-/gensync-1.0.0-beta.1.tgz#58f4361ff987e5ff6e1e7a210827aa371eaac269"
   integrity sha512-r8EC6NO1sngH/zdD9fiRDLdcgnbayXah+mLgManTaIZJqEC1MZstmnox8KpnI2/fxQwrp5OpCOYWLp4rBl4Jcg==
 
-geojson-flatten@^1.0.4:
-  version "1.0.4"
-  resolved "https://registry.yarnpkg.com/geojson-flatten/-/geojson-flatten-1.0.4.tgz#cdfef2e9042996fcaa14fe658db6d88c99c20930"
-  integrity sha512-PpscUXxO6dvvhZxtwuqiI5v+1C/IQYPJRMWoQeaF2oohJgfGYSHKVAe8L+yUqF34PH/hmq9JlwmO+juPw+95/Q==
-  dependencies:
-    get-stdin "^7.0.0"
-    minimist "^1.2.5"
-
 geojson-vt@^3.2.1:
   version "3.2.1"
   resolved "https://registry.yarnpkg.com/geojson-vt/-/geojson-vt-3.2.1.tgz#f8adb614d2c1d3f6ee7c4265cad4bbf3ad60c8b7"
@@ -5564,11 +5556,6 @@ get-stdin@^6.0.0:
   resolved "https://registry.yarnpkg.com/get-stdin/-/get-stdin-6.0.0.tgz#9e09bf712b360ab9225e812048f71fde9c89657b"
   integrity sha512-jp4tHawyV7+fkkSKyvjuLZswblUtz+SQKzSWnBbii16BuZksJlU1wuBYXY75r+duh/llF1ur6oNwi+2ZzjKZ7g==
 
-get-stdin@^7.0.0:
-  version "7.0.0"
-  resolved "https://registry.yarnpkg.com/get-stdin/-/get-stdin-7.0.0.tgz#8d5de98f15171a125c5e516643c7a6d0ea8a96f6"
-  integrity sha512-zRKcywvrXlXsA0v0i9Io4KDRaAw7+a1ZpjRwl9Wox8PFlVCCHra7E9c4kqXCoCM9nR5tBkaTTZRBoCm60bFqTQ==
-
 get-stream@^4.0.0:
   version "4.1.0"
   resolved "https://registry.yarnpkg.com/get-stream/-/get-stream-4.1.0.tgz#c1b255575f3dc21d59bfc79cd3d2b46b1c3a54b5"
```

src/client/components/map/Map.tsx
maurizi: The issue here is that this the code to make the layers visible again afterwards is _always_ running, even when in evaluate mode. So we make the layers invisible and then immediately show them again.

src/client/components/map/Map.tsx
```diff
@@ -408,51 +423,153 @@ const DistrictsMap = ({
...
+            ["zoom"],
+            6,
+            ["*", ["get", "outlineWidthScaleFactor"], 2],
+            14,
+            ["*", ["get", "outlineWidthScaleFactor"], 5]
+          ]);
+          // Add place label layers back into view
+          filteredLabelLayers.forEach(layer => {
+            map.setLayoutProperty(layer, "visibility", "visible");
+          });
```
maurizi, line 559: One last set of layers to account for here - the geolevel label layer (controlled by the LABELS dropdown in the `MapHeader`)
https://github.com/PublicMapping/districtbuilder/blob/04ad9bdddc6e7139a0ea65f13c0ca103845fcb31/src/client/components/map/index.ts#L379-L401

src/client/components/map/Map.tsx
```diff
@@ -408,51 +423,153 @@ const DistrictsMap = ({
...
+          map.setPaintProperty(DISTRICTS_OUTLINE_LAYER_ID, "line-width", [
+            "interpolate",
+            ["linear"],
+            ["zoom"],
+            6,
+            2,
+            14,
+            5
+          ]);
+          // Remove all label layers from view except for settlements
```
maurizi, line 524: Not quite working here either - this is handling going into/out of evaluate mode directly now, but I lose the labels and they don't come back if I enter/leave evaluate mode while on a detail view, until I re-enter the summary view.

I think the structure of this `useEffect` needs to be changed up, something like this would work (methods names in place of code for brevity):
```typescript
if (evaluateMode) {
  setup_common_eval_layers();
  if (evaluateMetric) {
    setup_eval_metric_layers();
  } else {
    setup_eval_summary_layers();
  }
} else {
  setup_draw_mode_layers();
}
```
BTW I'm curious about the "except for settlements" part - was that feedback from @jfrankl ?

maurizi (changes requested):
Just one more set of layer(s) to be dealt with, and some more issues w/ getting the place layers to appear / disappear at the right times, but this is almost there 👍

src/client/components/map/Map.tsx
kevinearldenny: Jeff actually said that they were fine to stay - so I'm going to remove this bit

src/client/components/map/Map.tsx
kevinearldenny: Hmmm, do we need to remove these? Or can we just leave as is?

maurizi: Yes I think we need to hide these in evaluate mode:
image

maurizi (approved):
👍 Looks good

PR #741, opened 2021-05-11
Fix off-by-one error in handling locked districts map interaction

### LINKED ISSUE #728: District lock regression
Some of our recent development has caused a regression in the district locking feature. It's now possible to select geounits that are part of a locked district and assign them to a different district. This should be made consistent with the previous behavior, as developed in #112 and #323.

lock-regression

maurizi:
## Overview

We were properly displaying `lockedDistricts` as 0-indexed, but the click-handler code was still treating it as `1-indexed` as we did before #677.

### Checklist

## Testing Instructions

- On `develop`, locking a district will appear to work, but interactions will as as though the district before it in the list is locked.
- On this branch, the visual lock indication should match behavior

Closes #728


kevinearldenny (approved):
LGTM 👍

PR #744, opened 2021-05-12
Fix process-geojson command when --voting flag is not passed

maurizi:
## Overview

Fix process-geojson command when `--voting` / `-v` flag is not passed.
This flag was intended to be optional and default to no data.

### Checklist

## Testing Instructions

- `scripts/manage process-geojson ...` should work w/o passing the `-v` flag


kevinearldenny (approved):
LGTM 👍

PR #746, opened 2021-05-12
Update copy on the org join modal

tgilcrest:
## Overview

I forgot to finalize the copy on the org join modal so I'm updating it here.

### Notes

This should look ok, but I haven't provisioned the frontend, so please make any formatting suggestions if it looks weird.

### DIFF
--- a/src/client/components/ConfirmJoinOrganization.tsx
+++ b/src/client/components/ConfirmJoinOrganization.tsx
@@ -80,9 +80,9 @@ const ConfirmJoinOrganization = ({
       </Box>
       <Flex sx={{ flexDirection: "column" }}>
         <Box>
-          Joining an organization allows you to participate their public mapping activities. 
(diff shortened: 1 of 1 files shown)


src/client/components/ConfirmJoinOrganization.tsx
```diff
@@ -80,16 +80,18 @@ const ConfirmJoinOrganization = ({
       </Box>
       <Flex sx={{ flexDirection: "column" }}>
         <Box>
```
kevinearldenny, line 82:
```suggestion
        <Box sx={{ marginBottom: 2 }}>
```

@tgilcrest I think this looks a little better with a bit of spacing between the two boxes:

With spacing:
image

Without:
image

tgilcrest: Looks great!

kevinearldenny (approved):
👍

PR #754, opened 2021-05-13
Change map visibility settings

### LINKED ISSUE #739: Map permission tweaks
Based on some feedback from advisors, we'd like to make the following changes:

* Make maps "published" by default
* Make maps marked "private" not visible to anyone in the app (including org admins, who could previously view private maps)

maurizi:
## Overview

 - Organization admins can no longer see Private maps
 - Maps now have a "Published" visibility by default

### Checklist

### Notes

## Testing Instructions

- `scripts/yarn server run build`
- `scripts/migration`
- New projects should default to a "Published" visibility
- Organization admins should no longer be able to go to a project detail or see them on the admin screen, if the project visibility is set to "Private"

Closes #739

### DIFF
--- a/src/server/src/projects/entities/project.entity.ts
+++ b/src/server/src/projects/entities/project.entity.ts
@@ -73,7 +73,7 @@ export class Project implements IProject {
   })
   lockedDistricts: readonly boolean[];
 
-  @Column({ type: "enum", enum: ProjectVisibility, default: ProjectVisibility.Private })
+  @Column({ type: "enum", enum: ProjectVisibility, default: ProjectVisibility.Published })
   visibility: ProjectVisibility;
 
   @Column({ type: "boolean", default: false })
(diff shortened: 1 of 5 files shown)


src/server/src/projects/entities/project.entity.ts
```diff
@@ -73,7 +73,7 @@ export class Project implements IProject {
   })
   lockedDistricts: readonly boolean[];
 
-  @Column({ type: "enum", enum: ProjectVisibility, default: ProjectVisibility.Private })
+  @Column({ type: "enum", enum: ProjectVisibility, default: ProjectVisibility.Published })
```
maurizi, line 76: @tgilcrest this is changing the default visibility for _all_ projects - I could also change things _just_ for projects for Organizations if we want to differentiate the two.

tgilcrest: Let's leave it as all projects for now. I think it's a good thing to aspire to that maps are public by default to encourage transparency.

kevinearldenny (approved):
LGTM 👍 

It _might_ be worthwhile to add a changelog entry since there is a migration included here, but up to you if you think it's necessary @maurizi

PR #758, opened 2021-05-13
Fix handling of switching into block editing when using keyboard shortcut

### LINKED ISSUE #747: Prevent user from toggling to block geolevel with keyboard shortcuts while selection in progress
## Description
User is able to toggle to the block geolevel when they have geounits selected at the county or block group level when accessing thru the keyboard shortcut, even though they are prevented from doing this in the UI. 

## Steps To Reproduce

1. Go to block group geolevel and select some units
2. Attempt to toggle to a smaller geolevel with 'f'

## Expected behavior
1. Application remains at block group level while geounits are selected

## Actual behavior
Application toggles to the block level

maurizi:
## Overview

I moved most of the business logic into a helper function, which is called both by `MapHeader.tsx` as it used to be, as well as by the redux ~action~ reducer, which allows the keyboard event handler to continue to be very simple.

### Checklist

## Testing Instructions

Switching between geolevels using S / F should behave _exactly_ as it does when clicking the buttons in the header

Scenarios to check:
 - A read-only map (I used an incognito window)
 - A map where you haven't done block editing before, as well as one where you have
 - How it behaves when there are pending changes

Closes #747

### DIFF
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -41,6 +41,7 @@ export const setSelectionTool = createAction("Set selection tool")<SelectionTool
 export const setGeoLevelIndex = createAction("Set geoLevel index")<{
   readonly index: number;
   readonly isReadOnly: boolean;
+  readonly skipModal?: boolean;
 }>();
 
 export const setMapLabel = createAction("Set map label")<string | undefined>();
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -38,7 +38,7 @@ export const clearHighlightedGeounits = createAction("Clear highlighted geounit 
 
 export const setSelectionTool = createAction("Set selection tool")<SelectionTool>();
 
-export const setGeoLevelIndex = createAction("Set geoLevel index")<number>();
+export const setGeoLevelIndex = createAction("Set geoLevel index")<readonly [number, boolean]>();
 
 export const setMapLabel = createAction("Set map label")<string | undefined>();
 
--- a/src/client/actions/districtDrawing.ts
+++ b/src/client/actions/districtDrawing.ts
@@ -38,7 +38,10 @@ export const clearHighlightedGeounits = createAction("Clear highlighted geounit 
 
 export const setSelectionTool = createAction("Set selection tool")<SelectionTool>();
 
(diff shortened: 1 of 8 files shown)


src/client/actions/districtDrawing.ts
```diff
@@ -38,7 +38,7 @@ export const clearHighlightedGeounits = createAction("Clear highlighted geounit
 
 export const setSelectionTool = createAction("Set selection tool")<SelectionTool>();
 
-export const setGeoLevelIndex = createAction("Set geoLevel index")<number>();
+export const setGeoLevelIndex = createAction("Set geoLevel index")<readonly [number, boolean]>();
```
kevinearldenny, line 41: Might be helpful to add a comment here explaining what the params are, since it isn't super obvious what the boolean var is here

src/client/reducers/districtDrawing.ts
```diff
@@ -259,10 +264,32 @@ const districtDrawingReducer: LoopReducer<ProjectState, Action> = (
...
+        const canSwitch = canSwitchGeoLevels(
+          state.undoHistory.present.state.geoLevelIndex,
+          index,
+          geoLevelHierarchy,
+          state.undoHistory.present.state.selectedGeounits
+        );
+
+        return !canSwitch
+          ? loop(state, Cmd.none)
+          : isReadOnly || advancedEditingEnabled
```
kevinearldenny, line 285:
```suggestion
          : isReadOnly || advancedEditingEnabled || index < geoLevelHierarchy.length - 1
```

Currently, if you attempt to toggle to a smaller geounit (e.g., `county` -> `blockGroup`), the modal is displayed even if you are not toggling to `block`. This modal should only be displayed toggling to `block`.

kevinearldenny (changes requested):
I noticed a few bugs here - the Advanced Editing modal appears to get triggered when toggling from County -> Block Group, which shouldn't happen although this seems to be a fairly easy fix.

I also noticed that clicking `Start block editing` in the AdvancedEditingModal doesn't seem to set the current geolevel to `Blocks`, which seems like it could potentially be confusing for users.

src/client/actions/districtDrawing.ts
maurizi: I decided to rework this to use an interface instead of a tuple for clarity.

kevinearldenny (approved):
LGTM 👍

PR #762, opened 2021-05-17
Add configurable population deviation to project

### LINKED ISSUE #599: Add configurable project poulation deviation
AC:
 - New `deviation` numeric field to `Project` defaulting to `5`
 - Option to set `deviation` added to Create Project Screen, defaulting to 5

kevinearldenny:
## Overview

- Add `populationDeviation` field to Project defaulting to 5 for existing projects
- Add field to CreateProjectScreen for `populationDeviation`
- Leverage `populationDeviation` field for Evaluate Mode equal population view
- Update map symbology and legend for Evaluate Mode equal population view

### Checklist

### Demo

image
image

### Notes

- My gut feeling said to use decimal (0.05) instead of percentage (5) for _storing_ the data - but this would make things a little trickier to get the frontend to work as specified in the design and was explicitly specified in the issue so I went with the latter.
- I couldn't get the % to line up properly in `CreateProjectForm` - seems like there is probably an easy solution here @jfrankl 
- ~Do we need to update the symbology / legend in the map here as well?~ Forgot that I already planned for this in #685 :)

## Testing Instructions

- `./scripts/update`
- Create a new project, and set the `populationDeviation` to be something other than 0.05
- Add some area to a district to get it close to the specified `populationDeviation` threshold
- Open evaluate mode - expect Equal Population metric to be 

Closes #599

### DIFF
--- a/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
+++ b/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
@@ -45,7 +45,7 @@ const ProjectEvaluateSidebar = ({
   const [avgCompactness, setAvgCompactness] = useState<number | undefined>(undefined);
   const [avgPopulation, setAvgPopulation] = useState<number | undefined>(undefined);
   const [geoLevel, setGeoLevel] = useState<string | undefined>(undefined);
-  const popThreshold = 0.05;
+  const popThreshold = project && project.populationDeviation / 100.0;
   useEffect(() => {
     if (geojson && !avgCompactness) {
       const features = geojson.features.slice(1).filter(f => f.properties.compactness !== 0);
@@ -84,6 +84,7 @@ const ProjectEvaluateSidebar = ({
       status:
         (avgPopulation &&
           geojson &&
+          popThreshold &&
           geojson?.features.filter(f => {
             return (
               f.id !== 0 &&
@@ -96,15 +97,18 @@ const ProjectEvaluateSidebar = ({
       description: "have equal population",
       shortText:
         "The U.S. constitution requires that each district have about the same population for a map to be considered valid.",
-      longText: `Districts are required to be "Equal Population" for a map to be considered valid. Districts are "Equal Population" when their population falls within the target threshold. The target population is the total population divided by the number of districts and the threshold is a set percentage deviation (${Math.floor(
-        popThreshold * 100
-      )}%) above or below that target.`,
+      longText:
+        (popThreshold &&
+          `Districts are required to be "Equal Population" for a map to be considered valid. Districts are "Equal Population" when their population falls within the target threshold. The target population is the total population divided by the number of districts and the threshold is a set percentage deviation (${Math.floor(
+            popThreshold * 100
+          )}%) above or below that target.`) ||
+        "",
       avgPopulation: avgPopulation || undefined,
       popThreshold: popThreshold,
       type: "fraction",
       total: geojson?.features.filter(f => f.id !== 0).length || 0,
       value:
-        avgPopulation && geojson
+        avgPopulation && geojson && popThreshold
           ? geojson?.features.filter(f => {
               return (
                 f.id !== 0 &&
--- a/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
+++ b/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
@@ -84,7 +84,7 @@ const ProjectEvaluateSidebar = ({
       status:
         (avgPopulation &&
           geojson &&
-          popThreshold &&
+          popThreshold !== undefined &&
           geojson?.features.filter(f => {
             return (
               f.id !== 0 &&
@@ -98,7 +98,7 @@ const ProjectEvaluateSidebar = ({
       shortText:
         "The U.S. constitution requires that each district have about the same population for a map to be considered valid.",
       longText:
-        (popThreshold &&
+        (popThreshold !== undefined &&
           `Districts are required to be "Equal Population" for a map to be considered valid. Districts are "Equal Population" when their population falls within the target threshold. The target population is the total population divided by the number of districts and the threshold is a set percentage deviation (${Math.floor(
             popThreshold * 100
           )}%) above or below that target.`) ||
@@ -108,7 +108,7 @@ const ProjectEvaluateSidebar = ({
       type: "fraction",
       total: geojson?.features.filter(f => f.id !== 0).length || 0,
       value:
-        avgPopulation && geojson && popThreshold
+        avgPopulation && geojson && popThreshold !== undefined
           ? geojson?.features.filter(f => {
               return (
                 f.id !== 0 &&
--- a/src/client/components/evaluate/detail/EqualPopulation.tsx
+++ b/src/client/components/evaluate/detail/EqualPopulation.tsx
@@ -67,7 +67,7 @@ const EqualPopulationMetricDetail = ({
 }) => {
   function computeRowFill(row: DistrictProperties) {
     const val = row.percentDeviation;
-    const choroplethStops = getChoroplethStops(metric.key);
+    const choroplethStops = getChoroplethStops(metric.key, metric.popThreshold);
     // eslint-disable-next-line
     for (let i = 0; i < choroplethStops.length; i++) {
       const r = choroplethStops[i];
--- a/src/client/components/evaluate/detail/EqualPopulation.tsx
+++ b/src/client/components/evaluate/detail/EqualPopulation.tsx
@@ -90,7 +90,7 @@ const EqualPopulationMetricDetail = ({
       <Heading as="h2" sx={{ variant: "text.h5", mt: 4 }}>
         {metric.value?.toString() || " "} of {metric.total} districts are within{" "}
         {"popThreshold" in metric &&
-          metric.popThreshold &&
+          metric.popThreshold !== undefined &&
           ` ${Math.floor(metric.popThreshold * 100)}%`}{" "}
         of the target ({metric.avgPopulation && Math.floor(metric.avgPopulation).toLocaleString()})
       </Heading>
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -333,7 +333,8 @@ const DistrictsMap = ({
         maxZoom,
         map,
         geojson,
-        evaluateMetric && "avgPopulation" in evaluateMetric ? evaluateMetric : undefined
+        evaluateMetric && "avgPopulation" in evaluateMetric ? evaluateMetric : undefined,
+        project && project.populationDeviation ? project.populationDeviation : undefined
       );
 
       setMap(map);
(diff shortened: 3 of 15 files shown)


src/client/screens/CreateProjectScreen.tsx
```diff
@@ -384,103 +388,142 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps)
...
+                                data: {
+                                  ...data,
+                                  populationDeviation
+                                }
+                              });
+                            }
+                          }}
+                        />
+                        <Box sx={{ display: "inline-block" }}>%</Box>
+                      </Box>
```
kevinearldenny, line 523: @jfrankl Can you help with CSS to make the % appear inline with the `InputField`?

src/server/migrations/1620920619830-project-population-deviation.ts
```diff
@@ -0,0 +1,16 @@
...
+        await queryRunner.query(`ALTER TABLE "project" ADD "populationDeviation" numeric NOT NULL DEFAULT '0.05'`);
+        await queryRunner.query(`ALTER TABLE "project_template" ADD "populationDeviation" numeric NOT NULL DEFAULT '0.05'`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "project_template" DROP COLUMN "populationDeviation"`);
+        await queryRunner.query(`ALTER TABLE "project" DROP COLUMN "populationDeviation"`);
+    }
+
+}
```
kevinearldenny, line 16: We might want to squash this into the other migration, but I left it separate in the event we wanted to revert it

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -45,7 +45,8 @@ const ProjectEvaluateSidebar = ({
   const [avgCompactness, setAvgCompactness] = useState<number | undefined>(undefined);
   const [avgPopulation, setAvgPopulation] = useState<number | undefined>(undefined);
   const [geoLevel, setGeoLevel] = useState<string | undefined>(undefined);
-  const popThreshold = 0.05;
+  const popThreshold =
+    (project && project.populationDeviation ? project.populationDeviation : 5) / 100.0;
```
maurizi, line 49: Can we find all the places where we are using `5` as a literal value and move it to a `DEFAULT_POP_DEVIATION` (or similarly named) constant in `src/shared/constants.ts`?

We might want to change the default value in the future, it would be nice to not need to hunt it down everywhere.

src/server/migrations/1620920619830-project-population-deviation.ts
maurizi: I say squash em

src/server/src/projects/entities/project.entity.ts
```diff
@@ -82,6 +82,9 @@ export class Project implements IProject {
   @Column({ type: "boolean", default: false })
   isFeatured: boolean;
 
+  @Column({ type: "numeric", default: 5 })
```
maurizi, line 85: A few issues here (and identical issues for `project-template.entity`)
 - If you look at the JSON response, it is returning `populationDeviation` as a string, not a number
 - If you look at the default value in the DB migration, it is setting it to the _string_ `'5'` not the number `5`.
  - We've _generally_ strived to use `camel_case` in the SQL schema by using the `name` option of `@Column({ ... })` to rename the columns (see `lockedDistricts` for an example)
  
Some small tweaks to the options will fix that
 - `"numeric"` is for arbitrary-precision numbers, `"double precision"` is what you want - that matches the underlying number type used by Javascript (https://www.postgresql.org/docs/13/datatype-numeric.html)
  - Using the lambda form of setting the default got it to make the default a number instead of a string cast to a number (which I guess would also work, but while you're updating this / reworking the migrations anyways you might as well change, it seems cleaner)
```suggestion
  @Column({ type: "double precision", name: "population_deviation", default: () => "5" })
```

maurizi (changes requested):
Found a few bugs (setting deviation to `0` or a non-number just sets it to `5`, the API is returning deviation as a string not a number) as well as some suggested type cleanups

src/client/screens/CreateProjectScreen.tsx
```diff
@@ -303,6 +306,7 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps)
                   createProject({
                     ...validatedForm,
                     chamber: validatedForm.chamber || undefined,
+                    populationDeviation: validatedForm.populationDeviation || 5,
```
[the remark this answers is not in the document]
kevinearldenny: @maurizi Is a deviation of 0 acceptable? It seems improbable that this would be something that someone would actually want to do, since it would be very difficult to _exactly_ match the expected population for all districts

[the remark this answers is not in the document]
maurizi: It is very much something people want to do - it's a legal requirement for Congressional districts within a state to have exactly equal populations (or as close to it as possible, off-by-one is allowed when needed), and it's a selling point of DistrictBuilder that we make it easy to work at the block-level to produce legal 0-deviation maps.

See this explainer from our partners Public Mapping http://www.publicmapping.org/what-is-redistricting/redistricting-criteria-equal-population

src/server/src/projects/entities/project.entity.ts
kevinearldenny: @maurizi Should this be:

```
  @Column({ type: "double precision", name: "population_deviation", default: () => DEFAULT_POPULATION_DEVIATION })
```

or does it need to be cast to a string first?

```
  @Column({ type: "double precision", name: "population_deviation", default: () => `${DEFAULT_POPULATION_DEVIATION}` })
```

maurizi: The latter is what we want, from looking at the SQL migrations generated after trying both options.

src/client/screens/CreateProjectScreen.tsx
jfrankl: Screen Shot 2021-05-21 at 2 40 16 PM

I just pushed a commit changing up the design slightly and addressing this.

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -96,15 +97,18 @@ const ProjectEvaluateSidebar = ({
...
+            popThreshold * 100
+          )}%) above or below that target.`) ||
+        "",
       avgPopulation: avgPopulation || undefined,
       popThreshold: popThreshold,
       type: "fraction",
       total: geojson?.features.filter(f => f.id !== 0).length || 0,
       value:
-        avgPopulation && geojson
+        avgPopulation && geojson && popThreshold
```
maurizi, line 111:
```suggestion
        avgPopulation && geojson && popThreshold !== undefined
```

src/client/components/map/Map.tsx
```diff
@@ -333,7 +333,8 @@ const DistrictsMap = ({
         maxZoom,
         map,
         geojson,
-        evaluateMetric && "avgPopulation" in evaluateMetric ? evaluateMetric : undefined
+        evaluateMetric && "avgPopulation" in evaluateMetric ? evaluateMetric : undefined,
+        project && project.populationDeviation ? project.populationDeviation : undefined
```
maurizi, line 337:
```suggestion
        project?.populationDeviation
```

src/client/components/map/index.ts
```diff
@@ -97,9 +103,12 @@ export function getChoroplethStops(metricKey: string, threshold?: number) {
...
 }
 
-export function getChoroplethLabels(metricKey: string) {
+export function getChoroplethLabels(metricKey: string, populationDeviation?: number) {
   const compactnessLabels = ["0-30%", "30-40%", "40-50%", "50-60%", ">60%"];
-  const steps = getChoroplethStops(metricKey);
+  const steps =
+    populationDeviation !== undefined
+      ? getChoroplethStops(metricKey, populationDeviation)
+      : getChoroplethStops(metricKey);
```
maurizi, line 111: This can be simplified:
```suggestion
  const steps = getChoroplethStops(metricKey, populationDeviation);
```

src/server/src/projects/entities/create-project.dto.ts
```diff
@@ -18,6 +27,10 @@ export class CreateProjectDto implements CreateProjectData {
   @IsOptional()
   readonly districtsDefinition: DistrictsDefinition;
   @IsOptional()
+  @IsNumber()
+  @Max(100, { message: "Population deviation must be between 0% and 100%" })
```
maurizi, line 31: We don't want to allow a negative deviation.
```suggestion
  @Max(100, { message: "Population deviation must be between 0% and 100%" })
  @Min(0, { message: "Population deviation must be between 0% and 100%" })
```

src/server/src/projects/entities/update-project.dto.ts
```diff
@@ -18,6 +28,11 @@ export class UpdateProjectDto implements UpdateProjectData {
   @IsBoolean()
   @IsOptional()
   readonly advancedEditingEnabled: boolean;
+  @IsOptional()
+  @IsPositive({ message: "Population deviation must be positive" })
```
maurizi, line 32: This won't work for `0`.
```suggestion
```

src/server/src/projects/entities/update-project.dto.ts
```diff
@@ -18,6 +28,11 @@ export class UpdateProjectDto implements UpdateProjectData {
...
+  @IsNumber()
+  @Max(100, { message: "Population deviation must be between 0% and 100%" })
```
maurizi, line 34: As above in `create-project.dto`:
```suggestion
  @Max(100, { message: "Population deviation must be between 0% and 100%" })
  @Min(0, { message: "Population deviation must be between 0% and 100%" })
```

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -84,6 +84,7 @@ const ProjectEvaluateSidebar = ({
       status:
         (avgPopulation &&
           geojson &&
+          popThreshold &&
```
maurizi, line 87:
```suggestion
          popThreshold !== undefined &&
```

src/client/components/evaluate/detail/EqualPopulation.tsx
```diff
@@ -67,7 +67,7 @@ const EqualPopulationMetricDetail = ({
 }) => {
   function computeRowFill(row: DistrictProperties) {
     const val = row.percentDeviation;
-    const choroplethStops = getChoroplethStops(metric.key);
```
maurizi, line 70: Another issue with a `0` deviation below here:

```diff
diff --git a/src/client/components/evaluate/detail/EqualPopulation.tsx b/src/client/components/evaluate/detail/EqualPopulation.tsx
index bd9e64a..c3264fc 100644
--- a/src/client/components/evaluate/detail/EqualPopulation.tsx
+++ b/src/client/components/evaluate/detail/EqualPopulation.tsx
@@ -90,7 +90,7 @@ const EqualPopulationMetricDetail = ({
       <Heading as="h2" sx={{ variant: "text.h5", mt: 4 }}>
         {metric.value?.toString() || " "} of {metric.total} districts are within{" "}
         {"popThreshold" in metric &&
-          metric.popThreshold &&
+          metric.popThreshold !== undefined &&
           ` ${Math.floor(metric.popThreshold * 100)}%`}{" "}
         of the target ({metric.avgPopulation && Math.floor(metric.avgPopulation).toLocaleString()})
       </Heading>
```

maurizi (changes requested):

src/client/screens/CreateProjectScreen.tsx
```diff
@@ -159,7 +163,8 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps)
       width: "100%",
       pt: 4,
       borderTop: "1px solid",
-      borderColor: "gray.2"
+      borderColor: "gray.2",
+      display: "flex"
```
maurizi, line 167: This change is causing a styling regression when errors are displayed
```suggestion
```

maurizi (changes requested):
Left 1 inline comment re: a styling regression

maurizi (approved):

src/client/screens/CreateProjectScreen.tsx
```diff
@@ -384,103 +389,141 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps)
...
+                      <legend sx={{ ...style.cardLabel, ...style.legend, ...{ flex: "0 0 100%" } }}>
+                        Population deviation tolerance
+                      </legend>
+                      <Box
+                        id="description-districts"
+                        as="span"
+                        sx={{ ...style.cardHint, ...{ flex: "0 0 100%" } }}
+                      >
+                        Depending on whether you are making a map for learning purposes or are
+                        trying to make the exercise as realistic as possible
```
maurizi, line 502: Let's have this match the new text used in https://github.com/PublicMapping/districtbuilder/pull/848/files

PR #766, opened 2021-05-20
Update race column to handle different race categories

### LINKED ISSUE #764: Update Race Column to Handle Different Race Categories
The race column in the sidebar currently expects a certain set of race categories in our data. We want to be able to handle a different set of categories that we will be adding with our new data. But we want to continue to support the original set of columns so we don't have to reprocess all the data at once.

The original categories are:
* White
* Black
* Asian
* Hispanic
* Other

The new categories are:
* White
* Black
* Asian
* Hispanic
* Native American
* Pacific Islander

We want the front end to support either of these lists of categories. We will need new colors for the new race categories. I will provide those shortly.

kevinearldenny:
## Overview

- Adds fields for `nativeAmerican` and `pacificIslander` to demographics chart
- Allows for the existing split (with `Other`) to still be read and displayed

### Checklist

### Demo

image
image

### Notes

I've noticed a few instances where `pacificIslander` is displayed as 0% - does it make sense to display this instead as "< 1%"?

## Testing Instructions

- Process the attached geojson for Delaware: `./scripts/manage process-geojson data/input/de.geojson -o data/output/Delaware -n 12,4,4 -x 12,12,12 -d "population,white,black,asian,hispanic,native:nativeAmerican,pacific:pacificIslander"`
- Add it to the database: `./scripts/manage publish-region data/output/Delaware US DE Delaware`
- Create a new project for Delaware and view it; expect tooltip to be displayed as screenshot above
- Create a new project for PA; expect tooltip to be displayed as it was previously

Closes #764

### DIFF
--- a/src/client/components/DemographicsTooltip.tsx
+++ b/src/client/components/DemographicsTooltip.tsx
@@ -67,10 +67,20 @@ const DemographicsTooltip = ({
     (population: number) =>
       (demographics.population ? population / demographics.population : 0) * 100
   );
-  const races = ["white", "black", "asian", "hispanic", "other"] as const;
-  const rows = races.map((id: typeof races[number]) => (
-    <Row key={id} label={id} percent={percentages[id]} color={demographicsColors[id]} />
-  ));
+  const races = [
+    "white",
+    "black",
+    "asian",
+    "hispanic",
+    "nativeAmerican",
+    "pacificIslander",
+    "other"
+  ] as const;
+  const rows = races
+    .filter(race => percentages[race])
+    .map((id: typeof races[number]) => (
+      <Row key={id} label={id} percent={percentages[id]} color={demographicsColors[id]} />
+    ));
   return (
     <Box sx={{ width: "100%", minHeight: "100%" }}>
       <Styled.table sx={{ margin: "0", width: "100%" }}>
--- a/src/client/components/DemographicsTooltip.tsx
+++ b/src/client/components/DemographicsTooltip.tsx
@@ -9,6 +9,7 @@ const style: ThemeUIStyleObject = {
     textAlign: "left",
     py: 0,
     pr: 2,
+    whiteSpace: "nowrap",
     textTransform: "capitalize"
   },
   number: {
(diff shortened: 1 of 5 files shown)


src/client/constants/colors.ts
```diff
@@ -438,5 +438,7 @@ export const demographicsColors = {
   black: "#6AC59B",
   asian: "#F194C4",
   hispanic: "#C8BFF2",
+  nativeAmerican: "#9969F2",
+  pacificIslander: "#89A53A",
```
kevinearldenny, line 442: @tgilcrest I just chose these kinda randomly, let me know once you have the correct colors decided upon

tgilcrest: I worked the colors out with Jeff. Let's use these for all the colors:

```
White - #FC8D62
Black - #66C2A5
Asian - #E78AC3
Hispanic - #8DA0CB
Native American - #A6D854
Pacific Islander - #FFD92F
```

tgilcrest: (This updates all of them to a palette from colorbrewer that is approved.)

src/client/components/DemographicsTooltip.tsx
```diff
@@ -67,10 +72,20 @@ const DemographicsTooltip = ({
...
+    "white",
+    "black",
+    "asian",
+    "hispanic",
+    "nativeAmerican",
+    "pacificIslander",
+    "other"
+  ] as const;
+  const rows = races
+    .filter(race => percentages[race])
```
maurizi, line 85: We want to continue to show the row if the field is there but the value is `0`
```suggestion
    .filter(race => percentages[race] !== undefined)
```

maurizi (changes requested):
Left one comment as just an FYI, but I think we need to fix the 2nd one that I left suggested changes for.

maurizi:
> Process the attached geojson for Delaware: `./scripts/manage process-geojson data/input/de.geojson -o data/output/Delaware -n 12,4,4 -x 12,12,12 -d "population:population,white:white,black:black,asian:asian,hispanic:hispanic,native:nativeAmerican,pacific:pacificIslander"`

FYI you can simplify that command a bit (you only need to provide key-value pairs if you're renaming): `./scripts/manage process-geojson data/input/de.geojson -o data/output/Delaware -n 12,4,4 -x 12,12,12 -d "population,white,black,asian,hispanic,native:nativeAmerican,pacific:pacificIslander"`

maurizi (approved):

PR #776, opened 2021-05-21
Cleanup text in evaluate mode

### LINKED ISSUE #767: Evaluate mode text cleanup
We want to deploy an early version of evaluate mode but there are still a few places we need to update the placeholder text. And we also want to hide the two features that are not complete yet (competitiveness and majority minority).

Here are the places we should update (*NOTE: Please also grep for any additional `lorem ipsum`s I may have missed*):

#### Main page
* Hide Majority Minority button (this isn't done yet, so we should just hide this element)
* Hide Competitiveness button (same as MM)
* Updated Equal Population description
  * The U.S. Constitution requires that each district have about the same population for a map to be considered valid.
* Update Contiguity description
  * All parts of a district must be in physical contact with some other part of the district, to the extent possible.
* Update Compactness description
  * A district that is more spread out has a lower compactness. Low compactness can be an indicator of gerrymandering.
* Update County split description
  * County splits occur when a county is split between two or more districts. Some states require minimizing county splits, to the extent practicable.

#### Equal population page
* Update Description
  * Districts are required to be "Equal Population" for a map to be considered valid. Districts are "Equal Population" when their population falls within the target threshold. The target population is the total population divided by the number of districts and the threshold is a set percentage deviation above or below that target.

#### Contiguity page
* Update Description
  * Each district should be contiguous, meaning it must be a single, unbroken shape. Some exceptions are allowed, such as the inclusion of islands in a coastal district. Two areas connected only by a single point  (touching just their corners) are not considered contiguous. 

#### Compactness page
* Update Description
  * A district that efficiently groups constituents together has higher compactness. Low compactness or districts that branch out to different areas can be indicators of gerrymandering. Compactness is calculated using the Polsby-Popper method. Higher numbers are better.

#### County Splits page
* Update Description
  * County splits occur when a county is split between two or more districts.  Some states require minimizing county splits, to the extent practicable.

kevinearldenny:
## Overview

- Make text updates as specified in #767 
- Update flag on equal population

### Checklist

### Demo

image

### Notes

One copy Q for the Equal Population page - does it make sense to specify the threshold for the project in the header text? My inclination is yes, but just wanted to confirm @tgilcrest 

image

## Testing Instructions

- Review all pages in Evaluate Mode to confirm that the text matches what is expected

Closes #767

### DIFF
--- a/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
+++ b/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
@@ -87,9 +87,7 @@ const ProjectEvaluateSidebar = ({
           geojson?.features.filter(f => {
             return (
               f.id !== 0 &&
-              // @ts-ignore
               f.properties.percentDeviation &&
-              // @ts-ignore
               Math.abs(f.properties.percentDeviation) <= popThreshold
             );
           }).length ===
@@ -110,9 +108,7 @@ const ProjectEvaluateSidebar = ({
           ? geojson?.features.filter(f => {
               return (
                 f.id !== 0 &&
-                // @ts-ignore
                 f.properties.percentDeviation &&
-                // @ts-ignore
                 Math.abs(f.properties.percentDeviation) <= popThreshold
               );
             }).length
--- a/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
+++ b/src/client/components/evaluate/ProjectEvaluateSidebar.tsx
@@ -81,12 +81,26 @@ const ProjectEvaluateSidebar = ({
     {
       key: "equalPopulation",
       name: "Equal Population",
-      status: false,
+      status:
+        (avgPopulation &&
+          geojson &&
+          geojson?.features.filter(f => {
+            return (
+              f.id !== 0 &&
+              // @ts-ignore
+              f.properties.percentDeviation &&
+              // @ts-ignore
+              Math.abs(f.properties.percentDeviation) <= popThreshold
+            );
+          }).length ===
+            geojson?.features.length - 1) ||
+        false,
       description: "have equal population",
-      shortText: "Lorem ipsum",
-      longText: `This map has a goal of getting within ${Math.floor(
+      shortText:
+        "The U.S. constitution requires that each district have about the same population for a map to be considered valid.",
+      longText: `Districts are required to be "Equal Population" for a map to be considered valid. Districts are "Equal Population" when their population falls within the target threshold. The target population is the total population divided by the number of districts and the threshold is a set percentage deviation (${Math.floor(
         popThreshold * 100
-      )}% of ${avgPopulation?.toLocaleString()}`,
+      )}%) above or below that target.`,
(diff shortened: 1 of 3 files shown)


src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -80,12 +80,26 @@ const ProjectEvaluateSidebar = ({
...
+      status:
+        (avgPopulation &&
+          geojson &&
+          geojson?.features.filter(f => {
+            return (
+              f.id !== 0 &&
+              // @ts-ignore
+              f.properties.percentDeviation &&
+              // @ts-ignore
+              Math.abs(f.properties.percentDeviation) <= popThreshold
```
maurizi, line 92: There aren't any type errors here?
```suggestion
              f.properties.percentDeviation &&
              Math.abs(f.properties.percentDeviation) <= popThreshold
```

tgilcrest:
> One copy Q for the Equal Population page - does it make sense to specify the threshold for the project in the header text? My inclination is yes, but just wanted to confirm

Yes, that's a nice addition.

tgilcrest:
@kevinearldenny Can you also update the lower header text on the equal pop. page to be structured like this?

> 2 of 2 districts are within 5% of the target population (478,624)

tgilcrest (approved):
From the screenshots it looks good to me. Just had the one change requested in the other comment.

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
maurizi: Realizing now you copy/pasted this from below where the `@ts-ignore` comments are... also not needed?

maurizi (approved):
Working 👍

kevinearldenny:
> @kevinearldenny Can you also update the lower header text on the equal pop. page to be structured like this?
> 
> > 2 of 2 districts are within 5% of the target population (478,624)

@tgilcrest @maurizi Do we still want this change? It looks like it's breaking to a second line, which looks a little weird to me.

maurizi:
> @tgilcrest @maurizi Do we still want this change? It looks like it's breaking to a second line, which looks a little weird to me.

In that case I say skip it

tgilcrest:
Can we do this?

>2 of 2 districts are within 5% of the target (478,624)

src/client/components/evaluate/detail/EqualPopulation.tsx
```diff
@@ -88,11 +88,11 @@ const EqualPopulationMetricDetail = ({
   return (
     <Box>
       <Heading as="h2" sx={{ variant: "text.h5", mt: 4 }}>
-        {metric.value?.toString() || " "} of the {metric.total} districts are within{" "}
+        {metric.value.toString() || " "} of {metric.total} districts are within{" "}
```
maurizi, line 91: This should fix the lint
```suggestion
        {metric.value?.toString() || " "} of {metric.total} districts are within{" "}
```

PR #778, opened 2021-05-21
Fix calculation of average compactness

### LINKED ISSUE #772: Evaluate: Remove zero geo-unit and non-contiguous districts from average compactness calculation
The "Average compactness" calculation on the Evaluate Compactness page includes all the districts in the map. This makes sense for maps that are complete, but for maps that are incomplete it skews the average. In the calculation, it seems like we are assigning a 0% compactness to the district and including it in the calculation. We should instead remove these districts from the average compactness calculation, and only calculate average compactness for the districts for which we display a compactness value (empty districts/districts that contain >0 geo-units).

I'm not sure how non-contiguous districts are being treated in the average compactness calculation, but we should also remove these from the calculation, rather than assigning 0% compactness to them (if we are).

maurizi:
## Overview

Excludes empty or non-contiguous districts when calculating average compactness

### Checklist

### Demo

image

## Testing Instructions

- Go to the compactness detail on evaluate mode. Empty & non-contiguous districts should not be included in the calculation

Closes #772

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,18 +14,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add voting info to map tooltip [#751](https://github.com/PublicMapping/districtbuilder/pull/751)
 - Disable some keyboard shortcuts in reads-only mode [#768](https://github.com/PublicMapping/districtbuilder/pull/768)
 
-- Update race column to handle different race categories [#766](https://github.com/PublicMapping/districtbuilder/pull/766)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


tgilcrest:
In the screenshot, is something off with the calculation? The average is 44% but all the numbers I see in the list are less than that.  The average I calculate is 37% (42+41+39+24=146 -> /4 = 36.5)

maurizi:
@tgilcrest good catch!

I think I have everything fixed now, and I also caught an unrelated issue; when all the districts were empty / non-contiguous it would show the compactness as `NaN%`

kevinearldenny:
image

LGTM 👍

kevinearldenny (approved):

PR #779, opened 2021-05-22
Include unassigned district when switching using keyboard shortcuts

### LINKED ISSUE #771: Keyboard shortcuts should not skip the null district
## Description
Pressing the keyboard shortcut for previous or next district should not skip the null/unassigned district. A map maker sometimes need to quickly switch back to this unassigned layer, and making it accessible from the keyboard would help.

maurizi:
## Overview

Allow switching to the unassigned district when using keyboard shortcuts.

### Checklist

### Demo

### Notes

## Testing Instructions

- Use `e` and `d` to switch between districts. Ensure you can switch to every district and the unassigned district

Closes #771


pcaisse (approved):
LGTM! I think this should reference issue #771 though

PR #784, opened 2021-05-24
Disable keyboard shortcuts in evaluate mode

### LINKED ISSUE #782: Disable keyboard shortcuts on Evaluate mode (except toggle)
When testing evaluate mode I found the keyboard shortcuts related to map editing were still active. This allows you to do things like accept pending changes and switch between geounit levels, which shouldn't be possible in Evaluate mode.

##### AC
* When Evaluate mode is toggled on, the only shortcuts that are active are `T`, to toggle off Evaluate mode, and `?` to show the shortcut menu.
* Only those two shortcuts show up in the shortcut menu when evaluate mode is on
* If the user has an active selection when they toggle into Evaluate mode, preserve that when the toggle back out (this is the current behavior; we don't want this to change.)
* Hitting `?` when the shortcut menu is open closes it (tangent here, but something I think we should fix).

##### Example
Here I:
1. Select a geounit (I do not click Accept)
2. Toggle Evaluate mode with `T`
3. Hit `G` to accept changes while in Evaluate mode (see that the district is updated while in Evaluate mode.)

kevinearldenny:
## Overview

- Disables all keyboard shortcuts in evaluate mode expect 'Toggle evaluate mode'
- Closes keyboard shortcuts modal with "?"

### Checklist

### Demo

image

## Testing Instructions

- Open a project and go to evaluate mode
- Confirm that none of the keyboard shortcuts are available in evaluate mode besides `T` to toggle evaluate mode and `?` to toggle the modal

Closes #782

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Update race column to handle different race categories [#766](https://github.com/PublicMapping/districtbuilder/pull/766)
 - Cleanup text in evaluate mode [#776](https://github.com/PublicMapping/districtbuilder/pull/776)
+- Disable keyboard shortcuts in evaluate mode [#784](https://github.com/PublicMapping/districtbuilder/pull/784)
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


maurizi (approved):
Perfect 👍

PR #786, opened 2021-05-24
Display evaluate mode button for read only maps

### LINKED ISSUE #785: Add Evaluate mode to view-only plans
We should make Evaluate mode available to view-only plans.

##### AC
* When a user opens a view-only plan, the Evaluate button is available for them to use
* Clicking Evaluate opens Evaluate mode as it does normally for non-view-only plans

kevinearldenny:
## Overview

- Displays the evaluate mode button for read only maps just like for a user's maps

### Checklist

### Demo

image

## Testing Instructions

- Go to a read only map; expect Evaluate button to visible in header

Closes #785

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Display evaluate mode toggle button in read-only mode [#786](https://github.com/PublicMapping/districtbuilder/pull/786)
+
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


tgilcrest:
This is probably a dumb question, but the button also opens Evaluate mode for the map, right? And all the detail pages work?

kevinearldenny:
> This is probably a dumb question, but the button also opens Evaluate mode for the map, right? And all the detail pages work?

Yep, that's correct. Was a bit of an oversight that this button was previously hidden for read-only maps tbh

maurizi (approved):
Working great 👍

PR #675, opened 2021-04-05
Refactor: Use local state to avoid table dissappearing during API update

maurizi:
## Overview

Refactors the organization admin screen to avoid the table flickering when featuring/unfeaturing a project.

### Checklist

- [x] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

## Testing Instructions

- Go to the org admin screen, and feature/unfeature some projects. On `develop`, there will be a flicker where the table disappears momentarily, on this branch there will not.

Closes #640

PR #796, opened 2021-05-26
List all published maps

### LINKED ISSUE #719: Add page listing all published maps
No wireframes for this feature yet - we'd like to crib the look/feel of featured maps on the Organization screen (w/ minimaps) but with pagination.

AC:
 - Add a new screen at `/maps` that shows all maps w/ a Published `visibility`
 - Maps should show mini-map like Featured maps in Organizations, but sorted by date, w/ most recently updated first
 - Show Map name, region, first name, created
 - Server-side pagination is almost certainly required for this, as the # of maps could grow to be very large
 - We want an API to power this page. API endpoint should return:
   - user name / id
   - plan id
   - region / region id
   - chamber / chamer id
   - create date
   - update date

kevinearldenny:
## Overview

- Adds a new endpoint, `api/globalProjects` with server-side pagination
- Adds a new page in the frontend `/global` containing a display of all projects paginated with a limit of 10 projects per page
- Created a new reusable `PaginationFooter` component 

### Checklist

### Demo

image
image

### Notes

I struggled a bit with the pagination implementation in Redux - initially, I was thinking that `globalProjectsFetchPage` should trigger `globalProjectsFetch`, but ended up having this handled through a param function + `useEffect` in `PaginationFooter`. After further reflection, this actually might not be a bad thing as far as making the component modular for usage in other pagination implementations - but we probably should clean up / rename the actions if we want to stick with this method.

## Testing Instructions

- Go to http://localhost:3003/global
- Make sure that you have at least 11 maps; expect pagination footer to be displayed
- Click on page 2 in the pagination footer; expect maps displayed on screen to change

Closes #719

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "lodash": "^4.17.21",
     "mapbox-gl": "1.6.1",
     "memoizee": "0.4.14",
+    "nestjs-typeorm-paginate": "^2.5.1",
     "polylabel": "^1.1.0",
     "prop-types": "15.7.2",
     "react": "16.11.0",
--- a/package.json
+++ b/package.json
@@ -31,7 +31,6 @@
     "lodash": "^4.17.21",
     "mapbox-gl": "1.6.1",
     "memoizee": "0.4.14",
-    "nestjs-typeorm-paginate": "^2.5.1",
     "polylabel": "^1.1.0",
     "prop-types": "15.7.2",
     "react": "16.11.0",
--- a/src/client/components/PaginationFooter.tsx
+++ b/src/client/components/PaginationFooter.tsx
@@ -1,0 +1,63 @@
+/** @jsx jsx */
+import { jsx, Box, Button } from "theme-ui";
+import "../App.css";
+import { range } from "lodash";
+
+interface StateProps {
+  readonly currentPage: number;
+  readonly totalPages: number;
+  // eslint-disable-next-line
+  readonly setPage: (number: number) => void;
+}
+
+const style = {
+  pagination: {
+    // TODO: Make this not look bad
+    cursor: "pointer",
+    display: "inline",
+    mr: 1
+  },
+  paginationSelected: {
+    // TODO: Make this not look bad
+    cursor: "pointer",
+    fontWeight: 700,
+    display: "inline"
+  },
+  pageList: {
+    display: "inline-block"
+  }
+};
+
+const PaginationFooter = ({ currentPage, totalPages, setPage }: StateProps) => {
+  // This component is to be used for server-side pagination with a list of elements rendered separately.
+
+  const pageNumbers = range(1, totalPages + 1);
+
+  const renderPageNumbers =
+    pageNumbers &&
+    pageNumbers.map(number => {
+      return (
+        <li
+          key={number}
+          id={number.toString()}
+          sx={number === currentPage ? style.paginationSelected : style.pagination}
+        >
+          <Button disabled={number === currentPage} onClick={() => setPage(number)}>
+            {number}
+          </Button>
+        </li>
+      );
+    });
+
+  return (
+    <Box>
+      {
+        <ul id="page-numbers" sx={style.pageList}>
+          {renderPageNumbers}
+        </ul>
+      }
+    </Box>
+  );
+};
+
+export default PaginationFooter;
--- a/src/client/reducers/projects.ts
+++ b/src/client/reducers/projects.ts
@@ -9,22 +9,29 @@ import {
   projectsFetch,
   projectsFetchFailure,
   projectsFetchSuccess,
+  globalProjectsFetch,
+  globalProjectsFetchSuccess,
+  globalProjectsFetchFailure,
   setDeleteProject
 } from "../actions/projects";
 
 import { IProject } from "../../shared/entities";
-import { fetchProjects, patchProject } from "../api";
+import { fetchAllPublishedProjects, fetchProjects, patchProject } from "../api";
 import { showResourceFailedToast } from "../functions";
 import { Resource } from "../resource";
 
 export interface ProjectsState {
   readonly projects: Resource<readonly IProject[]>;
+  readonly globalProjects: Resource<readonly IProject[]>;
   readonly deleteProject?: IProject;
   readonly archiveProjectPending: boolean;
 }
 
 export const initialState = {
+  // User projects
   projects: { isPending: false },
+  // All projects
+  globalProjects: { isPending: false },
   archiveProjectPending: false
 };
 
@@ -58,6 +65,31 @@ const projectsReducer: LoopReducer<ProjectsState, Action> = (
         },
         Cmd.run(showResourceFailedToast)
       );
+      case getType(globalProjectsFetch):
+        return loop(
+          {
+            ...state,
+            globalProjects: { isPending: true }
+          },
+          Cmd.run(fetchAllPublishedProjects, {
+            successActionCreator: globalProjectsFetchSuccess,
+            failActionCreator: globalProjectsFetchFailure,
+            args: [] as Parameters<typeof fetchProjects>
+          })
+        );
+      case getType(globalProjectsFetchSuccess):
+        return {
+          ...state,
+          globalProjects: { resource: action.payload }
+        };
+      case getType(globalProjectsFetchFailure):
+        return loop(
+          {
+            ...state,
+            globalProjects: { errorMessage: action.payload }
+          },
+          Cmd.run(showResourceFailedToast)
+        );
     case getType(setDeleteProject):
       return {
         ...state,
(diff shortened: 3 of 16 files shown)


src/client/components/PaginationFooter.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [startingPage, setStartingPage] = useState<number | undefined>(undefined);
+
+  useEffect(() => {
+    // Load additional elements every time page changes after initial render
+    if (!startingPage) {
+      setStartingPage(currentPage);
+    } else {
+      onPageChange();
+    }
+  }, [currentPage]);
```
kevinearldenny, line 42: Does this feel clunky?

src/client/reducers/projects.ts
```diff
@@ -58,6 +77,43 @@ const projectsReducer: LoopReducer<ProjectsState, Action> = (
...
+          );
+        }
+      case getType(globalProjectsFetchPage):
+        return loop(
+          {
+            ...state,
+            globalProjectsPagination: { ...state.globalProjectsPagination , currentPage: action.payload }
+          },
+          Cmd.run(globalProjectsFetch)
+        );
```
kevinearldenny, line 102: Think I just figured out why this isn't working - it should be `Cmd.action` instead of `Cmd.run`, right?

That being said... do you like chaining these together better? Or would it be better to take the chained action out and rename this to `globalProjectsSetPage`?

src/client/types.d.ts
```diff
@@ -17,6 +18,17 @@ export interface DynamicProjectData {
...
+  readonly items: T[];
+  readonly meta: {
+    currentPage: number;
+    itemCount: number;
+    itemsPerPage: number;
+    totalItems: number;
+    totalPages: number;
+  }
+}
+
```
kevinearldenny, line 31: Reusable type created for any other places we decide to use `nestjs-typeorm-paginate`

src/client/reducers/projects.ts
```diff
@@ -58,6 +77,42 @@ const projectsReducer: LoopReducer<ProjectsState, Action> = (
...
```
kevinearldenny, line 101:
```suggestion
      case getType(globalProjectsSetPage):
        return {
            ...state,
            globalProjectsPagination: { ...state.globalProjectsPagination , currentPage: action.payload }
          };
```

In the current solution (where `PaginationFooter` is watching for a change in `pagination.currentPage` and then calling `globalProjectsFetch()` in `onPageChange`, this is unnecessary. However, I'm going back and forth on whether I like the current architecture more than what I was _originally_ thinking here of `globalProjectsFetchPage`. Curious your thoughts, @maurizi

src/client/screens/PublishedMapsListScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+    !globalProjects && store.dispatch(globalProjectsFetch());
+  }, [globalProjects]);
+
+  return (
+    <Flex sx={{ height: "100%", flexDirection: "column" }}>
+      <SiteHeader user={user} />
+      <Box>
+        {globalProjects ? (
+          // @ts-ignore
+          <Box sx={style.projects}>
```
maurizi, line 66: Using `as const` for `style` will fix the typing issues.

```suggestion
} as const;

const PublishedMapsListScreen = ({ globalProjects, user, pagination }: StateProps) => {
  useEffect(() => {
    !globalProjects && store.dispatch(globalProjectsFetch());
  }, [globalProjects]);

  return (
    <Flex sx={{ height: "100%", flexDirection: "column" }}>
      <SiteHeader user={user} />
      <Box>
        {globalProjects ? (
          <Box sx={style.projects}>
```

src/server/src/projects/controllers/globalProjects.controller.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import { JwtAuthGuard, OptionalJwtAuthGuard } from "../../auth/guards/jwt-auth.guard";
+import { RegionConfig } from "../../region-configs/entities/region-config.entity";
+import { Project } from "../entities/project.entity";
+import { ProjectsService } from "../services/projects.service";
+
+import { UpdateProjectDto } from "../entities/update-project.dto";
+import { Pagination } from "nestjs-typeorm-paginate";
+
+@Crud({
```
maurizi, line 39: Why are we using `@Crud()` at all for this controller? Do we even use any of it's features? It doesn't look like `getAllGlobalProjects` does.

yarn.lock
```diff
@@ -7850,6 +7850,11 @@ nested-property@^1.0.2:
   resolved "https://registry.yarnpkg.com/nested-property/-/nested-property-1.0.4.tgz#4068c4289d1f3ac9f367e784502ff342d277ee2f"
   integrity sha512-6fNIumJJUyP3rkB4FyVYCYpdW+PKUCaxRWRGLLf0kv/RKoG4mbTvInedA9x3zOyuOmOkGudKuAtPSI+dnhwj2g==
 
+nestjs-typeorm-paginate@^2.5.1:
+  version "2.5.1"
+  resolved "https://registry.yarnpkg.com/nestjs-typeorm-paginate/-/nestjs-typeorm-paginate-2.5.1.tgz#2e50935113443e29da4dc5e45a16a4281a8f852f"
+  integrity sha512-cSUKRpY3jmKT/ngOlYa0kmZ3XXY4VNgCPXiud6RxmPv5swZRFnZ1uy65A9rI9TEboSzvo0nwUNQfnz9g5FX2SQ==
+
```
maurizi, line 7857: Similar to above - Why is this in the `client` `yarn.lock`?

package.json
```diff
@@ -31,6 +31,7 @@
     "lodash": "^4.17.21",
     "mapbox-gl": "1.6.1",
     "memoizee": "0.4.14",
+    "nestjs-typeorm-paginate": "^2.5.1",
```
maurizi, line 34: Isn't this a server-side library? Why is it in the `package.json` for `client`?

src/client/screens/PublishedMapsListScreen.tsx
```diff
@@ -0,0 +1,110 @@
...
+const PublishedMapsListScreen = ({ globalProjects, user, pagination }: StateProps) => {
+  useEffect(() => {
+    !globalProjects && store.dispatch(globalProjectsFetch());
+  }, [globalProjects]);
+
+  return (
+    <Flex sx={{ height: "100%", flexDirection: "column" }}>
+      <SiteHeader user={user} />
+      <Box>
+        {globalProjects ? (
```
maurizi, line 64: In practice I find it very difficult to interact with this UI element when it is switched to a loading icon after I click the pagination button.

Can we follow the strategy used in https://github.com/PublicMapping/districtbuilder/pull/675/files and keep a copy of `globalProjects` to keep this from happening?

src/client/components/PaginationFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+    pageNumbers.map(number => {
+      return (
+        <li
+          key={number}
+          id={number.toString()}
+          sx={number === currentPage ? style.paginationSelected : style.pagination}
+          onClick={() => setPage(number)}
+        >
+          {number}
+        </li>
```
maurizi, line 65:
```suggestion
        <li
          key={number}
          id={number.toString()}
          sx={number === currentPage ? style.paginationSelected : style.pagination}
        >
          <Button disabled={number === currentPage} onClick={() => setPage(number)}>{number}</Button>
        </li>
```

src/client/components/PaginationFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+    // Populate page numbers
+    /* eslint-disable */
+    if (totalPages) {
+      const pn = [];
+      for (let i = 1; i <= totalPages; i++) {
+        pn.push(i);
+      }
+      setPageNumbers(pn);
+    }
+    /* eslint-enable */
```
maurizi, line 51: Consider replacing this with [`_.range(1, totalPages)`](https://lodash.com/docs/4.17.15#range)

src/client/components/PaginationFooter.tsx
maurizi: Yes - I think your initial instinct to have `globalProjectsFetchPage` trigger `globalProjectsFetch` was the right one.

src/client/components/PaginationFooter.tsx
```diff
@@ -0,0 +1,72 @@
...
+    cursor: "pointer",
+    fontWeight: 700
+  }
+};
+
+const PaginationFooter = ({ currentPage, totalPages, setPage, onPageChange }: StateProps) => {
+  // This component is to be used for server-side pagination with a list of elements rendered separately.
+
+  const [pageNumbers, setPageNumbers] = useState<readonly number[] | undefined>(undefined);
+  const [startingPage, setStartingPage] = useState<number | undefined>(undefined);
```
maurizi, line 30: `pageNumbers` is just a simple transformation of the provided parameters, there is no need to keep these in component state - you can just directly assign it:

```typescript
  const pageNumbers = _.range(1, totalPages);
```

maurizi (changes requested):
In general this is functional (and I like `nestjs-typeorm-paginate`, that made the backend very simple 😄 ), but could use some polish - I left a few comments with changes I'd like to see.

Also, can you open an issue to do a style pass on this page + label it `design`?

kevinearldenny:
> In general this is functional (and I like `nestjs-typeorm-paginate`, that made the backend very simple 😄 ), but could use some polish - I left a few comments with changes I'd like to see.
> 
> Also, can you open an issue to do a style pass on this page + label it `design`?

Added a new design issue #799

maurizi (changes requested):
This is looking really nice, just needs some lint cleanup as well as the comment I left a suggestion for addressed.

maurizi (approved):
Looking good 👍

PR #797, opened 2021-05-27
Ignore unassigned districts when computing flag for Contiguity and Equal Population(*)

### LINKED ISSUE #749: Evaluate Mode: Contiguity should not necessarily fail on a partially completed map
## Description
For an in progress map, an empty/not-started district should not cause contiguity to fail on evaluate mode. As a map maker, I only want to see that test fail when one or more of my districts are failing the contiguity check, so I can go in and fix them.

## Steps To Reproduce

1. Create a map
2. Add one district that is contiguous
3. Go to evaluate mode -> contiguity

## Expected behavior
I'd expect to see a green check mark

## Actual behavior
I see a red "X"

## Screenshots
image

kevinearldenny:
## Overview

- Ignores districts without assigned areas when computing contiguity and equal population flags in evaluate mode

### Checklist

### Demo

image

### Notes

* I added similar logic for the Equal Population metric, since it doesn't really make sense that it would fail if all of the districts created were acceptable. However, thinking through this further - does it make sense to add a 3rd 'incomplete' state for these flags? I think this might help to acknowledge that the plan is not truly valid yet, since there are still unassigned areas and missing districts.

## Testing Instructions

1. Create a map
2. Add one district that is contiguous and has a population within the accepted deviation tolerance
3. Go to evaluate mode and review metrics - expect that the contiguity and equal population flags are both displayed as true

Closes #749


maurizi (approved):
Looks good to me 👍

PR #806, opened 2021-06-03
Fix sidebar and EqualPopulation metric for population deviation of 0

### LINKED ISSUE #801: We should correctly grade projects with a population deviation of 0
For the following 0-deviation map:
[PA 0 dev.csv](https://github.com/PublicMapping/districtbuilder/files/6584303/PA.0.dev.csv)

We currently incorrectly show the map as having unequal-population districts, in both the sidebar and evaluate mode:
localhost_3003_projects_4ce7f052-51d0-4e46-bce4-cc1a86e8243f

localhost_3003_projects_4ce7f052-51d0-4e46-bce4-cc1a86e8243f (1)

AC:
For maps with a population deviation of `0`:
 - Districts have equal population if they have a deviation of exactly `0`, or a deviation of `1` where the total population does not divide evenly into the number of districts
 - This should be reflected on both the sidebar and the evaluate mode Equal population detail & summary

kevinearldenny:
## Overview

- Fixes the flag in ProjectSidebar to display correctly with a population deviation of 0
- Fixes choropleth steps for Equal Population metric for a population deviation of 0
- Fixes computation of EqualPopulation metric for evaluate mode for a population deviation of 0
- Distinguishes between the unique cases of a whole integer `avgDistrictPopulation`, and an average population that contains a fraction

### Checklist

### Demo

Whole integer avgPopulation

image

image

Non-integer avgPopulation

image

image

## Testing Instructions

- Create a map in PA with an even number of districts (e.g., 4), so the average population contains a fractional remainder
- Add a district that exactly matches the target population, or target population + 1 - expect it to be treated as equal population
- Add a district that is -1 the target population, expect to be treated as invalid

- Create a map in VA with an even number of districts (e.g., 4), so the average population is a whole integer
- Add a district that exactly matches the target population - expect it to be treated as equal population
- Add a district that not exactly equal to the target population, expect to be treated as invalid

Closes #801

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Don't show limit to county option in read-only mode [#777](https://github.com/PublicMapping/districtbuilder/pull/777)
 - Include unassigned district when switching using keyboard shortcuts [#779](https://github.com/PublicMapping/districtbuilder/pull/779)
 - Fix calculation of average compactness [#778](https://github.com/PublicMapping/districtbuilder/pull/778)
+- Fix computation of equal population flags for population deviation of zero [#806](https://github.com/PublicMapping/districtbuilder/pull/806)
 
 ## [1.5.0] - 2021-05-13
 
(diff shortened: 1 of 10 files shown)


maurizi:
@pcaisse I did a bunch of rework on this, could you take a quick look?

pcaisse (approved):

PR #818, opened 2021-06-14
Add support for multiple elections

### LINKED ISSUE #807: Align bar charts in tooltip
The bars in the tooltip would be easier to read if they were left aligned across Race/Politics bars. I recommend abbreviating the political labels to `Rep.`, `Dem.`, `Other` as part of this.

Screen Shot 2021-06-04 at 1 23 23 PM

### LINKED ISSUE #808: Support data for multiple elections on project screen
Thus far we have supported regions with only one elections worth of data. Going forward, we'll be supporting the _possibility_ of having multiple elections (likely 2016 and 2020 presidential, but possible 2016 presidential / 2018 midterms), showing a calculated PVI value in the sidebar, and allowing control of which election is shown in the map 

AC:
 - If data is available for multiple elections, we should calculate and show the [PVI for the district](https://en.wikipedia.org/wiki/Cook_Partisan_Voting_Index)
 - When hovering over the sidebar voting value, the tooltip should _always_ show data for all elections
 - There is a new control in the ⚙ menu that allows picking between elections, which will _only_ effect the data shown in the map tooltip
 - A keyboard shortcut is added to toggle through the options of the above control

maurizi:
## Overview

This PR adds support for regions with election data for both the 2016 & 2020 presidential elections.

### Checklist

### Demo

Election year control for map data:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (4)
Map tooltip:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (5)
Sidebar tooltip:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (3)
Sidebar label tooltip:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (2)
New keyboard shortcut:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (1)
Toast for new keyboard shortcut:
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8

### Notes

We had originally written this in a more generic fashion, and I'm choosing in this PR to hardcode in support for the following three options instead, to make the code easier to work with:
 - Unspecified election, keys are `democrat` / `republican` / `other party` (what we currently have, I assume the election is 2016)
 - Both 2016 & 2020, keys are  `democrat16` / `republican16` / `other party16` /  `democrat20` / `republican20` / `other party20` (this is what the new IL data has)
 - Just 2016, but explicitly specified by the key name (We don't have any states like this yet, but I added support as I know some upcoming states will only have 2016 data and this should make the preprocessing easier)
 
 
 I played with putting this into the type signature for `IStaticMetadata.voting` (which currently allows any `string` keys), but backed away from it when it looked like a more extensive refactor than I wanted to get into.
 
 
## Testing Instructions

- Use the `load-dev-data` script or update your database to use the `IL` data in `s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z/`
  - Interactions on the map page should show election data, and the sidebar should show PVI
- Also ensure you have a state in your database with pre-existing election data, either `PA` or `VA`, to test for regressions
  - You should not be able to switch election years for a state with data for only 1 election
- For a state/region with no election data at all, there should also be no regressions

Closes #808 
Closes #807

### DIFF
--- a/src/client/components/MapSelectionOptionsFlyout.tsx
+++ b/src/client/components/MapSelectionOptionsFlyout.tsx
@@ -1,21 +1,31 @@
 /** @jsx jsx */
-import { jsx, Checkbox, Label } from "theme-ui";
+import { jsx, Checkbox, Label, Select } from "theme-ui";
 import { Button as MenuButton, Wrapper, Menu } from "react-aria-menubutton";
 import { style } from "./MenuButton.styles";
 import store from "../store";
-import { toggleLimitDrawingToWithinCounty } from "../actions/districtDrawing";
+import {
+  toggleLimitDrawingToWithinCounty,
+  ElectionYear,
+  setElectionYear
+} from "../actions/districtDrawing";
 import Tooltip from "./Tooltip";
 import Icon from "./Icon";
-
-type MenuKeys = "limit-drawing-to-within-county";
+import { IStaticMetadata } from "../../shared/entities";
 
 const MapSelectionOptionsFlyout = ({
   limitSelectionToCounty,
-  topGeoLevelName
+  metadata,
+  topGeoLevelName,
+  electionYear
 }: {
   readonly limitSelectionToCounty: boolean;
+  readonly metadata?: IStaticMetadata;
   readonly topGeoLevelName?: string;
+  readonly electionYear: ElectionYear;
 }) => {
+  const votingIds = metadata?.voting?.map(file => file.id) || [];
+  const hasMultipleElections =
+    votingIds.some(id => id.endsWith("16")) && votingIds.some(id => id.endsWith("20"));
   return (
     <Wrapper closeOnSelection={false}>
       <Tooltip content="Drawing options">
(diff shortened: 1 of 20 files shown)


src/client/components/MapSelectionOptionsFlyout.tsx
```diff
@@ -59,6 +69,28 @@ const MapSelectionOptionsFlyout = ({
...
+                sx={{ display: "inline-block", width: "auto", mb: 0, mr: 2 }}
+              >
+                Election data:
+              </Label>
+              <Select
+                id="election-dropdown"
+                value={electionYear}
+                onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+                  const year = e.currentTarget.value;
+                  (year === "16" || year === "20") && store.dispatch(setElectionYear(year));
```
maurizi, line 85: @pcaisse Do you know if there is something I can use here like `assertNever(...)` to get the compiler to guarantee I've checked all the possible options for `ElectionYear` here?

maurizi:
One thing I realized I missed here - Brian provided the new race columns as `native` and `pacific`, instead of `nativeAmerican` and `pacificIslander` - will push a commit to handle that shortly.

kevinearldenny:
LGTM 👍

kevinearldenny (approved):

maurizi:
Added support for new race columns in https://github.com/PublicMapping/districtbuilder/pull/818/commits/eb6eedbb48e7c917021f74fb76c79dff1ffa322c
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (6)
localhost_3003_projects_659b2a51-b89e-4c61-b505-19677ec659a8 (7)

PR #823, opened 2021-06-14
Don't allow keyboard shortcuts to mess with form elements

### LINKED ISSUE #792: Keyboard shortcuts should not fire while user renames map
While typing in the input field to rename a map, keyboard shortcuts continue to fire. For shortcuts, we should ignore input while a text input is focused.

Screen Shot 2021-05-26 at 1 36 13 PM

maurizi:
## Overview

Prevents keyboard shortcuts from firing when changing the project name, or using the keyboard to choose an option in the labels dropdown.

### Checklist

### Notes

I don't think we have any `textarea` elements on the project screen right now, but I figure future-proofing can't hurt - and I couldn't think of any other elements that could interfere w/ our keyboard shortcuts if they have focus.

## Testing Instructions

- On this branch, you should be able to type in the prject name input, or use the keyboard to choose an option in the labels dropdown or election year dropdown
- When not focused on those, keyboard shortcuts should continue to work as normal

Closes #792

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Improved padding in the sidebar district rows, which had become unbalanced [#795](https://github.com/PublicMapping/districtbuilder/pull/795)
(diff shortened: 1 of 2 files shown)


kevinearldenny (approved):
LGTM 👍

PR #827, opened 2021-06-21
Add expandable metrics viewer

### LINKED ISSUE #115: Expandable metrics viewer
[wireframe](https://app.abstract.com/projects/9f988240-1dad-11e8-9a40-cba930331af7/branches/master/commits/238ea75510f11bde6dcba5f841bd99aa326bd167/files/DCA38482-FC7F-4F48-BECD-44E07D9F7443/layers/C4E68FD7-B7C0-40D9-85B2-932B16FCE42B)
[wireframe 2](https://app.abstract.com/projects/9f988240-1dad-11e8-9a40-cba930331af7/branches/master/commits/09b392079f20d92a20fe38878b27420f0bd1e703/files/DCA38482-FC7F-4F48-BECD-44E07D9F7443/layers/917A65DA-42F8-45E1-AB95-F35E6E0A9E78?mode=design&sha=09b392079f20d92a20fe38878b27420f0bd1e703)
- Button on side bar expands to full screen
- Button on expanded version to contract back to sidebar
- Columns can be pinned to the sidebar to view
- Only a certain number of columns can be pinned
- Only certain columns can be pinned (some always show, District Color and Number, District population)

Note:
* Jeff would like to talk to the developer about this. The design is good, but implementation details need clarified.

kevinearldenny:
## Overview

Adds an expandable metrics viewer, with the ability to pin metrics so they are displayed in the default / collapsed metrics view project sidebar.

### Checklist

### Demo

image

### Notes

- Are there any additional metrics that should be included in the viewer? 
- I didn't implement a max pinned metrics length yet, I'm wondering what makes sense to set here @tgilcrest . Also not sure if it makes more sense to implement this client side or server side. Additionally, we will need to create some notifications / error handling around this?
- This will definitely need a style pass @jfrankl - namely, on the CSS animations and on the overlap between the 'Saved' message and the collapse icon

## Testing Instructions

- `./scripts/update`
- Open a project, and click the expand button to open the full screen metrics viewer
- Expect: all available metrics are displayed
- Click the pin icon to pin / unpin a metric, then click the collapse button
- Expect: pinned metrics are displayed

Closes #115


src/client/components/ProjectSidebar.tsx
```diff
@@ -170,11 +214,20 @@ const ProjectSidebar = ({
   const polLabel = multElections
     ? "Cook Partisan Voting Index (2016 / 2020)"
     : "Political Lean (2016)";
+  // Fix the formatting weirdness introduced by setting defaults
+  const pinnedMetrics: readonly string[] | undefined = project?.pinnedMetricFields.map(field =>
+    // eslint-disable-next-line
+    field.replace(/['"]+/g, "").replace(/[\[\]]/g, "")
+  );
```
kevinearldenny, line 221: @maurizi This seems ugly, I'm not sure how the formatting is getting all messed up for the `DEFAULT_PINNED_METRIC_FIELDS`. If we can reduce the need for this, then we can re-implement the `ProjectMetric` type

src/server/src/projects/entities/project.entity.ts
```diff
@@ -90,6 +93,14 @@ export class Project implements IProject {
   })
   populationDeviation: number;
 
+  @Column({
+    type: "simple-array",
+    name: "pinned_metric_fields",
+    default: DEFAULT_PINNED_METRIC_FIELDS
+  })
```
kevinearldenny, line 100: Is there a better way to set `DEFAULT_PINNED_METRIC_FIELDS` that won't make the formatting screwed up?

maurizi: You basically _never_ want to use the `simple-array` type - Postgres has built-in support for arrays, which works better.
This switches to the built-in array type:

```diff
diff --git a/src/client/components/ProjectSidebar.tsx b/src/client/components/ProjectSidebar.tsx
index 1f2ede7..924f08f 100644
--- a/src/client/components/ProjectSidebar.tsx
+++ b/src/client/components/ProjectSidebar.tsx
@@ -9,7 +9,8 @@ import {
   GeoUnits,
   IProject,
   IStaticMetadata,
-  LockedDistricts
+  LockedDistricts,
+  MetricField
 } from "../../shared/entities";
 
 import {
@@ -215,10 +216,7 @@ const ProjectSidebar = ({
     ? "Cook Partisan Voting Index (2016 / 2020)"
     : "Political Lean (2016)";
   // Fix the formatting weirdness introduced by setting defaults
-  const pinnedMetrics: readonly string[] | undefined = project?.pinnedMetricFields.map(field =>
-    // eslint-disable-next-line
-    field.replace(/['"]+/g, "").replace(/[\[\]]/g, "")
-  );
+  const pinnedMetrics: readonly string[] | undefined = project?.pinnedMetricFields;
   return (
     <Flex
       sx={expandedProjectMetrics ? style.sidebarExpanded : style.sidebar}
diff --git a/src/server/migrations/1624042176675-PinnedMetricFields.ts b/src/server/migrations/1624042176675-PinnedMetricFields.ts
deleted file mode 100644
index b06e18e..0000000
--- a/src/server/migrations/1624042176675-PinnedMetricFields.ts
+++ /dev/null
@@ -1,24 +0,0 @@
-import {MigrationInterface, QueryRunner} from "typeorm";
-
-export class PinnedMetricFields1624042176675 implements MigrationInterface {
-    name = 'PinnedMetricFields1624042176675'
-
-    public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "project" ADD "pinned_metric_fields" text NOT NULL DEFAULT '["population","populationDeviation","raceChart","pvi","compactness"]'`);
-        await queryRunner.query(`ALTER TABLE "project_template" ADD "pinned_metric_fields" text NOT NULL DEFAULT '["population","populationDeviation","raceChart","pvi","compactness"]'`);
-        await queryRunner.query(`COMMENT ON COLUMN "project"."population_deviation" IS NULL`);
-        await queryRunner.query(`ALTER TABLE "project" ALTER COLUMN "population_deviation" SET DEFAULT 5`);
-        await queryRunner.query(`COMMENT ON COLUMN "project_template"."population_deviation" IS NULL`);
-        await queryRunner.query(`ALTER TABLE "project_template" ALTER COLUMN "population_deviation" SET DEFAULT 5`);
-    }
-
-    public async down(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(`ALTER TABLE "project_template" ALTER COLUMN "population_deviation" SET DEFAULT '5'`);
-        await queryRunner.query(`COMMENT ON COLUMN "project_template"."population_deviation" IS NULL`);
-        await queryRunner.query(`ALTER TABLE "project" ALTER COLUMN "population_deviation" SET DEFAULT '5'`);
-        await queryRunner.query(`COMMENT ON COLUMN "project"."population_deviation" IS NULL`);
-        await queryRunner.query(`ALTER TABLE "project_template" DROP COLUMN "pinned_metric_fields"`);
-        await queryRunner.query(`ALTER TABLE "project" DROP COLUMN "pinned_metric_fields"`);
-    }
-
-}
diff --git a/src/server/migrations/1624309135081-PinnedMetricFields.ts b/src/server/migrations/1624309135081-PinnedMetricFields.ts
new file mode 100644
index 0000000..73c873e
--- /dev/null
+++ b/src/server/migrations/1624309135081-PinnedMetricFields.ts
@@ -0,0 +1,16 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class PinnedMetricFields1624309135081 implements MigrationInterface {
+    name = 'PinnedMetricFields1624309135081'
+
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "project" ADD "pinned_metric_fields" character varying array NOT NULL DEFAULT ARRAY['population','populationDeviation','raceChart','pvi','compactness']`);
+        await queryRunner.query(`ALTER TABLE "project_template" ADD "pinned_metric_fields" character varying array NOT NULL DEFAULT ARRAY['population','populationDeviation','raceChart','pvi','compactness']`);
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE "project_template" DROP COLUMN "pinned_metric_fields"`);
+        await queryRunner.query(`ALTER TABLE "project" DROP COLUMN "pinned_metric_fields"`);
+    }
+
+}
diff --git a/src/server/src/project-templates/entities/project-template.entity.ts b/src/server/src/project-templates/entities/project-template.entity.ts
index dec6022..ad3cd68 100644
--- a/src/server/src/project-templates/entities/project-template.entity.ts
+++ b/src/server/src/project-templates/entities/project-template.entity.ts
@@ -63,14 +63,15 @@ export class ProjectTemplate implements IProjectTemplateWithProjects {
   @Column({
     type: "double precision",
     name: "population_deviation",
-    default: () => `${DEFAULT_POPULATION_DEVIATION}`
+    default: DEFAULT_POPULATION_DEVIATION
   })
   populationDeviation: number;
 
   @Column({
-    type: "simple-array",
+    type: "character varying",
+    array: true,
     name: "pinned_metric_fields",
-    default: DEFAULT_PINNED_METRIC_FIELDS
+    default: () => `ARRAY[${DEFAULT_PINNED_METRIC_FIELDS.map(c => `'${c}'`).join(",")}]`
   })
   pinnedMetricFields: MetricField[];
 
diff --git a/src/server/src/projects/entities/project.entity.ts b/src/server/src/projects/entities/project.entity.ts
index 7d8dd58..a2f0294 100644
--- a/src/server/src/projects/entities/project.entity.ts
+++ b/src/server/src/projects/entities/project.entity.ts
@@ -2,7 +2,12 @@ import { FeatureCollection, MultiPolygon } from "geojson";
 import { Column, Entity, JoinColumn, ManyToOne, PrimaryGeneratedColumn } from "typeorm";
 
 import { ProjectVisibility } from "../../../../shared/constants";
-import { DistrictProperties, DistrictsDefinition, IProject } from "../../../../shared/entities";
+import {
+  DistrictProperties,
+  DistrictsDefinition,
+  IProject,
+  MetricField
+} from "../../../../shared/entities";
 import { RegionConfig } from "../../region-configs/entities/region-config.entity";
 import { Chamber } from "../../chambers/entities/chamber.entity";
 import { User } from "../../users/entities/user.entity";
@@ -89,17 +94,17 @@ export class Project implements IProject {
   @Column({
     type: "double precision",
     name: "population_deviation",
-    default: () => `${DEFAULT_POPULATION_DEVIATION}`
+    default: DEFAULT_POPULATION_DEVIATION
   })
   populationDeviation: number;
 
   @Column({
-    type: "simple-array",
+    type: "character varying",
+    array: true,
     name: "pinned_metric_fields",
-    default: DEFAULT_PINNED_METRIC_FIELDS
+    default: () => `ARRAY[${DEFAULT_PINNED_METRIC_FIELDS.map(c => `'${c}'`).join(",")}]`
   })
-  pinnedMetricFields: string[];
-  // pinnedMetricFields: MetricField[];
+  pinnedMetricFields: MetricField[];
 
   // Strips out data that we don't want to have available in the read-only view in the UI
   getReadOnlyView(): Project {
diff --git a/src/server/src/projects/entities/update-project.dto.ts b/src/server/src/projects/entities/update-project.dto.ts
index bd8ea4f..205c689 100644
--- a/src/server/src/projects/entities/update-project.dto.ts
+++ b/src/server/src/projects/entities/update-project.dto.ts
@@ -11,7 +11,7 @@ import {
 } from "class-validator";
 
 import { ProjectVisibility } from "../../../../shared/constants";
-import { DistrictsDefinition, UpdateProjectData } from "../../../../shared/entities";
+import { DistrictsDefinition, UpdateProjectData, MetricField } from "../../../../shared/entities";
 
 export class UpdateProjectDto implements UpdateProjectData {
   @IsNotEmpty({ message: "Please enter a name for your project" })
@@ -37,7 +37,7 @@ export class UpdateProjectDto implements UpdateProjectData {
   @IsOptional()
   readonly visibility: ProjectVisibility;
   @IsOptional()
-  readonly pinnedMetricFields: string[];
+  readonly pinnedMetricFields: MetricField[];
   @IsBoolean()
   @IsOptional()
   readonly archived: boolean;
```
```

File: src/client/components/ProjectSidebarHeader.tsx [hunk 3]
```diff
@@ -45,6 +60,14 @@ const ProjectSidebarHeader = ({
...
           Districts
         </Heading>
+        <Box sx={style.expandedToggle}>
+          <Button
+            sx={style.expandButton}
+            onClick={() => store.dispatch(toggleExpandedMetrics(!expandedProjectMetrics))}
+          >
+            {expandedProjectMetrics ? <Icon name="compress" /> : <Icon name="expand" />}
+          </Button>
+        </Box>
```
maurizi, line 70: To clean up the styling here:
 - Moving this element to after the spinner/saved and getting rid of absolute will position it correctly
 - Adds some padding to make up for the `right: 5px;` you had when you absolutely positioned it
```
diff --git a/src/client/components/ProjectSidebarHeader.tsx b/src/client/components/ProjectSidebarHeader.tsx
index 0e01e13..d4574d5 100644
--- a/src/client/components/ProjectSidebarHeader.tsx
+++ b/src/client/components/ProjectSidebarHeader.tsx
@@ -28,8 +28,7 @@ const style: ThemeUIStyleObject = {
     m: 0
   },
   expandedToggle: {
-    position: "absolute",
-    right: "5px"
+    p: 1
   },
   expandButton: {
     variant: "buttons.icon",
@@ -60,14 +59,6 @@ const ProjectSidebarHeader = ({
         <Heading as="h2" sx={{ variant: "text.h4", m: "0" }}>
           Districts
         </Heading>
-        <Box sx={style.expandedToggle}>
-          <Button
-            sx={style.expandButton}
-            onClick={() => store.dispatch(toggleExpandedMetrics(!expandedProjectMetrics))}
-          >
-            {expandedProjectMetrics ? <Icon name="compress" /> : <Icon name="expand" />}
-          </Button>
-        </Box>
       </Flex>
       {isLoading || saving === "saving" ? (
         <Flex sx={{ alignItems: "center", justifyContent: "center" }}>
@@ -120,6 +111,14 @@ const ProjectSidebarHeader = ({
           </Flex>
         </Tooltip>
       ) : null}
+      <Box sx={style.expandedToggle}>
+        <Button
+          sx={style.expandButton}
+          onClick={() => store.dispatch(toggleExpandedMetrics(!expandedProjectMetrics))}
+        >
+          {expandedProjectMetrics ? <Icon name="compress" /> : <Icon name="expand" />}
+        </Button>
+      </Box>
     </Flex>
   );
 };
```

maurizi (changes requested):
Left a few comments here - additionally, this isn't working correctly for regions w/ our old race breakdown, it only works w/ the new breakdown (i.e. IL) for me.

maurizi:
> Are there any additional metrics that should be included in the viewer?

From the wireframes, we're missing
 - Dem / Rep / Oth voting info (I think we want potentially _6_ columns here: Dem '16, Rep '16, Oth '16, Dem '20, Rep '20, Oth '20 - if we have the data available)
 - Target Population
 - Majority Race
 - Contiguous
 - Competitive
image

If you want to skip "Competitive" and "Majority Race" for now that's fine, since we don't show those anywhere else (please open a followup issue if you do).

For the others we should be able to add the column pretty easily - we already show it elsewhere

kevinearldenny:
A few of these are already included in existing columns:
> 
> * Target Population

Included as part of the population deviation column, with the check or X next to the population deviation

> * Contiguous

Included as part of compactness - compactness is only displayed if the district is contiguous. If the district is not contiguous, a flag is displayed here indicating that it is not contiguous.


> If you want to skip "Competitive" and "Majority Race" for now that's fine, since we don't show those anywhere else (please open a followup issue if you do).

Yeah, both of these seem like they need to be defined first before we display them. 

> Dem / Rep / Oth voting info (I think we want potentially 6 columns here: Dem '16, Rep '16, Oth '16, Dem '20, Rep '20, Oth '20 - if we have the data available)

I will add these in 👍

src/client/components/ProjectSidebar.tsx
```diff
@@ -160,21 +237,165 @@ const ProjectSidebar = ({
...
+  /// This is a little wonky to preserve the column ordering where PVI is displayed before compactness
+  const metricHeaders: readonly MetricHeader[] = [
+    ...allDemographicHeaders,
+    ...electionMetricHeaders,
+    {
+      metric: "compactness",
+      text: "Comp.",
+      tooltip: "Compactness score (Polsby-Popper)"
+    }
+  ];
```
kevinearldenny, line 387: @maurizi Can you think of a better way to handle this? I imagine it is important that we preserve the existing column ordering here where compactness is displayed all the way to the right

maurizi (changes requested):
Just a few small suggestions - 1 I missed in the first review plus a few suggestions to the most recent changes.

src/client/components/ProjectSidebar.tsx
```diff
@@ -385,95 +600,271 @@ const SidebarRow = memo(
...
+            <Styled.td sx={{ ...style.td, ...style.number, ...{ color: textColor } }}>
+              <span>{demographics.pacific.toLocaleString()}</span>
+            </Styled.td>
+          )}
+        {otherDemographics &&
+          (pinnedMetricFields.includes("otherPopulation") || expandedProjectMetrics) && (
+            <Styled.td sx={{ ...style.td, ...style.number, ...{ color: textColor } }}>
+              <span>{demographics.other.toLocaleString()}</span>
+            </Styled.td>
+          )}
```
maurizi, line 687: Consider adding a loop here to DRY things up:
e.g.
```tsx
{
  [
    {id: "white", field: "whitePopulation"},
    ...
  ].map(({ field, key }) =>pinnedMetricFields.includes(field) || expandedProjectMetrics) && (
    ...
  )
)}

File: src/client/components/ProjectSidebar.tsx [hunk 6]
```diff
@@ -385,95 +600,271 @@ const SidebarRow = memo(
...
+                        vote totals
+                      </em>
+                    )
+                  }
+                >
+                  <span>{voting["other party20"].toLocaleString()}</span>
+                </Tooltip>
+              </Styled.td>
+            )
+          : null}
```
maurizi, line 834: Similar to the above suggestion, you can DRY this up a bit:

```tsx
{
  voting && [
    {id: "democrat16", field: "dem16"},
    ...
  ].map(({ field, key }) => key in voting && pinnedMetricFields.includes(field) || expandedProjectMetrics) && (
    <Styled.td>
      ...
    </Styled.td>
  )
)}
```

maurizi (changes requested):
Sorry for _just_ realizing this now - in the wireframes, the sidebar values are _percentages_, not total votes/population/etc. - we should match the wireframes unless you have direction from @jfrankl to do otherwise.

Also, since you'll likely have to go in and make changes to the code I left some suggestions on how to clean it up a bit while you're there.

src/client/components/ProjectSidebar.tsx
kevinearldenny: @maurizi I thought about this, but seems like this may be more trouble than it's worth considering that the displays are different for different types of metrics (demographics, voting) and there are additional nuances here (e.g., `!otherDemographics` for `pacific` and `native`)

src/client/components/ProjectSidebar.tsx
kevinearldenny: Similar to the above comment, since pinned metrics are all included in the same array this seems like this may be more trouble that it's worth.

maurizi:
This is really close, but I think it's important that we stay consistent in how we round these numbers here & in the tooltip:

Difference in rounding for '20 numbers:
image
Difference in rounding for Native:
image

kevinearldenny:
Okay, the percentage formatting bug should be fixed now - @pcaisse can you take a look when you have a minute?

src/client/components/ProjectSidebar.tsx
```diff
@@ -26,17 +28,24 @@ import {
 import {
   areAnyGeoUnitsSelected,
   assertNever,
+  getPartyColor,
```
ddohler, line 31: I'm getting linter warnings that this and `calculatePVI` are unused, should they be removed?

src/client/components/ProjectSidebar.tsx
```diff
@@ -135,9 +156,69 @@ const style: ThemeUIStyleObject = {
...
+  readonly pinnedMetrics: readonly MetricField[];
+}) => {
+  const metricIsPinned = pinnedMetrics.includes(metric);
+  const otherMetrics = pinnedMetrics.filter(m => m !== metric);
+  return (
+    <React.Fragment>
+      {metricIsPinned ? (
+        <Button
+          sx={style.pinButton}
+          onClick={() => store.dispatch(updatePinnedMetrics([...otherMetrics]))}
```
ddohler, line 181: I think this button and the one below need to be disabled if saving is in progress so that the click handler can only be triggered if saving has completed. Currently, if I un-pin all the columns and then click on each pin button as quickly as I can, I can trigger a race condition where some columns end up un-pinned even though I've clicked them.

Starting this way:
Screen Shot 2021-07-12 at 11 41 47 AM

And then clicking every pin quickly in succession (without waiting for the previous one to finish saving before clicking the next), I get this:
Screen Shot 2021-07-12 at 11 42 12 AM

However, I would expect all pins to be black in this case.

What's happening is that if I can manage to fire off two PATCH requests in quick enough succession such that the second one is made before the first one completes, the first patch payload will look something like
```
{"pinnedMetricFields":["pvi","population","populationDeviation","raceChart"]}
```

and then the second request payload will be:
```
{"pinnedMetricFields":["pvi","population","populationDeviation","whitePopulation"]}
```
You can see that `"raceChart"` is missing from the second request payload, so the server interprets the second request as un-pinning that column. This will cause a race condition: whichever one arrives at the server last will win, and one of the pins will be momentarily pinned and then immediately unpinned when the second response comes back.

I think disabling the button while saving would be the easiest way to address this, but another approach might be to express the pinned fields using an Object rather than an array. Then it would be possible to make a PATCH request that looked more like `{ "pinnedMetricFields":{ "population": true }}` or `{ "pinnedMetricFields": { "population": false }}`. However, I think that would be a significant rework of what's here, so I suspect that disabling the button will be easier.

ddohler (changes requested):
I did not review this thoroughly because it has already been looked over several times, but I did confirm that the tooltip rounding matched the column rounding in every instance I tested.

I found a couple more things that I think should be addressed before merging, but I'm expecting they'll be minor.

src/client/components/ProjectSidebar.tsx
kevinearldenny: Yep - looks like these were remnants that got cleaned up in the rebase

ddohler (approved):
Works great! I'm not able to fake out the pinning buttons any more, it just doesn't let me click if the previous pin request hasn't completed.

Issue #828, opened 2021-06-22
Add management command to update 'districts' column for a given region


maurizi:
When we added voting data for PA & VA, we failed to go back and update the cached `districts` GeoJSON column to include the newly added voting information.

This hasn't been an issue so far on the Project Screen, due to an oversight introduced in https://github.com/PublicMapping/districtbuilder/pull/636 that meant we weren't using the `districts` column when using the `export/geojson/` endpoint, but instead always recalculating it from the TopoJSON 😭 

However, we'd now like to make use of the `districts` column to be able to handle archived regions, and so we'll need to handle this case.

# AC
 - A management command is added that will iterate through all projects for a given region, calculate it's districts GeoJSON, and update the database
 - A method for running this command in production is devised (coordinate with Ops)
 
 # Notes
  - We haven't had to run Node.js-based commands like this, that need access to the TopoJSON, before. Limiting ourselves to a single region at a time should make it feasible to create a new Fargate task, as we do when running migrations - handling all regions at once would incur excessive startup & memory requirements.
 
 - If we _don't_ do this, we could handle it client-side by simply not showing the voting columns if we don't have data for them for a given project (e.g. because it was last editing _before_ we added voting data for that region)

maurizi:
Superseded by #884

PR #831, opened 2021-06-22
Add support for archived regions

### LINKED ISSUE #814: Allow marking 2010 regions as archived
In order to reduce resource consumption, we will stop allowing editing of 2010 census geographies when we make 2020 geographies available. To support that, we're going to make it possible to have a region be in a read-only mode, which will prevent editing of any maps for that region and will cause the server to not download the TopoJSON for the region.

AC:
 - Add a new `archived` field to `RegionConfig`
 - Archived regions should:
   - Be allowed to coexist alongside a non-archived region for the same state (we'll need to update our uniqueness DB constraint)
   - The `TopologyService` should not download the topojson for the region
   - API calls that would require the topology should return an error
   - Projects for that region are in read-only mode, even you are looking at your own project
- Projects for archived regions should still appear on the Home screen / Organization featured projects / Maps screen

maurizi:
## Overview

Adds support for archiving regions.

Archived regions:
 - Don't have topojson data kept in memory (so they are not resource intensive)
 - Projects for archived regions can't be edited

### Checklist

### Demo

localhost_3003_projects_3e64db60-0e9a-4c7f-8785-5771716eb39d (1)

### Notes

 - Downloading the full TopoJSON only to throw away the geometries isn't the most efficient way to go about things - if we find the startup times are not acceptable on production, a simple optimization would be to create new static files from our existing regions that contain the feature properties in the TopoJSON files as well as a pre-computed `HierarchyDefinition`.

## Testing Instructions

- `scripts/update`
- Set one of your `region_config` rows to be `archived, and restart `scripts/server`
   - existing projects for that region should still be viewable, but no longer editable
   - all evaluate mode subpages should function correctly
   - all exports should function correctly
   - there should be no regressions for projects for a non-archived region

Closes #814

### DIFF
--- a/src/client/components/ProjectSidebar.tsx
+++ b/src/client/components/ProjectSidebar.tsx
@@ -166,6 +166,7 @@ const ProjectSidebar = ({
   readonly saving: SavingState;
   readonly isReadOnly: boolean;
 } & LoadingProps) => {
+  const hasElectionData = Object.keys(geojson?.features[0].properties.voting || {}).length > 0;
   const multElections = hasMultipleElections(staticMetadata);
   const polLabel = multElections
     ? "Cook Partisan Voting Index (2016 / 2020)"
@@ -201,7 +202,7 @@ const ProjectSidebar = ({
                   <span>Race</span>
                 </Tooltip>
               </Styled.th>
-              {staticMetadata?.voting && (
+              {hasElectionData && (
                 <Styled.th sx={{ ...style.th, ...style.number }}>
                   <Tooltip content={polLabel}>
                     <span>{multElections ? "PVI" : "Pol."}</span>
--- a/src/server/src/districts/entities/geo-unit-properties.entity.ts
+++ b/src/server/src/districts/entities/geo-unit-properties.entity.ts
@@ -6,7 +6,7 @@ import {
 } from "../../../../shared/entities";
 import { GeoUnitTopology } from "./geo-unit-topology.entity";
 
-// A slimmed down version of GeoUniTopology without the geometries, only the feature properties
+// A slimmed down version of GeoUnitTopology without the geometries, only the feature properties
 export class GeoUnitProperties {
   public readonly hierarchy: ReadonlyArray<GeoUnitHierarchy>;
   static fromTopology(geoUnitTopology: GeoUnitTopology): GeoUnitProperties {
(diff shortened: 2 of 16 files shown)


src/client/components/ProjectSidebar.tsx
```diff
@@ -166,6 +166,7 @@ const ProjectSidebar = ({
   readonly saving: SavingState;
   readonly isReadOnly: boolean;
 } & LoadingProps) => {
+  const hasElectionData = Object.keys(geojson?.features[0].properties.voting || {}).length > 0;
```
maurizi, line 169: This is working around not having implemented #828

src/server/src/districts/services/topology.service.ts
```diff
@@ -82,14 +101,18 @@ export class TopologyService {
...
-        return new GeoUnitTopology(
+        const geoUnitTopology = new GeoUnitTopology(
           topology,
           { groups: geoLevelHierarchy.slice().reverse() },
           staticMetadata,
           demographics,
           voting,
           geoLevels
         );
+        this.logger.debug(`finished calculating computed data for s3URI ${s3URI}`);
```
maurizi, line 112: Note to self: remove this
```suggestion
        this.logger.debug(`finished calculating computed data for s3URI ${s3URI}`);
```

kevinearldenny (approved):
LGTM 👍

src/server/src/region-configs/entities/region-config.entity.ts
```diff
@@ -46,6 +46,12 @@ export class RegionConfig implements IRegionConfig {
   @Column({ type: "timestamp with time zone", default: () => "NOW()" })
   version: Date;
 
+  // Hidden regions have data loaded and can be used to edit projects,
+  // but do not appear in the list of regions when creating a new project
   @Column({ type: "boolean", default: false })
   hidden: boolean;
+
+  // Archived regions are hidden, and also do not have data loaded and so their projects cannot be editted
```
maurizi, line 54:
```suggestion
  // Archived regions are hidden, and also do not have data loaded and so their projects cannot be edited
```

src/server/src/districts/entities/geo-unit-properties.entity.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  GeoUnitDefinition,
+  HierarchyDefinition,
+  GeoUnitHierarchy,
+  IStaticMetadata
+} from "../../../../shared/entities";
+import { GeoUnitTopology } from "./geo-unit-topology.entity";
+
+// A slimmed down version of GeoUniTopology without the geometries, only the feature properties
```
maurizi, line 9:
```suggestion
// A slimmed down version of GeoUnitTopology without the geometries, only the feature properties
```

Issue #833, opened 2021-06-23
Add state dropdown to Community Maps page


maurizi:
AC
 - This should show the same list of states as the "Add Project", as well as an "All States/Regions" option
 - We should update the page URL as the state is selected, and preselect a state based on the page URL
 - We should display appropriate text when no maps match the state filter

PR #836, opened 2021-06-24
Add completed filter to community maps page

### LINKED ISSUE #826: Filter maps on /maps page to exclude uncompleted maps
# AC 
 - An optional `completed` flag is added to the [global projects controller](https://github.com/PublicMapping/districtbuilder/blob/c1a5c7f172dfbd83626a6698d4c3d9dba79d8d88/src/server/src/projects/controllers/globalProjects.controller.ts#L12-L18) to only include maps which do not have any population not assigned to a district
- Use this flag on the Community Maps page to only show completed projects

# Notes 
We should be able to check for this in SQL by looking at the `population` of the first district in the saved geojson `districts` column.

The API should retain the ability to show all projects, including uncompleted ones, even though we don't plan to use this API w/o supplying the `completed` flag (to support external users of the API)

maurizi:
## Overview

Filters out maps on the Community Maps page that have population left in the unassigned district. This is intended to keep uninteresting maps filtered out, making the page more useful.

### Checklist

### Demo

Before:
localhost_3003_maps

After:
localhost_3003_maps (1)

### Notes

Note that by default the API will still return uncompleted maps, unless `completed=true` is passed as a query parameter.

## Testing Instructions

- `scripts/server`
- go to the community maps page

Closes #826

### DIFF
--- a/src/server/src/projects/controllers/globalProjects.controller.ts
+++ b/src/server/src/projects/controllers/globalProjects.controller.ts
@@ -1,5 +1,12 @@
 // eslint-disable-next-line
-import { Controller, Get, ParseIntPipe, Query } from "@nestjs/common";
+import {
+  Controller,
+  Get,
+  ParseIntPipe,
+  Query,
+  ParseBoolPipe,
+  DefaultValuePipe
+} from "@nestjs/common";
 import { Project } from "../entities/project.entity";
 import { ProjectsService } from "../services/projects.service";
 
@@ -12,8 +19,9 @@ export class GlobalProjectsController {
   @Get()
   async getAllGlobalProjects(
     @Query("page", ParseIntPipe) page = 1,
-    @Query("limit", ParseIntPipe) limit = 10
+    @Query("limit", ParseIntPipe) limit = 10,
+    @Query("completed", new DefaultValuePipe(false), ParseBoolPipe) completed = false
   ): Promise<Pagination<Project>> {
-    return this.service.findAllPublishedProjectsPaginated({ page, limit });
+    return this.service.findAllPublishedProjectsPaginated({ page, limit, completed });
   }
 }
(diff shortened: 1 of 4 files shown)


kevinearldenny (changes requested):
The error text here could be improved a bit - probably worth noting that only completed maps show up here. Although I imagine we will always have a few completed maps in production, this might come in handy one we have filtering enabled:

image

src/server/src/projects/controllers/globalProjects.controller.ts
```diff
@@ -1,5 +1,12 @@
 // eslint-disable-next-line
-import { Controller, Get, ParseIntPipe, Query } from "@nestjs/common";
+import {
+  Controller,
+  Get,
+  ParseIntPipe,
+  Query,
+  ParseBoolPipe,
+  DefaultValuePipe
```
kevinearldenny, line 8: These are returning a lint error for some reason saying that they are not used (even though they are). Any idea what's up here? Might be a good idea to throw an `eslint-disable` here to prevent the build from failing

maurizi: Indeed, there was an `eslint-disable-next-line` comment, but it stopped working when `prettier` reformatted this import to be over multiple lines 😭 

Fixed in https://github.com/PublicMapping/districtbuilder/pull/836/commits/3f48a81084337bf29d4c0c9e3898ec0c4304e0e5

maurizi:
> The error text here could be improved a bit - probably worth noting that only completed maps show up here. Although I imagine we will always have a few completed maps in production, this might come in handy one we have filtering enabled:

We already have at least 1 completed map in production, so we'll never see this message except w/ our dev server. I added changing this text to the acceptance criteria for https://github.com/PublicMapping/districtbuilder/issues/833 so we remember to take care of it as part of that issue (where it will be likely to occur, as you noted).

kevinearldenny (approved):
LGTM 👍

PR #840, opened 2021-06-29
Add voting info to labels selector

### LINKED ISSUE #731: Add voting info to Labels selector
It should be relatively easy to add support for voting totals by party here:
image

kevinearldenny:
## Overview

Adds voting info to labels selector, which display vote totals for geounits when selected.

### Checklist

### Demo

image

### Notes

We may want to handle the formatting of the field names here - I wasn't able to devise an easy solution for this, and I guess the labels aren't _terrible_ currently (Democrat20, Other party20). I was thinking that it might make sense to change these labels further upstream in the geojson processing, but that likely requires some input from other stakeholders first.

## Testing Instructions

- Go to a state with voting data and look in the labels dropdown - expect to see voting fields displayed
- Click a voting field, expect to see voting totals displayed

Closes #731

### DIFF
--- a/src/client/components/MapHeader.tsx
+++ b/src/client/components/MapHeader.tsx
@@ -145,10 +145,34 @@ const MapHeader = ({
   const topGeoLevelName = metadata
     ? metadata.geoLevelHierarchy[metadata.geoLevelHierarchy.length - 1].id
     : undefined;
-  const labelOptions = metadata
-    ? metadata.demographics.map(val => (
+  const labelFields =
+    metadata && metadata.demographics
+      ? [
+          ...metadata.demographics.map(file => {
+            return {
+              id: file.id,
+              label: file.id
+            };
+          }),
+          ...(metadata.voting
+            ? metadata.voting.map(file => {
+                return {
+                  id: file.id,
+                  // Fix suffix on voting IDs if needed
+                  label:
+                    file.id.endsWith("16") || file.id.endsWith("20")
+                      ? file.id.slice(0, -2) + " '" + file.id.slice(-2)
+                      : file.id
+                };
+              })
+            : [])
+        ]
+      : undefined;
+
+  const labelOptions = labelFields
+    ? labelFields.map(val => (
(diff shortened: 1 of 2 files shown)


src/client/components/MapHeader.tsx
```diff
@@ -145,8 +145,14 @@ const MapHeader = ({
...
+  const labelFields =
+    metadata && metadata.demographics
+      ? metadata.voting
+        ? [...metadata.demographics, ...metadata.voting]
+        : [...metadata.demographics]
+      : undefined;
+  const labelOptions = labelFields
+    ? labelFields.map(val => (
         <option key={val.id} value={val.id}>
           {capitalizeFirstLetter(val.id)}
```
maurizi, line 157:
```suggestion
  const labelFields =
    metadata && metadata.demographics
      ? metadata.voting
        ? [
           ...metadata.demographics.map(file => { id: file.id, label: file.id }),
           ...metadata.voting.map(file => {
             id: file.id,
             // Fix suffix on voting IDs if needed
             label: file.id.endsWith("16") || file.id.endsWith("20") ? file.id.slice(0, -2) + " '" + file.id.slice(-2) : file.id
           })]
        : [...metadata.demographics.map(file => { id: file.id, label: file.id })]
      : undefined;
  const labelOptions = labelFields
    ? labelFields.map(val => (
        <option key={val.id} value={val.id}>
          {capitalizeFirstLetter(val.label)}
```

maurizi (changes requested):
>  I was thinking that it might make sense to change these labels further upstream in the geojson processing, but that likely requires some input from other stakeholders first.

We've made quite a few code changes that make use of the `16` / `20` suffix, so I don't think changing these upstream now would a good use of ours or Brian's time.

Left a (rough & untested) suggestion of how we might coax these IDs into appropriate labels

kevinearldenny:
Okay, updated labels are now implemented: 
image

pcaisse (approved):
:tada:

PR #842, opened 2021-06-30
Add user ID to server-side rollbar logging

### LINKED ISSUE #789: Add more information to server-side rollbar error reporting
We should be able to capture this info as part of our error reporting but are not right now:
 - User ID
 - User IP address

kevinearldenny:
## Overview

Adds user_id to server-side rollbar implementation, enabling for grouping of occurrences by user

### Checklist

### Demo

image

### Notes

It looks like there is already some sort of ip_address field being captured (as can be seen in the above screenshot). Is this value correct? It is strange that it doesn't seem to ever load if you go to the IP addresses tab, though:

image

## Testing Instructions

- Find a way to make the server return a 500 error (I had to introduce a `throw new Error("foo");` to test this in development)
- Expect: rollbar error is logged and user ID is captured (if logged in)

Closes #789

### DIFF
--- a/src/server/src/rollbar/rollbar-exception.filter.ts
+++ b/src/server/src/rollbar/rollbar-exception.filter.ts
@@ -6,9 +6,15 @@ import {
   ServiceUnavailableException,
   UnauthorizedException
 } from "@nestjs/common";
-import { BaseExceptionFilter } from "@nestjs/core";
 import { Request } from "express";
+import { BaseExceptionFilter } from "@nestjs/core";
 import { RollbarService } from "./rollbar.service";
+
+export interface IGetUserAuthInfoRequest extends Request {
+  user?: {
+    id: number;
+  };
+}
 
 function isWhitelisted(exception: HttpException) {
   // Note that we don't need to whitelist BadRequestException as it has it's
@@ -18,6 +24,18 @@ function isWhitelisted(exception: HttpException) {
     exception instanceof ServiceUnavailableException ||
     exception instanceof UnauthorizedException
   );
+}
+
+function parseIp(req: IGetUserAuthInfoRequest): string | undefined {
+  if (req.headers["x-forwarded-for"]) {
+    if (Array.isArray(req.headers["x-forwarded-for"])) {
+      return [...req.headers["x-forwarded-for"]].shift();
+    } else {
+      return req.headers["x-forwarded-for"]?.split(",").shift();
+    }
+  } else {
+    return req.socket?.remoteAddress;
+  }
 }
 
 @Catch()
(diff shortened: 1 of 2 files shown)


maurizi:
> It looks like there is already some sort of ip_address field being captured (as can be seen in the above screenshot). Is this value correct? It is strange that it doesn't seem to ever load if you go to the IP addresses tab, though:

This is likely the IP address of the _load balancer_ - we need to use the `X-Forwarded-For` header to get the IP address of the original client (the load balancer will add this header). You can test this by making sure you see *your own* public IP address in Rollbar when you trigger an error for testing.

kevinearldenny:
Hmm, so the recent changes are now returning:

- The `person` field filled in with user_id
image

- The `request.user_ip` field filled in with an IP address that appears to correspond with the IPv6 address that I see from https://whatismyipaddress.com/

However, there are a number of fields that are not being filled in that I'm a bit confused by - particularly, that these fields were being filled in [c0094b1e06c6134c5f55ffc6205a8aad437f4280](https://github.com/PublicMapping/districtbuilder/pull/842/commits/c0094b1e06c6134c5f55ffc6205a8aad437f4280)

image

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -345,7 +345,8 @@ export class ProjectsController implements CrudController<Project> {
       relations: ["regionConfig"]
     });
     if (!project) {
-      throw new NotFoundException(`Project ${id} not found`);
+      throw new Error("foo");
```
kevinearldenny, line 348: This is for testing, and should be removed before merge

src/server/src/rollbar/rollbar-exception.filter.ts
```diff
@@ -33,9 +51,12 @@ export class RollbarExceptionFilter extends BaseExceptionFilter {
       (exception instanceof Error && !(exception instanceof HttpException))
     ) {
       const ctx = host.switchToHttp();
-      const request = ctx.getRequest<Request>();
-
-      this.rollbar.error(exception, request);
+      const request = ctx.getRequest<IGetUserAuthInfoRequest>();
+      this.rollbar.error(exception, {
+        ...request,
+        user_id: request.user ? request.user.id : null,
```
ddohler, line 57:
```suggestion
        person: request.user ? { id: request.user.id } : {},
```

src/server/src/rollbar/rollbar-exception.filter.ts
```diff
@@ -20,6 +26,18 @@ function isWhitelisted(exception: HttpException) {
   );
 }
 
+function parseIp(req: IGetUserAuthInfoRequest): string | undefined {
+  if (req.headers["x-forwarded-for"]) {
+    if (Array.isArray(req.headers["x-forwarded-for"])) {
+      return [...req.headers["x-forwarded-for"]].shift();
```
maurizi, line 32: Is this different from the following?
```suggestion
      return req.headers["x-forwarded-for"][0];
```

maurizi (approved):
The code changes look reasonable to me, though I haven't tested them myself - but it looks like it's working correctly from the Rollbar logs.

src/server/src/rollbar/rollbar-exception.filter.ts
kevinearldenny: There is no guarantee that the index has at least one element, probably safer to use shift

maurizi: Playing around in the Chrome JS console shows no difference in that case:
```
>> a = []
⬅ Array []

>>a.shift()
⬅ undefined
>>a[0]
⬅ undefined

​```

Issue #843, opened 2021-07-07
Add shortText and longText for Competitiveness metric


kevinearldenny:
## Detailed Description
We need an explanatory `shortText` for the `ProjectEvaluateSummary` page as well as a `longText` to live on the Competitiveness detail page. There are a few dev cards related to this page in flight currently, but once we have this copy written it should be very easy to add it in to the application.

dmcglone:
Short text:
A competitiveness metric evaluates the plan based on the average partisan lean of each district.

Long text:
A competitiveness metric evaluates the plan based on the average partisan lean of each district, calculated using the Partisan Voting Index (PVI). A partisan lean of the district plan which deviates from the overall lean of the state can be indicative of gerrymandering.

maurizi:
This is done now

PR #844, opened 2021-07-07
Add histogram and text description to evaluate mode Competitiveness page

### LINKED ISSUE #837: Add histogram to Evaluate: Competitiveness
We want to include a histogram on the Evaluate: Competitiveness page, that shows how many districts fall into each category.

kevinearldenny:
## Overview

- Adds the `visx` visualization library to support rendering of charts
- Adds a histogram grouping a count of districts for each of the 5 political leanings buckets
- Adds the `shortText` and `longText` specified in #843 to the Competitiveness metric page

### Checklist

### Demo

image

### Notes

- I ran into some type ugliness around the visx scales, and it also seems like my current computation of the PVI buckets could likely be improved a little bit. I am curious if you have any better suggestions for the `computeRowBucket` function - it seems like we would probably want to refactor this into the existing `computeRowFill` function if there is not a better solution.

## Testing Instructions

- `./scripts/update`
- Open a completed project for a state with political data (e.g., IL)
- Go to Evaluate - Competitiveness, expect histogram to be displayed
- Select a different year and expect data in histogram to change

Closes #837 and #843


src/client/components/evaluate/detail/CompetitivenessChart.tsx
```diff
@@ -0,0 +1,140 @@
...
+        const bucket = pvi && computeRowBucket(pvi);
+        return (bucket && bucket.name) || "";
+      })
+      // @ts-ignore
+      // eslint-disable-next-line
+      .reduce((a, c) => ((a[c] = (a[c] || 0) + 1), a), Object.create(null));
+
+  const chartData: readonly PviBucket[] = getPviBuckets().map(bucket => {
+    return { ...bucket, count: buckets[bucket.name] || 0 };
+  });
```
kevinearldenny, line 33: My goal is to get an array that looks like:

```
[{ name: "D", count: 5, color: "#385d7a"}, ...]
```

I was running into some hangups with using `f.properties.pvi` and the ElectionYear selector where data was not displayed for the proper election, hence the re-computation here.

src/client/components/evaluate/detail/CompetitivenessChart.tsx
```diff
@@ -0,0 +1,140 @@
...
+            return buckets[i];
+          }
+        } else {
+          return buckets[i];
+        }
+      } else {
+        return buckets[i];
+      }
+    }
+  }
```
kevinearldenny, line 54: This is definitely cribbed from `computeRowFill`, but the structure of what is required from Mapbox makes me hesitant to consider changing the source `getPviSteps` array. I am also hesitant to change `computeRowFill` after we refactored the other evaluate mode views to use it.

src/client/components/evaluate/detail/CompetitivenessChart.tsx
```diff
@@ -0,0 +1,140 @@
...
+  //   accessor: (d: PviBucket[]) => number | string
+  // ) => (data: PviBucket[]) => scale(accessor(data));
+
+  // @ts-ignore
+  // eslint-disable-next-line
+  const compose: any = (scale, accessor) => data => scale(accessor(data));
+  // @ts-ignore
+  const xPoint = compose(xScale, x);
+  // @ts-ignore
+  const yPoint = compose(yScale, y);
```
kevinearldenny, line 93: The commented out block at the top here was my attempt at deriving the types for the `compose` functions, but I didn't quite get it. Assistance would be appreciated here!

src/client/components/evaluate/detail/CompetitivenessChart.tsx
maurizi: This seems fine to me 👍

maurizi:
@kevinearldenny I pushed another commit - https://github.com/PublicMapping/districtbuilder/pull/844/commits/da8a5e7fb6f1894ec3fc465feb64890866c922af - to swap out the fixed size of the chart for one that relies on the containing element for size - which should be less brittle in the face of styling changes.

maurizi (approved):
👍

PR #845, opened 2021-07-08
Add populationDeviation and Chamber selection to Import Project screen

### LINKED ISSUE #628: Complete import plan form features
image

AC:
 - Can select a chamber associated with the detected region, in addition to existing custom district option
 - Can specify target population

kevinearldenny:
## Overview

- Add populationDeviation and Chamber selection to Import Project screen
- Create chambers in `./scripts/load-dev-data` so that chambers are available for PA and IL

### Checklist

### Demo

image

### Notes

- To test this, we need to have chambers in the DB - I thought we had these, but they seem to have been removed from `load-dev-data` at some point. I added back in SQL to create chambers for PA and IL

- I considered refactoring `CreateProjectScreen` and `ImportProjectScreen` to become one component, or to leverage shared functions across the two since they are now relatively similar - however, I decided against it with my rationale being mostly based around the fact that we are unlikely to use any of these components in other places, and we also don't currently have anything around Organization included as part of the Import workflow, which I think would be pretty difficult to implement.

## Testing Instructions

- `./scripts/load-dev-data`
- Import Project - drag a CSV (such as one attached)
- Expect: Chambers listed for PA or IL, population deviation box displayed
- Select a chamber, and a populationDeviation that is something other than 5
- Expect: Project is imported with the specified parameters

Closes #628

### DIFF
--- a/scripts/load-dev-data
+++ b/scripts/load-dev-data
@@ -64,6 +64,25 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
               ON CONFLICT DO NOTHING;
             "
 
+        # Add chambers to the database
+        docker-compose \
+            exec database psql -U districtbuilder districtbuilder -c "
+              INSERT INTO chamber
+              VALUES (
+                'e7f5999f-88d9-4c58-ab85-f65eae843ea1',
+                'House of Representatives',
+                 18,
+                '6c440b8d-c26e-4bae-850f-dbff37fe0209'
+              ),
+              (
+                '2202e945-161b-4f84-a282-5d65c7ba16b7',
+                'House of Representatives',
+                 18,
+                '96bca832-99b3-4c4d-8913-ab4ca82ec442'
+              )
+              ON CONFLICT DO NOTHING;
+            "
+
         # Add testing organization
         ./scripts/manage update-organization dev-data/azavea.yaml
         
--- a/src/client/screens/ImportProjectScreen.tsx
+++ b/src/client/screens/ImportProjectScreen.tsx
@@ -4,10 +4,27 @@ import React, { useCallback, useEffect, useRef, useState } from "react";
 import { FileDrop } from "react-file-drop";
 import { connect } from "react-redux";
 import { Link, Redirect } from "react-router-dom";
-import { Box, Button, Card, Flex, Heading, jsx, Spinner, ThemeUIStyleObject } from "theme-ui";
+import {
+  Box,
+  Button,
+  Card,
+  Flex,
+  Heading,
+  jsx,
+  Spinner,
+  ThemeUIStyleObject,
+  Label,
+  Radio
+} from "theme-ui";
 
-import { FIPS, MaxUploadFileSize } from "../../shared/constants";
-import { DistrictsDefinition, ImportRowFlag, IProject, IRegionConfig } from "../../shared/entities";
+import { DEFAULT_POPULATION_DEVIATION, FIPS, MaxUploadFileSize } from "../../shared/constants";
+import {
+  DistrictsDefinition,
+  ImportRowFlag,
+  IProject,
+  IRegionConfig,
+  IChamber
+} from "../../shared/entities";
 
 import { regionConfigsFetch } from "../actions/regionConfig";
 import { setImportFlagsModal } from "../actions/districtDrawing";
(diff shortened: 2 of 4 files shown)


scripts/load-dev-data
```diff
@@ -64,6 +64,25 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
...
+              ),
+              (
+                '2202e945-161b-4f84-a282-5d65c7ba16b7',
+                'House of Representatives',
+                 18,
+                '96bca832-99b3-4c4d-8913-ab4ca82ec442'
+              )
+              ON CONFLICT DO NOTHING;
+            "
+
```
maurizi, line 85: 🎆

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
   const regionConfig = importResource.data;
   const districtsDefinition = "resource" in importResource ? importResource.resource : null;
   const numberOfDistricts = form.numberOfDistricts;
+  const populationDeviation = form.populationDeviation || DEFAULT_POPULATION_DEVIATION;
```
maurizi, line 53:
```suggestion
  const populationDeviation = form.populationDeviation;
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
   return numberOfDistricts &&
     maxDistrictId &&
     regionConfig &&
+    populationDeviation &&
```
maurizi, line 57:
```suggestion
    populationDeviation !== null &&
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
   readonly valid: true;
 }
 
 interface InvalidForm {
   readonly regionConfig: IRegionConfig | null;
+  readonly chamber?: IChamber;
   readonly districtsDefinition: DistrictsDefinition | null;
   readonly numberOfDistricts: number | null;
+  readonly isCustom?: boolean;
+  readonly populationDeviation: number;
```
maurizi, line 90: You should match the existing style of the surrounding code - it uses `null` for missing data, not `undefined` (which is what using `?` implies).
Additionally
 - it is possible for someone to leave the population deviation field empty, so we should represent that state in our `InvalidForm` type.
 - `isCustom` shouldn't be optional
```suggestion
  readonly chamber: IChamber | null;
  readonly districtsDefinition: DistrictsDefinition | null;
  readonly numberOfDistricts: number | null;
  readonly isCustom: boolean;
  readonly populationDeviation: number | null;
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
 }
 
 type ProjectForm = ValidForm | InvalidForm;
 
 interface ValidForm {
   readonly regionConfig: IRegionConfig;
+  readonly chamber?: IChamber;
   readonly districtsDefinition: DistrictsDefinition;
   readonly numberOfDistricts: number;
+  readonly isCustom?: boolean;
```
maurizi, line 79:
```suggestion
  readonly isCustom: boolean;
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
-    : { numberOfDistricts, regionConfig, districtsDefinition, valid: false };
+    ? { numberOfDistricts, regionConfig, districtsDefinition, populationDeviation, valid: true }
+    : { numberOfDistricts, regionConfig, districtsDefinition, populationDeviation, valid: false };
 };
 
 type ImportResource = WriteResource<IRegionConfig | null, DistrictsDefinition>;
 
 interface ConfigurableForm {
   readonly numberOfDistricts: number | null;
+  readonly isCustom?: boolean;
```
maurizi, line 68: I don't think `isCustom` should ever be `null` / `undefined` right? - the checkbox is either checked or not.
```suggestion
  readonly isCustom: boolean;
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
+  readonly populationDeviation: number;
 }
 
 type ProjectForm = ValidForm | InvalidForm;
 
 interface ValidForm {
   readonly regionConfig: IRegionConfig;
+  readonly chamber?: IChamber;
```
maurizi, line 76: The code you added actually uses `null` not undefined.

```suggestion
  readonly chamber: IChamber | null;
```

src/client/screens/ImportProjectScreen.tsx
```diff
@@ -33,39 +50,50 @@ const validate = (
...
   return numberOfDistricts &&
     maxDistrictId &&
     regionConfig &&
+    populationDeviation &&
     districtsDefinition &&
     numberOfDistricts >= maxDistrictId
-    ? { numberOfDistricts, regionConfig, districtsDefinition, valid: true }
-    : { numberOfDistricts, regionConfig, districtsDefinition, valid: false };
+    ? { numberOfDistricts, regionConfig, districtsDefinition, populationDeviation, valid: true }
+    : { numberOfDistricts, regionConfig, districtsDefinition, populationDeviation, valid: false };
```
maurizi, line 61: Note that this function needs to add `chamber` to the returned data here as well, which will require that type being added to `ConfigurableForm`.

Right now when I select a `Chamber` the frontend is _merely_ using that to set `numberOfDistricts` - we should include `chamber` in our POST data as well.

maurizi (changes requested):
Thanks for the changes to `load-dev-data`, that's a nice productivity-improvement.

I left a few comments re: cleaning up the types and making sure we post the correct data (we're not handling deviation = 0 correct, and we're not including the selected Chamber).

src/client/screens/ImportProjectScreen.tsx
kevinearldenny: Chamber should now be saved with an imported project, and popDeviation of 0 should be handled properly

maurizi (approved):
Working great 👍

PR #847, opened 2021-07-12
Add button to competitiveness page for uploading a plan to PlanScore

### LINKED ISSUE #838: Integrate with PlanScore API
PlanScore has an API now, and we'd like to integrate with it so we can provide detailed competitive analysis in DistrictBuilder.

API documentation here: https://github.com/PlanScore/PlanScore/blob/main/API.md

AC:
* There's a button on the Evaluate: Competitiveness page that opens a new tab and calls out to the PlanScore API to load the report for the given map.
* Text: "View more advanced political metrics for your map with PlanScore"

kevinearldenny:
## Overview

- Add button to competitiveness page for uploading a plan to PlanScore
- Add new endpoint to upload to PlanScore API from backend
- Open PlanScore in new window when API response received

### Checklist

### Demo

image
image
image

### Notes

- The "Open in new tab" suggestion feels a bit clunky to me. By default, Chrome seems to block the popup for me on localhost unless I explicitly allow it. Personally, I might prefer a link to be displayed in the sidebar or in a new modal allowing the user to open the PlanScore result in a new tab at their leisure.

- it seems like bad etiquette for us to allow users to ping PlanScore's API multiple times for a district plan that has not changed. One potential idea I had to rectify this would be to store the `plan_score_url` received from the API and the `plan_score_last_updated_dt` of when the PlanScore was last computed, and to disable the 'Send to PlanScore' button unless the Project's `updatedDt` is more recent than the last time PlanScore was checked. 

- It might also be nice to add some feedback that the PlanScore API is processing the request - we could use the existing spinner from the project, but I'm curious to see if @jfrankl has any other ideas

## Testing Instructions

- `./scripts/update`
- Open a finished project and navigate to the Competitiveness metric page
- Click the `Send to PlanScore API` button and wait for a new window to appear. Note that you may need to enable popups on localhost for the popup to appear
- Expect: A new window indicating successful upload to PlanScore appears
- Wait a minute and refresh the PlanScore window - expect the PlanScore report for the district to be computed

Closes #838


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -484,4 +485,35 @@ export class ProjectsController implements CrudController<Project> {
...
   }
+
+  @UseInterceptors(CrudRequestInterceptor)
+  @UseGuards(JwtAuthGuard)
+  @Post(":id/planScore")
+  async sendToPlanScoreAPI(
+    @ParsedRequest() req: CrudRequest,
+    @Param("id") projectId: ProjectId
+  ): Promise<Project> {
+    const planScoreToken = "Hrnt68Zy";
```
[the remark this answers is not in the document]
maurizi: Um... by posting this on a public repository we're going to need to ask PlanScore retire this API key and provide us w/ a new one (I'm assuming it's not meant to be public since it is being used in a server-side API).

As for how we _should_ handle this - https://github.com/PublicMapping/districtbuilder/pull/338 shows a good example of how to add a new secret to the application - look for how `rollbar_access_token` / `ROLLBAR_ACCESS_TOKEN` get threaded through the application. Note also that you'll need to add the new environment to the Terraform configs in AWS - they're located in these buckets:
https://github.com/PublicMapping/districtbuilder/blob/e38c66bb6f079c69670d782a3a50910df5f85d67/Jenkinsfile.release#L43
https://github.com/PublicMapping/districtbuilder/blob/e38c66bb6f079c69670d782a3a50910df5f85d67/Jenkinsfile#L14

Additionally we should add something in the README on setting up the PlanScore API key locally for dev/testing - likely by using a `.env` file we hide via our `.gitignore`, like we do for [Treetective](https://github.com/azavea/treetective#setup) (maybe with a `.env.sample` file for documentation purposes?)

maurizi (changes requested):
> The "Open in new tab" suggestion feels a bit clunky to me. By default, Chrome seems to block the popup for me on localhost unless I explicitly allow it. Personally, I might prefer a link to be displayed in the sidebar or in a new modal allowing the user to open the PlanScore result in a new tab at their leisure.

Yeah - based on how their API works I agree that I think we'll want the user to do two clicks - one to initiate uploading to PlanScore, and a second to open the results in a new tab - triggering the popup-blocker seems like the wrong way to go about this to me.

@jfrankl what do you think of the following:
 - Users clicks "Send to PlanScore" button
 - Buttons changes to "Sending to PlanScore 🔃" w/ a built-in spinner
 - When ready button changes to "View on PlanScore"
 
 If that is too subtle I'd say we should go w/ a new modal as Kevin suggested.

jfrankl:
> * Users clicks "Send to PlanScore" button
> * Button is disabled, & changes to "Sending to PlanScore 🔃" w/ a built-in spinner
> * When ready button is re-enabled & changes to "View on PlanScore", and remains this way until the user makes changes to their plan

@maurizi this is great. One change I'd make: on that third step, instead of re-enabling the button, let's replace it with a link that reads "View on PlanScore". I think the change from a button to a link will make it extra clear to the user that something has changed and now a link is ready for them to click.

kevinearldenny:
> > * Users clicks "Send to PlanScore" button
> > * Button is disabled, & changes to "Sending to PlanScore 🔃" w/ a built-in spinner
> > * When ready button is re-enabled & changes to "View on PlanScore", and remains this way until the user makes changes to their plan
> 
> @maurizi this is great. One change I'd make: on that third step, instead of re-enabling the button, let's replace it with a link that reads "View on PlanScore". I think the change from a button to a link will make it extra clear to the user that something has changed and now a link is ready for them to click.

Okay, I have implemented this now. The spinner is a bit hard to see on the button currently, though so we could possibly use a style pass here although I'm not sure how necessary this is.

src/client/components/evaluate/detail/Competitiveness.tsx
```diff
@@ -114,6 +137,38 @@ const CompetitivenessMetricDetail = ({
...
+              ...{
+                variant: "buttons.primary",
+                fontWeight: "light",
+                maxHeight: "34px",
+                borderBottom: "none",
+                borderBottomColor: "blue.2"
+              },
+              ...style.menuButton
+            }}
+            onClick={() => sendToPlanScore()}
```
maurizi, line 153: We should disable the button while the request is pending to avoid duplicate requests
```suggestion
            onClick={() => sendToPlanScore()}
            disabled={planScoreLoaded === false}
```

docker-compose.ci.yml
```diff
@@ -23,6 +23,7 @@ services:
       - DB_DEBUG=1
       - DB_SETTINGS_BUCKET=${DB_SETTINGS_BUCKET:-districtbuilder-staging-config-us-east-1}
       - DB_ROLLBAR_ACCESS_TOKEN
+      - DB_PLAN_SCORE_API_TOKEN
```
maurizi, line 26: I don't think this is needed
```suggestion
```

src/server/.env.sample
```diff
@@ -0,0 +1 @@
+PLAN_SCORE_TOKEN=<YOUR KEY HERE>
```
maurizi, line 1:
```suggestion
PLAN_SCORE_API_TOKEN=<YOUR KEY HERE>
```

src/client/components/evaluate/detail/Competitiveness.tsx
```diff
@@ -53,17 +56,37 @@ const style: ThemeUIStyleObject = {
...
+  readonly project?: IProject;
 }) => {
+  const [planScoreLoaded, setPlanScoreLoaded] = useState<boolean | null>(null);
+  const [planScoreLink, setPlanScoreLink] = useState<string | null>(null);
   const choroplethStops = getPviSteps();
+  const openInNewTab = (url: string) => {
+    const newWindow = window.open(url, "_blank", "noopener,noreferrer");
+    // eslint-disable-next-line
+    if (newWindow) newWindow.opener = null;
+  };
```
maurizi, line 81:
```suggestion
```

maurizi (changes requested):
Looking good, just a few small changes needed

maurizi (approved):
👍

PR #848, opened 2021-07-13
Update population deviation helper text

### LINKED ISSUE #804: Rephrase helper text for "Population deviation tolerance"
In the New Map workflow, I find that the helper text for "Population deviation tolerance" does not provide enough support for a new user. It currently reads:

> Depending on whether you are making a map for learning purposes or are trying to make the exercise as realistic as possible

This seems like an incomplete thought. It tells me that this number has something to do with how detailed the map will be, but it doesn't give any advice for how I should set the value. Is a higher or lower number good for learning purposes?

I propose something like:

> How detailed of a map do you want to draw? We recommend using 5–10% for learning purposes, 1–4% for a realistic map, or 0% for a legally viable map.

@tgilcrest what do you think? I made up those ranges—do they check out to you?

kevinearldenny:
## Overview

Updates population deviation helper text

### Checklist

### Demo

image

## Testing Instructions

- Open create project screen and select a region

Closes #804


maurizi (approved):
👍

PR #850, opened 2021-07-14
Add maps to list of user projects on home screen

### LINKED ISSUE #790: Add maps to user project list
We want to add maps to the user's project list page, as seen in this mockup:

https://share.goabstract.com/14d5b608-1369-4ed6-b4f2-20b5ce7e4c63?mode=design&sha=09b392079f20d92a20fe38878b27420f0bd1e703

AC:
 - The list of user projects shows maps
 - The map list is paginated when it is longer than some set amount (12?)

kevinearldenny:
## Overview

Add maps to user project cards on home screen

- Add pagination to `/api/projects` endpoint

### Checklist

### Demo

image

### Notes

The CSS is a bit wonky currently, this might need a style pass.

Additionally, I'm not totally sold on my direction of editing the `/api/projects` endpoint - it seems like it might have been a better idea to refactor the `globalProjects` endpoint to support filtering to only the user's projects. Extrapolating this further - could we possibly combine both of these screens into a single screen, with an option to view user projects or global projects? It seems like there is a lot of redundancy here, and especially if we want to support filtering in global projects (#833 ) it would be nice to have this in user projects as well .

## Testing Instructions

- `./scripts/server`
- With more than 4 projects, expect to see pagination and maps displayed for each project

Closes #790

### DIFF
--- a/src/client/components/HomeScreenProjectCard.tsx
+++ b/src/client/components/HomeScreenProjectCard.tsx
@@ -1,0 +1,110 @@
+/** @jsx jsx */
+import { Box, Flex, Heading, jsx } from "theme-ui";
+import { useHistory } from "react-router-dom";
+
+import { IProject } from "../../shared/entities";
+import ProjectListFlyout from "./ProjectListFlyout";
+import TimeAgo from "timeago-react";
+import ProjectDistrictsMap from "./map/ProjectDistrictsMap";
+
+const style = {
+  featuredProject: {
+    width: "100%",
+    bg: "#fff",
+    borderRadius: "2px",
+    display: "inline-block",
+    mb: "20px",
+    boxShadow: "small",
+    position: "relative"
+  },
+  mapLabel: {
+    p: "15px",
+    display: "inline-block",
+    width: "600px",
+    pl: "100px",
+    borderColor: "gray.2",
+    position: "absolute"
+  },
+  projectTitle: {
+    display: "inline-block",
+    width: "300px",
+    "&:hover": {
+      cursor: "pointer"
+    }
+  },
+  projectRow: {
+    textDecoration: "none",
+    display: "flex",
+    alignItems: "baseline",
+    borderRadius: "med",
+    marginRight: "auto",
+    px: 1,
+    "&:hover:not([disabled])": {
+      bg: "rgba(256,256,256,0.2)",
+      h2: {
+        color: "primary",
+        textDecoration: "underline"
+      },
+      p: {
+        color: "blue.6"
+      }
+    },
+    "&:focus": {
+      outline: "none",
+      boxShadow: "focus"
+    }
+  },
+  flyoutButton: {
+    position: "absolute",
+    top: "5px",
+    right: "10px"
+  }
+} as const;
+
+const HomeScreenProjectCard = ({ project }: { readonly project: IProject }) => {
+  const history = useHistory();
+
+  function goToProject(project: IProject) {
+    history.push(`/projects/${project.id}`);
+  }
+
+  return (
+    <Flex sx={style.featuredProject}>
+      <ProjectDistrictsMap project={project} context={"home"} />
+      <Box sx={style.mapLabel}>
+        <Box sx={style.projectTitle} onClick={() => goToProject(project)}>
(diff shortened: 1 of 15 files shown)


src/server/src/projects/services/projects.service.ts
```diff
@@ -51,4 +51,30 @@ export class ProjectsService extends TypeOrmCrudService<Project> {
...
+        "project.updatedDt",
+        "project.createdDt",
+        "project.districts",
+        "regionConfig.name",
+        "user.id",
+        "user.name"
+      ])
+      .orderBy("project.updatedDt", "DESC");
+    return paginate<Project>(builder, options);
+  }
```
kevinearldenny, line 79: Refactor this + `findAllProjectsPaginated` into a single function with helper functions to apply the where clauses for `user` and `published`

src/server/src/projects/services/projects.service.ts
```diff
@@ -51,4 +56,13 @@ export class ProjectsService extends TypeOrmCrudService<Project> {
       : builder;
     return paginate<Project>(builderWithFilter, options);
   }
+
+  async findAllUserProjectsPaginated(
+    userId: string,
+    options: AllProjectsOptions
+  ): Promise<Pagination<Project>> {
+    const builder = this.getProjectsBase().andWhere("user.id = :userId", { userId });
```
maurizi, line 64: This _also_ needs to filter out `archived` projects.
```suggestion
    const builder = this.getProjectsBase().andWhere(
      "project.archived = FALSE AND user.id = :userId",
      { userId }
    );
```

maurizi (changes requested):
Overall working and looking good - I left some comments inline on suggested changes.

src/client/components/HomeScreenProjectCard.tsx
```diff
@@ -0,0 +1,191 @@
...
+        </div>
+      </Box>
+      <span sx={style.flyoutButton}>
+        <ProjectListFlyout project={project} sx={{ display: "inline-block" }} />
+      </span>
+    </Flex>
+  );
+};
+
+export default HomeScreenProjectCard;
```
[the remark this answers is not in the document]
kevinearldenny: Good idea. Implemented now with a new `ProjectDistrictsMap` component

src/client/components/map/ProjectDistrictsMap.tsx
```diff
@@ -0,0 +1,110 @@
...
+        source: "districts",
+        layout: {},
+        paint: {
+          "fill-color": { type: "identity", property: "color" }
+        }
+      });
+      map.resize();
+
+      setMapLoaded(true);
+    });
```
maurizi, line 84: React is showing a warning in the console re: this handler:
image

We need to turn off the `load` event handler if/when the React component gets unloaded - otherwise as the warning indicates we'll have a memory leak. In this case that includes the Mapbboxgl map, so this could potentially be a large performance concern

```suggestion
    function onLoad() {
      districts &&
        map.addSource("districts", {
          type: "geojson",
          data: districts
        });
      map.addLayer({
        id: "districts",
        type: "fill",
        source: "districts",
        layout: {},
        paint: {
          "fill-color": { type: "identity", property: "color" }
        }
      });
      map.resize();

      setMapLoaded(true);
    }
    map.on("load", onLoad);
    return () => {
      map.off("load", onLoad);
    };
```

src/client/components/map/ProjectDistrictsMap.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { ProjectNest } from "../../../shared/entities";
+import { DistrictGeoJSON } from "../../types";
+import { getDistrictColor } from "../../constants/colors";
+
+const style = {
+  mapContainer: {
+    height: "250px",
+    display: "inline-block",
+    position: "relative",
+    p: "15px"
```
maurizi, line 18: I assume you had this at some point and lost it?
All of the flyout menus were hiding up in the upper-right corner of the screen w/o this.
```suggestion
    p: "15px",
    position: "relative"
```

maurizi (changes requested):
Just a few small changes needed.

maurizi (approved):
👍

PR #851, opened 2021-07-16
Add filter by state functionality to community maps page

### LINKED ISSUE #833: Add state dropdown to Community Maps page

kevinearldenny:
## Overview

- Add filter by state functionality to community maps page
- Add URL param filter to enable direct navigation to maps for a specific state

### Checklist

### Demo

image
image

### Notes

I'm struggling to figure out where the "0" that's being rendered when no maps exist is coming from - it shouldn't be there. We also may need a style pass on this, since there were no wireframes for the filter included in the issue

## Testing Instructions

- `./scripts/update`
- Navigate to the community maps page and select a state with at least one completed map
- Expect: navigate to `maps?region={STATE}`, and maps for that state are displayed
- Navigate to a URL for a state
- Expect: Map for state are displayed

Closes #833

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -21,6 +21,8 @@ import theme from "./theme";
 import "./App.css";
 import StartProjectScreen from "./screens/StartProjectScreen";
 import PublishedMapsListScreen from "./screens/PublishedMapsListScreen";
+import { PushReplaceHistory, QueryParamProvider } from "use-query-params";
+import { createBrowserHistory } from "history";
 
 const PrivateRoute = ({ children, ...props }: RouteProps) => {
   const savedJWT = getJWT();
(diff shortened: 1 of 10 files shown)


maurizi (changes requested):
I left a couple of comments inline for how to clean up some style issues.

Additionally, the integration with `use-query-params` / `react-router` needs some improvement:
 - Using the back / forward buttons causes the URL to change, but the UI does not reflect those changes
 - If I refresh the page w/ a state filter applied, I am unable to select "All states" (it re-applies the state filter from page load)

src/client/App.tsx
```diff
@@ -39,6 +41,17 @@ const PrivateRoute = ({ children, ...props }: RouteProps) => {
...
+const history = createBrowserHistory(),
+  pushReplaceHistory: PushReplaceHistory = {
+    push: (location: Location): void => {
+      history.push(location);
+    },
+    replace: (location: Location): void => {
+      history.replace(location);
+    },
+    location: window.location
+  };
```
maurizi, line 53: Its a matter of opinion, but I find this more readable:
```suggestion
const history = createBrowserHistory();
const pushReplaceHistory: PushReplaceHistory = {
  push: (location: Location): void => {
    history.push(location);
  },
  replace: (location: Location): void => {
    history.replace(location);
  },
  location: window.location
};
```

maurizi (approved):
Working well 👍

PR #853, opened 2021-07-19
Add majority race metric to project sidebar

kevinearldenny:
## Overview

- Adds the majority race metric to project sidebar
- Allows the user to pin majority race metric

### Checklist

- [ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

### Demo

image

## Testing Instructions

- Open a project and toggle the expanded metrics
- Expect majority race is displayed at the end of demographic headers
- Pin the majority race metric
- Expect majority race to be displayed when expanded metrics are collapsed

Closes #830

PR #854, opened 2021-07-19
Add additional keyboard shortcuts

kevinearldenny:
## Overview

- Add keyboard shortcut (`P`) to toggle expanded project metrics
- Add keyboard shortcut (`L`) to zoom to district

### Checklist

### Demo

image

### Notes

While working on #853 , I thought it would be nice to have a toggle for collapsing / expanding project metrics. While I was digging back into keyboard shortcuts, I realized that we didn't currently have a shortcut for zooming to the currently selected district either which seems like it could be helpful 

## Testing Instructions

- Open a project and hit the `P` key - expect expanded project metrics to be toggled
- Open a project and hit the `L` key on a populated district - expect the map to zoom to that district


maurizi (approved):
👍

PR #864, opened 2021-07-21
Update display logic for political data columns in sidebar to correspond with alt data structure

### LINKED ISSUE #855: Political data missing from expandable metrics viewer
## Description
After reviewing the metrics viewer on staging I found that political data is no longer displaying as a column. This should continue to be displayed and have an option to pin just as the other demographic fields.

Metrics viewer on staging:
image

Metrics viewer in production:
image

kevinearldenny:
## Overview

Support displaying political data in sidebar for states where 2016 political data is displayed as just `republican` or `democrat` instead of `republican16` / `democrat16`

### Checklist

### Demo

image

### Notes

1. On some level, this fix seems clunky to me. Is there any way that we could just rename the columns in the data so that the year is always included? 
1. Are there any other potential data structures that we need to be aware of? Is it possible that 2020 data could be added later for some states where we currently have the party-only structure? If so, this seems like it may be problematic.

## Testing Instructions

- Go to a project in staging in a state with the data structure outlined above (e.g., Pennsylvania)
- Expect to see political data columns in expandable metrics viewer

Closes #855

### DIFF
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -212,8 +212,14 @@ export const hasMultipleElections = (staticMetadata?: IStaticMetadata) =>
   staticMetadata?.voting?.some(file => file.id.endsWith("16")) &&
   staticMetadata?.voting?.some(file => file.id.endsWith("20"));
 
-export const has16Election = (staticMetadata?: IStaticMetadata) =>
-  staticMetadata?.voting?.some(file => file.id.endsWith("16"));
+export const has16Election = (staticMetadata?: IStaticMetadata) => {
+  return (
+    staticMetadata?.voting?.some(file => file.id.endsWith("16")) ||
+    (staticMetadata?.voting &&
+      Object.keys(staticMetadata?.voting || {}).length > 0 &&
+      !has20Election(staticMetadata))
+  );
+};
 
 export const demographicsHasOther = (staticMetadata?: IStaticMetadata) =>
   staticMetadata?.demographics?.some(file => file.id === "other");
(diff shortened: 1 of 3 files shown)


src/client/functions.ts
```diff
@@ -212,8 +212,14 @@ export const hasMultipleElections = (staticMetadata?: IStaticMetadata) =>
...
 
-export const has16Election = (staticMetadata?: IStaticMetadata) =>
-  staticMetadata?.voting?.some(file => file.id.endsWith("16"));
+export const has16Election = (staticMetadata?: IStaticMetadata) => {
+  return (
+    staticMetadata?.voting?.some(file => file.id.endsWith("16")) ||
+    (staticMetadata?.voting &&
+      Object.keys(staticMetadata?.voting || {}).length > 0 &&
+      !has20Election(staticMetadata))
+  );
```
kevinearldenny, line 221: This is based around the assumption that any party-only values (`republican`, `democrat`) without a year explicitly specified in suffix will always be for 2016

maurizi (approved):
👍 It's fixed

maurizi:
>On some level, this fix seems clunky to me. Is there any way that we could just rename the columns in the data so that the year is always included?

I believe this would require:
 - renaming the columns in the files uploaded on S3 (these are embedded within the TopoJSON features, so the simplest way to do this would be to reprocess those states starting from the block-level data)
 - renaming them within the cached `districts` column (this is basically a more specialized variation of #828 )
 
 That's possible, but it seems like a lot more work to me than simply keeping our code backwards-compatible with our old format, even if it is admittedly clunky.

> Are there any other potential data structures that we need to be aware of? Is it possible that 2020 data could be added later for some states where we currently have the party-only structure? If so, this seems like it may be problematic.

We're going to archive (make read-only) our existing regions at the same time that we roll out the 2020 data. This won't be an issue.

PR #871, opened 2021-07-26
Fix find menu button so it properly opens the find menu

### LINKED ISSUE #859: Find menu doesn't open
Trying to open the Find Unassigned / Fin non-contiguous menu no longer works

I think this was broken by https://github.com/PublicMapping/districtbuilder/pull/835

kevinearldenny:
## Overview

- Fixes the bug identified in #859 around the find menu not opening
- Toggles the paintbrush size slider when the paintbrush icon is clicked 

### Checklist

### Demo

image

### Notes

It seemed like putting this on the right side was the _easiest_ solution here, but I'm not necessarily sure if it's the best solution from a UX perspective. Perhaps @jfrankl might have a better solution for how to make this more visually apparent / usable

## Testing Instructions

- Open a project
- Click the find menu; expect the find menu to appear

Closes #859

### DIFF
--- a/src/client/components/MapHeader.tsx
+++ b/src/client/components/MapHeader.tsx
@@ -3,7 +3,7 @@ import React, { useState } from "react";
 import { Flex, Box, Label, Button, jsx, Select, Slider, ThemeUIStyleObject } from "theme-ui";
 import { GeoLevelInfo, GeoLevelHierarchy, GeoUnits, IStaticMetadata } from "../../shared/entities";
 import { ElectionYear } from "../types";
-
+import { toggleFind } from "../actions/districtDrawing";
 import { geoLevelLabel, capitalizeFirstLetter, canSwitchGeoLevels } from "../functions";
 import MapSelectionOptionsFlyout from "./MapSelectionOptionsFlyout";
 
@@ -145,6 +145,7 @@ const MapHeader = ({
   selectionTool,
   paintBrushSize,
   geoLevelIndex,
+  findMenuOpen,
   selectedGeounits,
   isReadOnly,
   limitSelectionToCounty,
@@ -158,6 +159,7 @@ const MapHeader = ({
   readonly selectedGeounits: GeoUnits;
   readonly advancedEditingEnabled?: boolean;
   readonly isReadOnly: boolean;
+  readonly findMenuOpen: boolean;
   readonly limitSelectionToCounty: boolean;
   readonly electionYear: ElectionYear;
 }) => {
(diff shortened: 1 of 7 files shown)


src/client/components/MapHeader.tsx
```diff
@@ -285,7 +289,7 @@ const MapHeader = ({
         )}
         <Flex className="geolevel-button-group">{geoLevelOptions}</Flex>
       </Flex>
-      <Box sx={{ lineHeight: "1" }}>
+      <Box sx={{ lineHeight: "1", position: "fixed", right: "150px" }}>
```
maurizi, line 292: It's very rare that you want `position: "fixed";` - you almost always want  `position: "absolute";` instead (the exception that proves the rule is clarifying here: an element that is intended to remain fixed when the page scrolls _and_ be overlaid above normal page content, like a floating "Help" button, is a good candidate for using `position: "fixed";`).

`"fixed"` means absolute positioning in regards to the _viewport_, whereas `"absolute"` means absolute positioning in regards to the parent element (or parent's parent, etc.).

In particular for this case, this would mean that by using `"fixed"` positioning our styling would break if the header above this button changes height, whereas using `"absolute"` positioning it will not.

```suggestion
      <Box sx={{ lineHeight: "1", position: "absolute", right: "150px" }}>
```

src/client/components/MapHeader.tsx
```diff
@@ -307,6 +311,23 @@ const MapHeader = ({
           </Select>
         </Flex>
       </Box>
+      <Box sx={{ position: "relative", mr: "5px" }}>
+        <Button
+          sx={{ ...style.selectionButton }}
+          // eslint-disable-next-line
+          onClick={() => store.dispatch(toggleFind(!findMenuOpen))}
```
maurizi, line 318: This is a cleaner way to fix the lint error:
```suggestion
          onClick={() => {
            store.dispatch(toggleFind(!findMenuOpen));
          }
```

src/client/components/MapHeader.tsx
```diff
@@ -307,6 +311,23 @@ const MapHeader = ({
...
+        >
+          <Box
+            sx={{
+              borderBottom: findMenuOpen ? "solid 1px" : "none",
+              borderBottomColor: "muted",
```
maurizi, line 324: The white-on-white isn't showing up here - consider either switching to a dark border:
```suggestion
              borderBottomColor: "muted",
```

kevinearldenny: Good catch.

maurizi (approved):
I can find things again 🎉

PR #872, opened 2021-07-27
Fix the duplicate project functionality on the home screen

### LINKED ISSUE #870: Duplicate project button on home screen not working
## Description
It seems like in the refactor of user maps on the home screen (#790 ) we broke the duplicate project functionality on the user project card

## Steps To Reproduce
1. Go to home screen
2. Click on the flyout menu and select 'Duplicate Project'

## Expected behavior
Project is duplicated

## Actual behavior
An error is returned and project is not duplicated

kevinearldenny:
## Overview

- Adds support back in for duplicating projects from the home screen, which was broken previously in #850 when `districtsDefinition` was removed from API response
- Updates UX for duplicating projects so that a user is redirected to a duplicated project after the duplication completes
- Leverages `isSaving` and `duplicatedProject` variables in Redux store to support

### Checklist

### Notes

I don't _love_ the approach I took for handling the redirect - it seems like there may be a way to handle this with `Resource`, but it would require a bit of refactor at least and I'm not entirely sure that it's worthwhile. 

Also, I've noticed that locally the duplicate sometimes times out due to the region configs taking a while to load. I pushed to staging to confirm that this wasn't going to be a problem there, and it looks fine - might be a better idea to test this on staging, though.

## Testing Instructions

- Go to home screen and click the flyout menu button then 'Duplicate project'
- Expect: project is duplicated, and user is redirected to the new project page

Closes #870

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,13 +43,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Tweaks to org report downloads [#832](https://github.com/PublicMapping/districtbuilder/pull/832)
 - Update population deviation helper text [#848](https://github.com/PublicMapping/districtbuilder/pull/848)
 - Update display logic for political data in expandable metrics viewer [#864](https://github.com/PublicMapping/districtbuilder/pull/864)
-
+- Update UX for duplication, showing a spinner when duplication is pending and redirecting when it completes [#872](https://github.com/PublicMapping/districtbuilder/pull/872)
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


maurizi:
> Also, I've noticed that locally the duplicate sometimes times out due to the region configs taking a while to load. I pushed to staging to confirm that this wasn't going to be a problem there, and it looks fine - might be a better idea to test this on staging, though.

We don't consider our application servers "healthy" until they have finished loading the region configs / topojson, so this indeed won't be an issue on production (or in development, if you simply wait long enough - it's only an issue when the application has first started).

maurizi (approved):
This is great 👍

PR #874, opened 2021-07-27
Fix overlap between keyboard shortcuts and paintbrush slider

### LINKED ISSUE #860: Keyboard shortcuts modal overlapping with paintbrush slider
## Description

image

When the paintbrush slider is open and the user types `?` to trigger the keyboard shortcuts modal, the slider is displayed on top of the keyboard shortcuts modal

## Expected behavior
Paintbrush slider is displayed underneath modal

## Actual behavior
Paintbrush slider displayed atop modal

kevinearldenny:
## Overview

- Fixes the bug identified in #860 where the paintbrush slider rendered over top of the keyboard shortcuts modal
- Finalizes the UX improvements to paintbrush slider that were began in #871 . It was never explicitly spec'd out when the paintbrush size slider should be displayed, and it previously was not displaying the paintbrush slider if a user selected the paintbrush tool via keyboard shortcuts
- Adds new keyboard shortcuts for incrementing / decrementing the size of the paintbrush tool

### Checklist

### Demo

image

image

### Notes

Keyboard shortcuts weren't explicitly asked for around the paintbrush size slider, but they certainly are fun to use 😀 

I kind of randomly went with `B` and `M` for the increment / decrement functionality - perhaps there might be better choices here, @jfrankl ? 

## Testing Instructions

- Go to a project, select the paintbrush tool and then type `?` to open the shortcuts modal
- Expect: Shortcuts modal is rendered above the paintbrush slider
- Click `B` or `M` while the paintbrush tool is selected.
- Expect: Paintbrush is incremented / decremented

Closes #860

### DIFF
--- a/src/client/components/MapHeader.tsx
+++ b/src/client/components/MapHeader.tsx
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Flex, Box, Label, Button, jsx, Select, Slider, ThemeUIStyleObject } from "theme-ui";
 import { GeoLevelInfo, GeoLevelHierarchy, GeoUnits, IStaticMetadata } from "../../shared/entities";
 import { ElectionYear } from "../types";
@@ -77,7 +77,7 @@ const style: ThemeUIStyleObject = {
     borderColor: "gray.2",
     padding: "5px",
     top: "100px",
-    zIndex: 99999,
+    zIndex: 1,
     justifyContent: "space-evenly",
     width: "300px"
   }
@@ -191,6 +191,15 @@ const MapHeader = ({
         ]
       : undefined;
 
+  // Close paintbrush slider if the user switches to another tool, open it if toggled from another tool
+  useEffect(() => {
+    if (selectionTool !== SelectionTool.PaintBrush) {
+      setPaintBrushSizeSliderVisibility(false);
+    } else {
+      setPaintBrushSizeSliderVisibility(true);
+    }
+  }, [selectionTool]);
+
   const labelOptions = labelFields
     ? labelFields.map(val => (
         <option key={val.id} value={val.id}>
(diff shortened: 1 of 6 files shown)


src/client/components/MapHeader.tsx
```diff
@@ -272,6 +282,7 @@ const MapHeader = ({
                   }}
                   sx={{ width: "150px" }}
                   defaultValue={paintBrushSize}
+                  value={paintBrushSize}
```
maurizi, line 285: This line is causing a console error (https://fb.me/react-controlled-components):
image

```suggestion
                  value={paintBrushSize}
```

src/client/components/MapHeader.tsx
```diff
@@ -77,7 +77,7 @@ const style: ThemeUIStyleObject = {
     borderColor: "gray.2",
     padding: "5px",
     top: "100px",
-    zIndex: 99999,
+    zIndex: 999,
```
maurizi, line 80: This is still too large - in general, try and keep your `z-index` values as low as you can afford to; in this case just a `z-index` of `1` was sufficient to get this to appear above the map

image


```suggestion
    zIndex: 1,
```

maurizi (changes requested):
Left a few comments that need to be addressed - but did verify the new keyboard shortcuts are working.

jfrankl:
In terms of what the shortcuts should be, I was trying to stick with keys on the left hand in QWERTY, with the idea that a user could have one hand on the keyboard and the other hand on the mouse. That's not perfect because not everyone is right handed or uses QWERTY.

src/client/components/MapHeader.tsx
```diff
@@ -247,6 +256,7 @@ const MapHeader = ({
                     sx={{ ...style.selectionButton }}
                     className={buttonClassName(selectionTool === tool)}
                     onClick={() => {
+                      // Open slider on click if not already open, close it otherwise
                       setPaintBrushSizeSliderVisibility(
                         tool === SelectionTool.PaintBrush && !isPaintBrushSizeSliderVisible
                       );
```
maurizi, line 262: I don't think we want this behavior - we want the menu to be open so long as you have the paint brush tool selected, right?
```suggestion
                      // Open slider on click if not already open
                      setPaintBrushSizeSliderVisibility(tool === SelectionTool.PaintBrush);
```

kevinearldenny: Hmm, I personally like the ability to toggle the paintbrush slider visibility by clicking on the icon - when you're not using it it's nice to not have it taking up screen real estate. What are your thoughts, @jfrankl ?

maurizi (approved):
👍

PR #875, opened 2021-07-27
Add import file error for unavailable regions

### LINKED ISSUE #852: Import screen should return an error when a CSV for an unsupported state is uploaded
If someone uploads a CSV for a state for which we don't have a corresponding `RegionConfig`, currently no error is shown and the UI doesn't update.

We should return an error stating that we don't support that state yet.

kevinearldenny:
## Overview

Displays an import file error if a user attempts to import a CSV containing rows from a state not currently supported in DistrictBuilder

### Checklist

### Demo

image

### Notes

@dmcglone Thoughts on what this text should be?

## Testing Instructions

- Attempt to import a file for a state not included in local DB database (e.g., California)
- Expect: Error is displayed as shown above

Closes #852


dmcglone:
@kevinearldenny state `ST` not currently supported

maurizi (approved):
👍

PR #876, opened 2021-07-28
Clear eslint cache before running react-scripts start/build tasks

maurizi:
## Overview

I've been encountering `eslint` issues that weren't replicable on CI / other developers machines.
This SO post (https://stackoverflow.com/a/63788215) suggested the `eslint` cache could be the issue, and indeed adding these changes to our `package.json` seems to have fixed the issue.

### Checklist

## Testing Instructions

- this line from `scripts/update` would fail on `develop` w/ spurious lint issues, now it succeeds `docker-compose run --rm --no-deps client bash -c "yarn install && yarn compile && yarn build"`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,16 +3,17 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "react-scripts start",
-    "build": "react-scripts build",
-    "test": "react-scripts test --testPathPattern=src/client",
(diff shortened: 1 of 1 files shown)


kevinearldenny (approved):
Couldn't replicate the issue that spurred this fix locally, but the changes all LGTM 👍

PR #877, opened 2021-07-29
Increase healthcheck DB timeout to 3 seconds

### LINKED ISSUE #867: Database healthcheck is timing out
The production ECS tasks have been experiencing an elevated rate of healthcheck failures. This has resulted in a higher than normal number of unhealthy hosts. Upon closer examination in the ECS logs, it appears that the database query used to test connectivity to the database is timing out.

Screen Shot 2021-07-23 at 4 36 35 PM

Since the database healthchecks were timing out and the tasks were being declared unhealthy, ECS was killing and restarting the tasks, and seems to be continuing to do so semi-frequently. t looks like the application has been restarting semi-frequently for at least several weeks, but we have enough redundant tasks that it rarely resulted in a user-facing outage. 

Looking at the Terminus healthcheck code, the default timeout appears to be `1000ms`, one second. Based on the slow query log in RDS, we were regularly having routine read queries take longer than that around the time of a recent service disruption.

My theory is that the healthcheck query is regularly taking longer than one second to execute, even during periods of modest traffic. My recommendation would be to either:

- Increase the timeout for this healthcheck query, possibly to 3 seconds to match the healthcheck timeout we have defined for the load balancer target group.
- Evaluate the query we use to perform the healthcheck, determine if it's gotten slower as the amount of data managed by DB has grown, and use a less intensive query for regular healthchecks.

maurizi:
## Overview

We've been experiencing increased healthcheck failures lately from slow database response times, leading to application servers cycling out and occasional temporary site outages.

The 3 second timeout matches the timeout on the load balancer target group, which should help us at least keep the site up more during times of heavy load.

### Checklist

### Notes

@colekettler had suggested exploring if the healthcheck query could be simplified, but it is already incredibly simple (`SELECT 1`), so my assumption is that the slowness of the response is not specific to the healthcheck but rather a side effect of some other performance issue.

## Testing Instructions

I'm frankly unsure of how to test this, other than just assuming the documentation is correct.

Closes #867

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix duplicate project button on home screen projects [#872](https://github.com/PublicMapping/districtbuilder/pull/872)
 - Fix find menu button so that it actually opens the find menu [#871](https://github.com/PublicMapping/districtbuilder/pull/871)
 - Show file error when attempting to import a CSV for an unsupported region [#875](https://github.com/PublicMapping/districtbuilder/pull/875)
+- Increased timeout on database healthcheck [#877](https://github.com/PublicMapping/districtbuilder/pull/877)
 
 
 ## [1.6.0] - 2021-05-24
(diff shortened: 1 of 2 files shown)


colekettler (approved):
Nicely done, thanks for this fix! I pushed the `test/fix/mvm/healthcheck-timeout` branch to trigger a staging deploy. Once that finished up I confirmed in the ECS logs that the healthcheck service was reporting a 3000 ms timeout. That gives me confidence that the config is correct.

As for the impact, I think we're going to need to evaluate that empirically in lieu of load testing. This is a small enough change that I'm comfortable deploying it to production and seeing if the average number of unhealthy tasks in the CloudWatch logs decreases in the coming week.

PR #878, opened 2021-07-29
Convert 2010 census region maps to 2020 census regions

### LINKED ISSUE #815: Add button to projects with archived 2010 regions to convert to a 2020 region map
For maps created using archived data from the 2010 census, we want to support converting the map to use 2020 census data so that it can be edited.

PublicMapping has created a crosswalk file for mapping one census block to another: https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/T9VMJO

The basic process here is essentially:
 1. Export the map to the block level
 2. Use the crosswalk to convert from 2010 to 2020 block ids
 3. Import from the block level using the existing import mechanisms

AC:
 - There is a button "Create a 2020 map" in the header for any projects with an archived 2010 region, which will create a new project, with the same data, but converted to use 2020 geographies
 - The name of the newly created project should have ` (2020)` appended to it
 - Redirect to the new map

maurizi:
## Overview

This adds a button to projects for archived 2010 regions, allowing them to be converted to new 2020 regions.

### Checklist

### Demo

2010 map:
image

Converted map:
image

### Notes

The conversion process _mostly_ worked for the state I tested (DE), but note the 5 clearly incorrectly assigned blocks (4 of which have 0 population?), which I'm assuming is a data quality issue with the crosswalk file we need to look into further.

Overall the file format was more complex than I had initially assumed, and our conversion process is going to be inherently lossy, as census block boundaries don't entirely align from 1 census to the next.

## Testing Instructions

- Create a plan for a region that is on both our [2010 states list](https://docs.google.com/spreadsheets/d/1NW7P_W23OIFrMGKVvvQpaj2CbaXZfmxBFcgQU92i-08/edit#gid=0) currently on production as well as our [list of 2020 states w/ 2019 ACS data](https://docs.google.com/spreadsheets/d/1khiNAQp4UnfkgfvRSbfjDU9CU2zJrJ0pMRUeF5a4cuM/edit#gid=0), using a region configured w/ '10 data
- Using `scripts/dbshell`, set `ARCHIVED = TRUE` for that region, and create a new '20 region (I used DE for this, but it would be good to test some other states)
- There should be "Convert to 2020" button on your plan, clicking it should open a modal, which upon confirmation should create a copy of your map converted to work w/ 2020 block data

Closes #815

### DIFF
--- a/src/client/components/ConvertMapModal.tsx
+++ b/src/client/components/ConvertMapModal.tsx
@@ -1,0 +1,125 @@
+/** @jsx jsx */
+import React from "react";
+import AriaModal from "react-aria-modal";
+import { connect } from "react-redux";
+import { useHistory } from "react-router-dom";
+import { Box, Button, Flex, Heading, jsx, ThemeUIStyleObject } from "theme-ui";
+
+import { IProject, IUser } from "../../shared/entities";
+import { showConvertMapModal } from "../actions/districtDrawing";
+import { resetProjectState } from "../actions/root";
+import { convertAndCopyProject } from "../api";
+import { showActionFailedToast } from "../functions";
+import { State } from "../reducers";
+import store from "../store";
+import { AuthModalContent } from "./AuthComponents";
+import { Resource } from "../resource";
+
+const style: ThemeUIStyleObject = {
+  footer: {
+    display: "flex",
+    flexDirection: "column",
+    marginTop: 5
+  },
+  header: {
+    mb: 5
+  },
+  heading: {
+    fontFamily: "heading",
+    fontWeight: "light",
+    fontSize: 4
+  },
+  modal: {
+    bg: "muted",
+    p: 5,
+    width: "small",
+    maxWidth: "90vw",
+    overflow: "visible"
+  }
+};
+
+const ConvertMapModal = ({
+  project,
+  user,
+  showModal
+}: {
+  readonly project: IProject;
+  readonly user: Resource<IUser>;
+  readonly showModal: boolean;
+}) => {
+  const history = useHistory();
+
+  const hideModal = () => store.dispatch(showConvertMapModal(false));
+  const attributedName = `${project.name} by ${project.user.name}`;
+  const isLoggedIn = "resource" in user;
+
+  function goToProject(project: IProject) {
+    history.push(`/projects/${project.id}`);
+  }
+
+  return showModal ? (
+    <AriaModal
+      titleId="convert-map-modal-header"
+      onExit={hideModal}
+      initialFocus="#primary-action"
+      getApplicationNode={() => document.getElementById("root") as Element}
+      underlayStyle={{ paddingTop: "4.5rem" }}
+    >
+      <Box sx={style.modal}>
+        {!isLoggedIn ? (
+          <AuthModalContent project={project} />
+        ) : (
+          <React.Fragment>
+            <Box sx={style.header}>
+              <Heading as="h1" sx={style.heading} id="convert-map-modal-header">
+                Copy this map to use 2020 Census data?
+              </Heading>
+            </Box>
+            <Flex
+              as="form"
+              sx={{ flexDirection: "column" }}
+              onSubmit={(e: React.FormEvent) => {
+                e.preventDefault();
+                convertAndCopyProject(project.id)
+                  .then((newProject: IProject) => {
+                    // Need to reset the project state here, so when the redirect kicks in we don't have any
+                    // old state hanging around (undo history, etc.)
+                    store.dispatch(resetProjectState());
+                    goToProject(newProject);
+                  })
+                  .catch(showActionFailedToast);
+              }}
+            >
+              <Box>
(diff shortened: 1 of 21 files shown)


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -400,55 +415,108 @@ export class ProjectsController implements CrudController<Project> {
...
+        ConvertProjectErrors.NO_ACTIVE_REGION
+      );
+    }
+
+    const archivedTopoProperties = await this.getGeoUnitProperties(project.regionConfig.s3URI);
+    const newGeoUnitTopology = await this.getGeoUnitTopology(newRegion.s3URI);
+
+    const csvRows = project.exportToCsv(archivedTopoProperties);
+
+    // CSV export contains every block<->district pair
```
maurizi, line 467: This was an assumption on my part when I started that I'm no longer sure is true.

ddohler: If this were true then you'd expect that `csvRows.length === Object.keys(crosswalk).length` , right? Is that actually true? Or am I misunderstanding?

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -400,55 +415,107 @@ export class ProjectsController implements CrudController<Project> {
...
+    const archivedTopoProperties = await this.getGeoUnitProperties(project.regionConfig.s3URI);
+    const newGeoUnitTopology = await this.getGeoUnitTopology(newRegion.s3URI);
+
+    const oldBlockToDistricts = Object.fromEntries(project.exportToCsv(archivedTopoProperties));
+
+    // CSV export contains every block<->district pair
+    // Crosswalk maps new block -> old in many<->many fashion, specifying percent of old block in new block
+    // Sum district share for each new block and assign block to the district w/ largest share
+
+    // Written in non-functional style for higher perf.
```
maurizi, line 471: I originally had this written using `.reduce(...)` which was *surprisingly* slow compared to this

src/server/src/region-configs/services/region-configs.service.ts
```diff
@@ -7,34 +7,9 @@ import { Repository } from "typeorm";
...
-  private onlyLatestRegionConfigs(regionConfigs: RegionConfig[]): RegionConfig[] {
-    return map(
-      groupBy(regionConfigs, (regionConfig: RegionConfig) => [
-        regionConfig.countryCode,
-        regionConfig.regionCode,
-        regionConfig.name
-      ]),
-      (regionConfigs: RegionConfig[]) => last(sortBy(regionConfigs, "version")) as RegionConfig
-    );
-  }
```
maurizi, line 39: We now _can't_ have multiple `region_config`s that aren't hidden/archived, making this override not necessary

kevinearldenny: One tangentially related thing here - not sure if it's totally necessary, but having a `census_year` field on a RegionConfig denoting whether it contains census data from 2010 or 2020 would definitely have been nice to have during testing.

src/client/components/ConvertMapModal.tsx
```diff
@@ -0,0 +1,125 @@
...
+            >
+              <Box>
+                This will create a copy of <strong>{attributedName}</strong> in your account,
+                converted to use 2020 Census data. You will be able to make changes to the copied
+                version.
+              </Box>
+              <Flex sx={style.footer}>
+                <Button id="primary-action" sx={{ marginBottom: 3 }} type="submit">
+                  Yes, copy to my account
+                </Button>
```
kevinearldenny, line 101: It would be nice to have a spinner to let the user know that a project is being converted. The process took nearly a minute for me testing NC locally, so it seems important to let the user know that the conversion process is active and prevent them from clicking the button again to convert

kevinearldenny (changes requested):
This worked perfectly for NC! All of the blocks appear to have been mapped successfully to their 2020 equivalents:

image

That being said, I did leave two comments of things that would be nice to have here. The spinner when conversion is taking place seems like it's probably important, but perhaps we could address this in a follow on issue

src/client/components/ConvertMapModal.tsx
maurizi: Added a spinner in https://github.com/PublicMapping/districtbuilder/pull/878/commits/8bf022c82cf4b0750cc376423b5127a3c6e0e683

src/server/src/region-configs/services/region-configs.service.ts
maurizi: I think I came to the same conclusion myself partway though this PR, but in practice going forward `archived` and `census_year === "2010"` are going to be synonymous I think, so doing the refactor now didn't seem worth it.

I think we'll definitely want to do that if/when we acquire different kinds of archived regions besides our soon-to-be outdated '10 ones.

kevinearldenny (approved):
Okay, spinner looks good 👍  I also tested with MN and it worked perfectly as well, so I think we should be good to go here

PR #888, opened 2021-08-03
Clean up import error handling

### LINKED ISSUE #846: Request entity too large when importing block equivalency file csv
## Description
When attempting to import a district index file csv into DistrictBuilder, the webpage doesn't respond after loading the data and clicking "Create map". We've identified an error in Rollbar that is `PayloadTooLargeError: request entity too large`.  This was tested in New Jersey and Pennsylvania with Congressional Block Equivency files.

## Steps To Reproduce

1. Go to 'Import page'
2. Click on 'Browse files, select file'
3. Enter the number of districts
4. Click 'Create map'. No response.

## Sample NJ file attached
[nj_congress_bef.zip](https://github.com/PublicMapping/districtbuilder/files/6804154/nj_congress_bef.zip)

maurizi:
## Overview

Imports were returning such large numbers of errors that it was generating `PayloadTooLargeError` errors. It seems this was _mostly_ due to invalid block IDs from using 2020 block equivalency files for (likely exported from another tool, such as Dave's Redistricting) for a state for which we have 2010 data.

I addressed this specific case by marking an excessive number of invalid block IDs w/ the correct state FIPs as file-level error and include the expected census year in returned error.

In addition to that, I _also_ added a cut-off for the number of flags, displaying only the first 1000, and I cleaned up the flags modal to correctly scroll when there are  large number of flags.

### Checklist

### Demo

localhost_3003_import-project

localhost_3003_import-project (1)

## Testing Instructions

- Set up a NJ region using 2010 data (`s3://global-districtbuilder-dev-us-east-1/regions/US/NJ/2020-09-09T14:19:50.625Z/`)
- Attempt to upload this file https://github.com/PublicMapping/districtbuilder/files/6804154/nj_congress_bef.zip

To test the cut-off flags, I commented out the new check for invalid blocks over certain size

Closes #846

### DIFF
--- a/src/server/migrations/1628016841545-region_census_year_defaults.ts
+++ b/src/server/migrations/1628016841545-region_census_year_defaults.ts
@@ -1,0 +1,55 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class regionCensusYear1628016841545 implements MigrationInterface {
+  name = "regionCensusYear1628016841545";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // This is the list of 2020 states that have been processed as-of writing this migration
+    await queryRunner.query(`
+        UPDATE region_config SET census = CASE WHEN s3_uri = ANY('{
+        s3://global-districtbuilder-dev-us-east-1/regions/US/AK/2021-06-18T14:24:11.337Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/AZ/2021-07-01T15:29:22.217Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/CO/2021-06-30T16:20:02.781Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/DC/2021-06-12T00:25:53.517Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/DE/2021-06-11T23:03:36.938Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/GA/2021-06-30T17:12:22.649Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/HI/2021-06-18T15:51:03.204Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/IA/2021-06-12T14:45:27.207Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/ID/2021-06-12T14:48:46.371Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z,
+s3://global-districtbuilder-dev-us-east-1/regions/US/KS/2021-06-12T14:50:54.971Z/,
+s3://global-districtbuilder-dev-us-east-1/regions/US/LA/2021-06-12T14:53:58.291Z/,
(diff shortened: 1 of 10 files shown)


src/manage/package.json
```diff
@@ -33,7 +33,7 @@
     "topojson-server": "3.0.1",
     "topojson-simplify": "3.0.3",
     "tslib": "1.11.0",
-    "typeorm": "0.2.25"
+    "typeorm": "0.2.36"
```
maurizi, line 36: This seems to have caused a regression w/ our `nestjsx/crud` integration, breaking the Org Admin feature toggle. I'll likely back revert this commit if I can't figure it out.

src/server/migrations/1628016841545-region_census_year_defaults.ts
```diff
@@ -0,0 +1,55 @@
...
+s3://global-districtbuilder-dev-us-east-1/regions/US/WY/2021-06-17T20:44:48.259Z/}'::text[])
+        THEN '2020'::region_config_census_enum ELSE '2010'::region_config_census_enum END
+        `);
+    await queryRunner.query(
+      `ALTER TABLE "public"."region_config" ALTER COLUMN "census" SET NOT NULL`
+    );
+    await queryRunner.query(
+      `ALTER TABLE "public"."region_config" ALTER COLUMN "census" SET DEFAULT '2020'`
+    );
+  }
```
kevinearldenny, line 45: When I attempted to import the NJ file provided without explicitly setting a `census` value, it defaulted to 2020 which confused me a bit since it deviated from what the wireframes displayed. Probably won't be an issue in staging / production, but it's worth noting that we need to explicitly set this value for any 2010 census files that we add (if we add any new ones)

src/server/src/districts/controllers/districts.controller.ts
```diff
@@ -15,10 +15,11 @@ import {
   DistrictsImportApiResponse,
   DistrictImportField
 } from "../../../../shared/entities";
-import { FIPS } from "../../../../shared/constants";
+import { FIPS, MAX_IMPORT_ERRORS } from "../../../../shared/constants";
 import { TopologyService } from "../services/topology.service";
 
 import { RegionConfigsService } from "../../region-configs/services/region-configs.service";
+import _ from "lodash";
```
kevinearldenny, line 22:
```suggestion
```

This is producing a lint error for unused variable

kevinearldenny (commented):
This mostly looks good to me, but it might be a good idea to test on staging as well on a `test/` branch to make sure that the error has been resolved since it was previously only occurring on staging for me

src/server/migrations/1628016841545-region_census_year_defaults.ts
maurizi: Once we get the new census data for 2020 in a few weeks, 2010 regions will be basically legacy for us - we'll only be supporting them in readonly archived mode, so I set this default assuming any new regions we add in the near future will be for 2020.

src/manage/package.json
maurizi: I deleted this commit & opened https://github.com/PublicMapping/districtbuilder/issues/891

kevinearldenny (approved):
👍  looks good now

Issue #881, opened 2021-08-02
Deleted maps still display on Organization admin page


dmcglone:
After a user deletes a map, the map is still listed on the Organization admin page. When the admin clicks on the URL for the deleted map, they get a Page Not Found error. The map should no longer be listed as an option to Feature.

maurizi:
@dmcglone I'm not able to replicate this bug on staging - was this on production?

dmcglone:
@maurizi yes this is on production

maurizi:
This was fixed in https://github.com/PublicMapping/districtbuilder/pull/805

PR #892, opened 2021-08-05
Clean up Home Screen project handling

maurizi:
## Overview

In the course of investigating #881 I encountered two issues:
 - Duplicated projects lost their `projectTemplate` reference
 - Archiving a project would cause the page it was on to lose items. Archiving all projects on the page would cause it to revert to the "No projects" state you see when you first join
 
 I fixed the first by creating a new project duplicate endpoint (I explored adding the missing fields to the SQL query in `ProjectsService` but decided this way was less brittle, as we won't need to update it as we add new fields.

### Checklist

## Testing Instructions

- On the Home Screen, duplicate a project that is visible & associated w/ and organization `ProjectTemplate`. The duplicate project should appear on the Organization Admin screen.
- Archive a project from the Home Screen - the list of projects should refresh but maintain your current page

### DIFF
--- a/src/server/src/projects/services/projects.service.ts
+++ b/src/server/src/projects/services/projects.service.ts
@@ -27,13 +27,13 @@ export class ProjectsService extends TypeOrmCrudService<Project> {
   getProjectsBase(): SelectQueryBuilder<Project> {
     return this.repo
       .createQueryBuilder("project")
-      .innerJoinAndSelect("project.regionConfig", "regionConfig")
-      .innerJoinAndSelect("project.user", "user")
-      .leftJoinAndSelect("project.chamber", "chamber")
+      .innerJoin("project.regionConfig", "regionConfig")
+      .innerJoin("project.user", "user")
+      .leftJoin("project.chamber", "chamber")
       .select([
         "project.id",
-        "project.numberOfDistricts",
         "project.name",
+        "project.numberOfDistricts",
         "project.updatedDt",
         "project.createdDt",
         "project.districts",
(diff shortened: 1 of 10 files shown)


src/server/src/projects/services/projects.service.ts
```diff
@@ -67,9 +67,10 @@ export class ProjectsService extends TypeOrmCrudService<Project> {
     userId: string,
     options: AllProjectsOptions
   ): Promise<Pagination<Project>> {
-    const builder = this.getProjectsBase()
-      .addSelect("project.districtsDefinition")
```
kevinearldenny, line 71: It seems like removing this broke the duplicate project functionality - without this line included, districts are not copied when a project is duplicated

kevinearldenny (changes requested):
Districts are not currently being copied when projects are duplicated, which definitely needs to get fixed and should be fairly simple as noted in the comment. 

I'm also noticing that `projectTemplate` doesn't seem to be getting copied with a duplicated project for me either - I explored trying to solve this a bit by explicitly passing in `projectTemplate` to the dto, but this didn't seem to fix it for me.

src/server/src/projects/services/projects.service.ts
maurizi: This has been fixed in a totally different way - I've stopped using `createProject` for duplicate from the Home Screen, and I'm now using a new `duplicate` endpoint that re-fetches data for duplication and so doesn't rely on what is fetched for the Home Screen list.

This was the intent of the PR but I accidentally made this change for the _Project Screen_ duplication option (not ideal that the functionality is duplicated like this) - corrected here https://github.com/PublicMapping/districtbuilder/pull/892/commits/33b557c549660fe10c51ac9546c0a43f05a1dfb4

maurizi:
> I'm also noticing that `projectTemplate` doesn't seem to be getting copied with a duplicated project for me either - I explored trying to solve this a bit by explicitly passing in `projectTemplate` to the dto, but this didn't seem to fix it for me.

This is fixed via the new `duplicate` endpoint, but FYI I believe the problem w/ the old method was `ProjectTemplate` wasn't being queried in `ProjectsService`

kevinearldenny (approved):
Okay, this looks good to me now 👍

PR #894, opened 2021-08-05
Fix maps page slowness

### LINKED ISSUE #885: Community Maps page is too slow
It takes over 25 seconds to load just _4_ maps on the community maps page on staging.

Some investigation reveals the main culprit is the slowness of the `globalProjects` endpoint, in addition we are also making _2_ calls to that API on page load, which is clearly unneeded

#### Notes
 - We don't need the `districts` feature properties for the mini-map, so we could strip them out
 - Converting the geojson into geometries & simplifying them & then converting back to geojson would add database overhead, but would significantly reduce the network transit time at no UX penalty, since the map we're displaying on is so small
 - We could consider not including `districts` as part of the main API endpoint at all, but instead loading it separately for each project (and display a loading spinner in its place)
 
AC:
 - Only 1 call to `globalProjects` endpoint
 - Page performance is improved to load 1 page of maps in under 5s

maurizi:
## Overview

Both the Home screen and the Community Maps screen saw a big hit to performance after adding the mini-maps to them.
In addition, the Community Maps screen was accidentally querying for projects twice, exacerbating the issue.

Since we don't need the `districts` column to be extremely detailed for the mini-maps, we can simplify it server side to greatly reduce the amount of data sent (I saw a nearly 100x improvement in size) at the cost of increased application server time / CPU load.

### Checklist

### Notes

This needs the changes to how duplicate projects are handled in #892 or it would cause Project districts to be simplified every-time they are duplicated - obviously not good.

## Testing Instructions

- Ensure you have a good number of projects visible on your Home screen & the Community Maps screen
- On this branch, time to load both pages should be greatly reduced
- You should observe only one call to the ``globalProjects` endpoint in the Network tab of the Developer tools when the page is loaded

Closes #885


kevinearldenny (approved):
LGTM 👍

PR #897, opened 2021-08-06
Districts cache busting

### LINKED ISSUE #884: Add cache-busting based on RegionConfig version to GeoJSON export
We want to be able to update properties in the TopoJSON layers and have those changes propagated to cached `districts` data.

AC:
 - A new `version` field is added to `Project` which stores the `version` of the associated `RegionConfig` at the time the `districts` column is cached
 - The GeoJSON export endpoint checks the stored version, and if necessary recalculates the districts geojson & saves it to the database

maurizi:
## Overview

Adds cache-busting to the `districts` column by tracking the `version` of `RegionConfig` on the new `Project.regionConfigVersion` field, and regenerating the column when they are out-of-sync.

### Checklist

### Notes

We only make use of district feature geometries outside the project screen for the most part, so it was relatively easy to contain needing to regenerate the `districts` column to the use of the GeoJSON export endpoint, with a few exceptions:
 - I changed the definition of "completed" on the Community from having no population in the unassigned district to having no features in the unassigned district (you could have only blocks w/ no population left before and we'd consider it "complete").
 - The organization admin CSV map export won't recalculate districts; if this is desired I'll open a follow-on issue to tackle this separately

## Testing Instructions

I created a PA region w/o political data (`'s3://global-districtbuilder-dev-us-east-1/regions/US/PA/2021-02-08T18:07:18.957Z/'`) and created a map / assigned some districts.

_Then_ I updated my PA `region_config` row to point to updated data w/ political data (`'s3://global-districtbuilder-dev-us-east-1/regions/US/PA/2020-09-09T19:03:58.174Z/'`) and updated the `version` of the region_config.

Subsequently, loading the map you created previously should regenerate the `districts` column & show political data.

Closes #884

### DIFF
--- a/src/server/src/project-templates/controllers/project-templates.controller.ts
+++ b/src/server/src/project-templates/controllers/project-templates.controller.ts
@@ -102,6 +102,9 @@ export class ProjectTemplatesController {
         `User does not have admin privileges for organization: ${org.id}`
       );
     }
+    // The associated 'districts' column may be out-of-date if the RegionConfig has been updated
+    // We don't make an attempt to regenerate those here, and accept that we will return whatever data was
+    // available when the user last updated their project
     const projectRows = await this.service.findAdminOrgProjectsWithDistrictProperties(slug);
     const regionURIs = new Set(projectRows.map(row => row.regionS3URI));
     const topoLayers: { [s3uri: string]: GeoUnitTopology | GeoUnitProperties } = {};
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -37,7 +37,12 @@ import isUUID from "validator/lib/isUUID";
 import { Pagination } from "nestjs-typeorm-paginate";
 
 import { MakeDistrictsErrors, ConvertProjectErrors } from "../../../../shared/constants";
-import { DistrictsDefinition, ProjectId, PublicUserProperties } from "../../../../shared/entities";
+import {
+  DistrictsDefinition,
+  ProjectId,
+  PublicUserProperties,
+  UserId
+} from "../../../../shared/entities";
 import { ProjectVisibility } from "../../../../shared/constants";
 import { GeoUnitTopology } from "../../districts/entities/geo-unit-topology.entity";
 import { TopologyService } from "../../districts/services/topology.service";
@@ -193,8 +198,7 @@ export class ProjectsController implements CrudController<Project> {
     @ParsedRequest() req: CrudRequest,
     @ParsedBody() dto: UpdateProjectDto
   ) {
-    // @ts-ignore
-    const existingProject = await this.service.findOne({ id }, { relations: ["regionConfig"] });
+    const existingProject = await this.getProjectWithDistricts(id, req.parsed.authPersist.userId);
     if (dto.lockedDistricts && existingProject?.numberOfDistricts !== dto.lockedDistricts.length) {
       throw new BadRequestException({
         error: "Bad Request",
@@ -343,6 +347,30 @@ export class ProjectsController implements CrudController<Project> {
     return project;
   }
 
+  // Helper for obtaining a project for a given project request, throws exception if not found
+  async getProjectWithDistricts(id: ProjectId, userId: UserId): Promise<Project> {
+    if (!isUUID(id)) {
+      throw new NotFoundException(`Project ${id} is not a valid UUID`);
+    }
+    // Not using 'getProject' because we need to select the 'districts' column
+    // Unauthenticated access is allowed for individual projects if they are
+    // visible or published, and not archived.
+    const commonFilter = { id, archived: false };
+    const project = await this.service.findOne({
+      where: [
+        { ...commonFilter, user: { id: userId } },
+        { ...commonFilter, visibility: ProjectVisibility.Published },
+        { ...commonFilter, visibility: ProjectVisibility.Visible }
+      ] as const,
+      loadEagerRelations: false,
+      relations: ["regionConfig"]
+    });
+    if (!project) {
+      throw new NotFoundException(`Project ${id} not found`);
+    }
+    return project;
+  }
+
   // Helper for obtaining a topology for a given S3 URI, throws exception if not found
   async getGeoUnitTopology(s3URI: string): Promise<GeoUnitTopology> {
     const geoCollection = await this.getGeoUnitProperties(s3URI);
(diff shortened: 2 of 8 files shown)


src/server/src/project-templates/controllers/project-templates.controller.ts
```diff
@@ -102,6 +102,9 @@ export class ProjectTemplatesController {
         `User does not have admin privileges for organization: ${org.id}`
       );
     }
+    // The associated 'districts' column may be out-of-date if the RegionConfig has been updated
+    // We don't make an attempt to regenerate those here, and accept that we will return whatever data was
+    // available when the user last updated their project
```
maurizi, line 107: @dmcglone Just wanted to call your attention to this - we can tackle this in a follow-on issue if you think this is a problem.

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -414,6 +431,24 @@ export class ProjectsController implements CrudController<Project> {
...
+        numberOfDistricts: project.numberOfDistricts,
+        districtsDefinition: project.districtsDefinition
+      });
+
+      // Note we don't wait for save to return, and we throw away it's result
+      void this.service.save({
+        ...project,
+        districts,
+        regionConfigVersion: project.regionConfig.version
+      });
```
maurizi, line 447: This felt a little weird (updating the database in a `GET` request), but also it feels wasteful to regenerate this and _not_ save it. Thoughts?

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -593,17 +628,17 @@ export class ProjectsController implements CrudController<Project> {
...
+    @ParsedRequest() req: CrudRequest,
+    @Param("id") projectId: ProjectId
+  ): Promise<Project> {
     const planScoreToken = process.env.PLAN_SCORE_API_TOKEN || "";
-    const project = await this.service.findOne(projectId, { relations: ["regionConfig"] });
-    const geojson = project && {
-      ...project.districts,
-      features: project.districts.features.filter(f => f.id !== 0)
-    };
+    const geojson = await this.exportGeoJSON(req, projectId);
```
maurizi, line 636:
```suggestion
    const districts= await this.exportGeoJSON(req, projectId);
    const geojson = districts && {
      ...districts,
      features: districts.features.filter(f => f.id !== 0)
    };
```

src/server/src/project-templates/controllers/project-templates.controller.ts
dmcglone: Thanks @maurizi, clarification question:
Does this mean that a template stored in 2010 geometries will not automatically update to 2020? I have a couple of templates stored in 2010 data for a current client. Obviously won't be an issue once we switch over but I want to make sure those aren't broken, or if I need to re-generate them, I can do that as well...

maurizi: > Does this mean that a template stored in 2010 geometries will not automatically update to 2020? I have a couple of templates stored in 2010 data for a current client. Obviously won't be an issue once we switch over but I want to make sure those aren't broken, or if I need to re-generate them, I can do that as well...

A few things:
 - We don't have any mechanisms to _automatically_ update 2010 plans -> 2020 plans; only the "Convert to 2020" button in this PR: https://github.com/PublicMapping/districtbuilder/pull/878 , which only happens when a user clicks on it
 - We don't have _any_ mechanisms at all for converting Templates from 2010 -> 2020; we'll need to do this manually when we rollout 2020 replacement regions by updating the database.
 
 But neither of those are what I was trying to bring your attention to here, which has nothing to do with 2010 -> 2020 updates but instead concerns data updates that are not census geometry changes (e.g. we get 2020 political data for a state that previously only had 2016 political data & the current data and the new data both uses 2020 census geometry).
 
 This is for the Organization admin Map CSV export (the one that exports all plans for an Organization w/ 1 row per district).
 For that export, the way the code is written now, if we add '20 political data for a state, or make other demographic updates, those changes won't be reflected in the export for _any already existing maps_ from before the data updates happened. If a user opens one of those maps they would _then_ see the new political data, and then it would show up in the export.

dmcglone: Got it, thanks for clarifying. That issue is of pretty minor concern.

ddohler (approved):
Works great! I had never seen the expected behavior on `develop`, so I followed the testing instructions on `develop` and saw this when reloading the project after having added political data:
Screen Shot 2021-08-06 at 1 23 13 PM

And then I switched branches and restarted the server, and got this 🎉 
Screen Shot 2021-08-06 at 1 39 10 PM

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -343,6 +347,30 @@ export class ProjectsController implements CrudController<Project> {
...
+    }
+    // Not using 'getProject' because we need to select the 'districts' column
+    // Unauthenticated access is allowed for individual projects if they are
+    // visible or published, and not archived.
+    const commonFilter = { id, archived: false };
+    const project = await this.service.findOne({
+      where: [
+        { ...commonFilter, user: { id: userId } },
+        { ...commonFilter, visibility: ProjectVisibility.Published },
+        { ...commonFilter, visibility: ProjectVisibility.Visible }
```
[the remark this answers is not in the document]
maurizi: This is due to how TypeORM lets us specify AND / OR in conditionals - anything in an array gets joined with an OR, anything in an object gets joined with AND.

I'm not sure how to specify what I really want here, which is `commonFilter && (some_condition || other_condition || third_condition)` so I settled for `(some_condition && commonFilter) || (other_condition && commonFilter) || (third_condition && commonFilter)` which is redundant but still works - perhaps TypeORM will let us do what I really want and I just don't know how, I wouldn't be surprised.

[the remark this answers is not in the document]
ddohler: Oh interesting! I did a quick docs skim and it looks like you [maybe need to use Brackets](https://github.com/typeorm/typeorm/blob/master/docs/select-query-builder.md#adding-where-expression)?

maurizi:
@ddohler FYI I realized before merging this that there is a small data migration I wanted to include: https://github.com/PublicMapping/districtbuilder/pull/897/commits/66ea3af62025e01c6804b176ca98f70542f15ca5

It's pretty small, so I don't think you need to go through a full testing / review cycle.

PR #904, opened 2021-08-10
Style cleanup

### LINKED ISSUE #865: Styling issues in expanded metrics viewer
image

Issues
 - the table header has an extra column (for district actions) that shouldn't be visible when expanded, as we hide the corresponding "..." / 🔒 menus when expanded
 - The map area / project header buttons should not be visible on the right side of the page

jfrankl:
Style cleanup across application.

Screen Shot 2021-08-10 at 1 09 31 PM
Screen Shot 2021-08-10 at 1 12 41 PM
Screen Shot 2021-08-10 at 1 11 58 PM
Screen Shot 2021-08-10 at 1 11 37 PM
Screen Shot 2021-08-10 at 1 11 24 PM
Screen Shot 2021-08-10 at 1 10 03 PM

## Testing Instructions
- This PR makes a number of style fixes across the application, be sure to take a look around in every section, specifically
- The home screen (map list)
- Project screen with the sidebar expanded or not expanded
- Evaluate mode, Competitiveness view
- The find button

Fixes #865

### DIFF
--- a/src/client/components/Field.tsx
+++ b/src/client/components/Field.tsx
@@ -113,7 +113,7 @@ export function SelectField<D, R>({
             {...selectProps}
             id={field.toString()}
             aria-describedby={description ? `description-${field.toString()}` : undefined}
-            sx={{ borderColor: hasErrors ? "warning" : undefined }}
+            sx={{ borderColor: hasErrors ? "warning" : undefined, width: "100%" }}
           >
             {children}
           </Select>
--- a/src/client/components/HomeScreenProjectCard.tsx
+++ b/src/client/components/HomeScreenProjectCard.tsx
@@ -1,5 +1,5 @@
 /** @jsx jsx */
-import { Box, Flex, Heading, jsx } from "theme-ui";
+import { Box, Flex, Heading, jsx, Text } from "theme-ui";
 import { useHistory } from "react-router-dom";
 
 import { IProject } from "../../shared/entities";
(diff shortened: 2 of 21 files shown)


kevinearldenny (commented):
This mostly looks great! That being said, I do have a few comments (by page below)

### Home screen

- Could we make the maps a little bit bigger? And perhaps apply a bit more padding between them and the project title? The screen feels a bit out of balance to the left currently

### Project screen

#### Sidebar
image
- I think the expand button looks a little weird in it's new place. Could we move it back over to the corner when not expanded?

#### Find menu

- It looks a little weird not having "Find" written next to the icon IMO

jfrankl:
Thanks for taking a look @kevinearldenny.

I made the maps slightly larger and reduced some of the padding around them. I also centered this content and added a little more spacing. I'm trying to strike a balance between wanting to show off the maps, but also make it easy for users to see several projects and quickly scroll down the page. The larger the maps, the longer the page. I think this is looking better.

Screen Shot 2021-08-11 at 7 31 58 AM

I hear you about the "Find" text for the search. The reason I'm removing that label is to preserve space in the Map Header. I want to make sure the application works well on a variety of screen sizes, and I think this is a good tradeoff. I expect the search will still be discoverable enough given its relative importance, but I'm open to reconsidering this in the future if we learn otherwise. In the meantime, I added a tooltip and slightly cleaned up the hover state.

Screen Shot 2021-08-11 at 7 32 52 AM

This choice for the Expand Sidebar button location is so when the sidebar is expanded but is too wide to fit on a user's screen, they can still see the button so they can collapse the sidebar. Before I made this change, a user would need to scroll to the right before they could see the button.
Screen Shot 2021-08-11 at 7 29 47 AM

PR #907, opened 2021-08-10
Fix creating projects

maurizi:
## Overview

Adding `Project.region_config_version` as part of #884 as a not-null field unintentionally broke project creation.

## Testing Instructions

On `develop` you can't create new projects. On this branch you can.


kevinearldenny (approved):
LGTM 👍

PR #914, opened 2021-08-11
Create auto scaling group and stand up new app service

### LINKED ISSUE #879: Create ECS cluster and migrate application
In order to avoid the 30 GB memory limit for Fargate while allowing us to include new census data for all 50 states, we will need to temporarily run the application on ECS backed by EC2. This will allow us to run the application on a 64 GB EC2 instance and allow us to easily allocate at least 32 GB of RAM for loading TopoJSON and running the application.

This is a temporary measure until we are able to partition the TopoJSON data such that we can comfortably run instances of the application well under the 30 GB memory limit for Fargate. At this point we should revert the changes introduced for this issue, migrate the application back to Fargate, and delete the new ECS cluster.

rbreslow:
## Overview

I made the following changes to support allocating over 30 GB of memory to the app server for the updated Census:

- Create an Auto Scaling group and launch template based on the [Terraform 0.12 branch](https://github.com/azavea/terraform-aws-ecs-cluster/tree/feature/jrb%2Fterraform-0.12) of our ECS cluster module.
- Make changes to work with the existing ECS cluster and security group resources.
- Stand up a second ECS service for the application server that uses the [EC2 launch type](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_types.html#launch-type-ec2).
- Make use of the [`awsvpc` network mode](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-networking-awsvpc.html) to preserve the existing target group, allowing a zero-downtime deployment. 

In previous ECS clusters where we've managed container instances, we've used `bridge` networking. `awsvpc` is the same network mode that we use with Fargate, which makes it less complicated to bridge two launch types, but also means we must have a one-to-one mapping of application servers to container instances.

### Checklist

## Testing Instructions

Double check that Terraform is up-to-date:

```console
export GIT_COMMIT=56dc64e
./scripts/infra plan

. . .

No changes. Infrastructure is up-to-date.
```

See that both ECS services are up and healthy:

<details>

image

image

</details>

Verify that, aside from 4198e6c, there are **only additions** in the plan for production:

```console
export DB_SETTINGS_BUCKET=districtbuilder-production-config-us-east-1
export DB_DEPLOYMENT_ENVIRONMENT=production
export GIT_COMMIT=c006344
./scripts/infra plan

. . .

Plan: 9 to add, 1 to change, 1 to destroy.

```

Resolves #879

### DIFF
--- a/deployment/terraform/config.tf
+++ b/deployment/terraform/config.tf
@@ -1,6 +1,6 @@
 provider "aws" {
   region  = var.aws_region
-  version = "~> 3.6.0"
+  version = "~> 3.10.0"
 }
 
 terraform {
--- a/deployment/terraform/container_service.tf
+++ b/deployment/terraform/container_service.tf
@@ -1,0 +1,187 @@
+#
+# Auto Scaling resources
+#
+data "template_cloudinit_config" "container_instance_cloud_config" {
+  gzip          = true
+  base64_encode = true
+
+  part {
+    content_type = "text/cloud-config"
+    content = templatefile("${path.module}/cloud-config/base-container-instance.yml.tmpl", {
+      ecs_cluster_name = aws_ecs_cluster.app.name
+    })
+  }
+}
+
+# Pull the image ID for the latest Amazon ECS-optimized Amazon Linux 2 AMI
+# https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html#al2ami
+data "aws_ssm_parameter" "ecs_image_id" {
+  name = "/aws/service/ecs/optimized-ami/amazon-linux-2/recommended/image_id"
+}
+
+data "aws_ami" "ecs_ami" {
+  owners = ["self", "amazon", "aws-marketplace"]
+
+  filter {
+    name   = "image-id"
+    values = [data.aws_ssm_parameter.ecs_image_id.value]
+  }
+}
+
+resource "aws_launch_template" "container_instance" {
+  block_device_mappings {
+    device_name = data.aws_ami.ecs_ami.root_device_name
+
+    ebs {
+      volume_type = var.container_instance_root_block_device_type
+      volume_size = var.container_instance_root_block_device_size
+    }
+  }
+
+  credit_specification {
+    cpu_credits = "standard"
+  }
+
+  disable_api_termination = false
+
+  name_prefix = "lt${title(var.environment)}ContainerInstance-"
+
+  iam_instance_profile {
+    name = aws_iam_instance_profile.ecs_container_instance_role.name
+  }
+
+  image_id = data.aws_ami.ecs_ami.image_id
+
+  instance_initiated_shutdown_behavior = "terminate"
+  instance_type                        = var.container_instance_type
+  key_name                             = var.aws_key_name
+  vpc_security_group_ids               = [aws_security_group.app.id]
+  user_data                            = data.template_cloudinit_config.container_instance_cloud_config.rendered
+
+  monitoring {
+    enabled = true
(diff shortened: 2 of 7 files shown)


deployment/terraform/config.tf
```diff
@@ -1,6 +1,6 @@
 provider "aws" {
   region  = var.aws_region
-  version = "~> 3.6.0"
+  version = "~> 3.10.0"
```
rbreslow, line 3: Required to make use of the `aws_ec2_instance_type` data source.

See: https://github.com/hashicorp/terraform-provider-aws/blob/706a16a54181491ffd4f193ac366ca9760b0cc9b/CHANGELOG.md#3100-october-09-2020

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,187 @@
...
+
+    ebs {
+      volume_type = var.container_instance_root_block_device_type
+      volume_size = var.container_instance_root_block_device_size
+    }
+  }
+
+  credit_specification {
+    cpu_credits = "standard"
+  }
```
rbreslow, line 43: Only useful for [burstable performance instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/burstable-performance-instances-how-to.html).

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,187 @@
...
+
+  instance_initiated_shutdown_behavior = "terminate"
+  instance_type                        = var.container_instance_type
+  key_name                             = var.aws_key_name
+  vpc_security_group_ids               = [aws_security_group.app.id]
+  user_data                            = data.template_cloudinit_config.container_instance_cloud_config.rendered
+
+  monitoring {
+    enabled = true
+  }
```
rbreslow, line 63: Enables [detailed monitoring](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-cloudwatch-new.html).

deployment/terraform/container_service.tf
```diff
@@ -0,0 +1,187 @@
...
+
+  name = "asg${title(var.environment)}ContainerInstance"
+
+  launch_template {
+    id      = aws_launch_template.container_instance.id
+    version = "$Latest"
+  }
+
+  health_check_grace_period = var.container_instance_asg_health_check_grace_period
```
rbreslow, line 78: The default health checks for an Auto Scaling group are for EC2 status only.

See: https://docs.aws.amazon.com/autoscaling/ec2/userguide/healthcheck.html

rbreslow:
> My impression is that we're going to apply the changes to the ALB that will cut over the application separately, and then apply changes to remove the old Fargate stack (or at least scale it down to 0 tasks) until we figure out a way to migrate back after that.

The cool thing is there's no migration either way since this is just an ECS service that shares the same load balancing resources. There are no changes to the ALB or any other existing resource. After running the release pipeline, there'll be an identical set of tasks running on EC2 instances. All we need to do is crank the `desired_count` of the Fargate service down to `0`, and then we only have tasks running on EC2.

> We should definitely practice this cutover in staging first and verify that we can use the app without issue. I figure that's part of the plan but I want to make sure we were planning on it.

Wilco. We'll just see the Fargate-based targets drop out of the target group. So long as the EC2-based tasks are running, there will still be healthy targets.

rbreslow:
> Excellent work! I just reviewed #919 and the procedure in there explains everything. I see that the app is already attached to the ALB, so all we'll need to do is turn off the old Fargate tasks. Great thinking.

Ah, cool, thank you 😄!

PR #911, opened 2021-08-11
Fix community maps page

maurizi:
## Overview

It seems like some of our NJ data has invalid polygon data in it due to rounding points of a very small hole such that all corners are exactly equal.

This tripped up the simplification library, which caused the page to not load.

We probably can't find/fix every such data issue like this, so I made the simplification a bit more granular and have it just skip past any polygons it has issues with (and return them unsimplified).

Hopefully all the problematic polygons are similarly small, so it won't be an issue that we don't simplify them.

### Checklist

- [x] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

### Notes

The specific MultiPolygon throwing off `simplify`:
```
{
  "id":9,
  "type":"Feature",
  "geometry":{
    "type":"MultiPolygon",
    "coordinates":[
     [
        ...
     ],
      [
        [
          [
            -74.33104881808818,
            41.186675416504166
          ],
          [
            -74.33104881808818,
            41.186675416504166
          ],
          [
            -74.33104881808818,
            41.186675416504166
          ],
          [
            -74.33104881808818,
            41.186675416504166
          ]
        ]
      ]
    ]
  },
  "properties":{
    "voting":{},
    "contiguity":"non-contiguous",
    "compactness":0,
    "demographics":{
      "asian":32440,
      "black":73973,
      "other":8070,
      "white":359628,
      "hispanic":195294,
      "population":650491
    }
  }
}
```

## Testing Instructions

Create a new map for NJ using the region set up on staging (`s3://global-districtbuilder-dev-us-east-1/regions/US/PA/2020-09-09T19:03:58.174Z/`). I simply assigned every county to the first district.

On `develop`, loading the mini-map for this project will fail.
On this branch, it will succeed and log a debug message to the console.

Closes #910

PR #923, opened 2021-08-12
Fix simplification regresssion

maurizi:
## Overview

The previous fix to simplification in #911 caused a regression in performance

Instead of simplifying on a polygon-by-polygon basis, it seems we weren't actually returning simplified data at all anymore

Cleaning the data fixed the one case we knew of invalid polygon data, though I left the try/catch in place in case new data causes a similar issue.

### Notes

## Testing Instructions

- On `develop`, look at response size of the `globalProjects` endpoint in network tab of the developer tools, and compare it to this branch. You should see a reduction in size.

### DIFF
--- a/src/server/src/projects/services/projects.service.ts
+++ b/src/server/src/projects/services/projects.service.ts
@@ -3,6 +3,7 @@ import { InjectRepository } from "@nestjs/typeorm";
 import { TypeOrmCrudService } from "@nestjsx/crud-typeorm";
 import simplify from "@turf/simplify";
 import { Repository, SelectQueryBuilder, DeepPartial } from "typeorm";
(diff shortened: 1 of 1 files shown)


kevinearldenny (changes requested):
Hmmm, somehow the simplification of the geojsons is causing the [district bounds computation](https://github.com/PublicMapping/districtbuilder/blob/a5d1e36edc263d98fd9797dc7c612c5722d4b9f5/src/client/components/map/ProjectDistrictsMap.tsx#L56) to fail for the mini-maps:

image

kevinearldenny (approved):
The previous error is no longer occurring after you added the filter for empty coordinate arrays. LGTM 👍

PR #924, opened 2021-08-12
Fix communty maps region filter

### LINKED ISSUE #922: Community Maps page state filter can incorrectly show no results if not on first page
Steps to reproduce:
 - Go to Community maps page
 - Go to the 2nd page of results
 - Use the state filter to select a state for which there  are results but is only 1 page
 
 Expected:
 You see the first page of results for the state you selected
 
 Actual:
 You see a "No results" page

maurizi:
## Overview

 - Prevents showing "No results" incorrectly when setting the region filter
 - Eliminates duplicate API query when loading the page with a region filter

### Checklist

## Testing Instructions

- Load the community maps page, and navigate to page 2, then change the region filter to a state with only 1 page worth of results
- Refresh the page and check the network tab to see the `globalProjects` endpoint isn't called twice

Closes #922

### DIFF
--- a/src/client/reducers/projects.ts
+++ b/src/client/reducers/projects.ts
@@ -138,7 +138,11 @@ const projectsReducer: LoopReducer<ProjectsState, Action> = (
         ? loop(
             {
               ...state,
-              globalProjectsRegion: action.payload || null
+              globalProjectsRegion: action.payload || null,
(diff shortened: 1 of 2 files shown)


kevinearldenny (approved):
LGTM 👍

PR #926, opened 2021-08-12
Style cleanup for paintbrush size panel

jfrankl:
## Overview

Minor style cleanup for paintbrush size panel.

### Checklist

### Demo

Screen Shot 2021-08-12 at 11 57 06 AM

### Notes

To reduce the size of this panel (so it wouldn't cover as much of the map), I made the following changes:

- Used smaller text
- Reduced slider width
- Reduced panel padding

## Testing Instructions

- Switch to paintbrush drawing mode
- The panel should look similar to the attached screenshot


maurizi (approved):
👍

PR #928, opened 2021-08-12
Properly handle Projects with a null districts column

maurizi:
## Overview

Some of our projects on production are from before we were caching 'districts', and don't have a precomputed districts column for us to use.

Gracefully handle these cases in by simply not displaying a mini-map in the frontend and skipping simplification in the backend.

### Checklist

### Demo

localhost_3003_

## Testing Instructions

- Null out the `districts` column of your projects (I simply ran `UPDATE project SET districts = NULL` in a SQL shell).
- On `develop` the Home Screen will crash
- On this branch it should load, but not display the mini-map


ddohler (approved):
Works great!

PR #938, opened 2021-08-13
Increase ECS memory and fix health check

maurizi:
## Overview

This fixes two separate issues that prevented us from starting to rollout new 2020 data:
 - The new EC2-based deployment strategy was accidentally referencing the old container image, which was limited to 30Gb of memory due to Fargate compatibility. Fixed in https://github.com/PublicMapping/districtbuilder/commit/312ea4c61ce12dfff125ad135f0481da92e165ae
 - We refactored topology service to load archived regions before loading active regions, but this broke in staging, which indirectly ended up calling `Object.freeze` on the `_layers` private member variable of `TopologyService` as part of the health check, which then broke loading topologies. Fixed in https://github.com/PublicMapping/districtbuilder/commit/8b97b0308b551f0213c463a73d34479c8ed3783b

### Checklist

## Testing Instructions

This has already been tested on staging, but to test locally https://github.com/PublicMapping/districtbuilder/pull/938/commits/8b97b0308b551f0213c463a73d34479c8ed3783b you could exercise the `healthcheck` endpoint while archived TopoJSON layers are still loading, which on `develop` will prevent non-archived layers from ever loading.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed export GeoJSON endpoint cache busting [#940](https://github.com/PublicMapping/districtbuilder/pull/940)
+- Fixed EC2-based ECS setup to use correct container [#938](https://github.com/PublicMapping/districtbuilder/pull/938)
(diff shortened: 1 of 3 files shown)


rbreslow (approved):
👍 with a stipulation: I haven't been able to test the changes to the topology service myself. Mike and I discussed over a DM that tasks were hanging at 4/38 topology layers loaded. After Mike applied the fix, tasks were able to start up.

PR #940, opened 2021-08-13
Fix export geojson endpoint

maurizi:
## Overview

This is a regression introduced by #897 

Date objects are never equal to each other, so we were always recreating the geojson, which made it impossible to load projects for archived regions

### Checklist

### Demo

localhost_3003_ (3)

## Testing Instructions

- Create a project for a 2010 region
- Archive the region and add a new 2020 region for the same state. You should be able to convert your project to use the new 2020 region.


ddohler (approved):
Discussed with @maurizi , the code looks good, but it'll be easier to test on staging so we're going to merge as-is and test there.

Issue #939, opened 2021-08-13
Handle I/O failures in TopologyService more gracefully


maurizi:
We have occasionally seen errors like the following logged during application startup:

```
[Nest] 28   - 08/13/2021, 1:42:07 AM   [TopologyService] Unable to download static files from for s3://global-districtbuilder-dev-us-east-1/regions/US/NY/2020-09-09T20:34:33.802Z/: SyntaxError: Unexpected end of JSON input +29286ms
```

Unfortunately after experiencing an error like during application startup, the application will never retry downloading that layer, and will never become healthy, and it will take until the health check grace period ends for it to be cycled out.

I could see fixing this issue in one of two ways:
 - Retry failed layer requests (a set number of times? indefinitely until the health check grace period ends?)
 - Immediately shutdown the application after a failed TopoJSON layer load, so a new service will be started in it's place

maurizi:
Closed by #942

PR #942, opened 2021-08-13
Utilize all available memory on a container instance

rbreslow:
## Overview

Even with 64 GB of memory on the `r5.2xlarge`, we can't load the new regions configured for production. This pull request changes the default `container_instance_type` to an `r5.4xlarge`, removes the hard CPU limit, and expands the hard memory limit to utilize the EC2 instance as possible. 

See: https://azavea.slack.com/archives/C6UD5FG3T/p1628872715113500

**Update:**

A lot has transpired since I opened this pull request, but I think its purpose is still genuine: to remove the limitations that prevented us from scaling. We determined a few, interesting things:

1) Despite removing CPU and memory limitations, and placing the task on an instance with 128 GB of memory, the task was dying once it hit ~17 GB of memory.
2) I found a few posts on the internet, [like this Stack Overflow question](https://stackoverflow.com/questions/67280057/fatal-error-scavenger-semi-space-copy-allocation-failed), where other folks were running into garbage collection errors at the same memory usage as us (~17 GB).
3) I had a feeling that this was a kernel limitation vs. anything on the application side for a few reasons:
   1) We're storing a **very large** data set in memory.
   2) Cole shared a [blog post](https://deepu.tech/memory-management-in-v8/) on V8 memory management that said:
      > **Large object space:** This is where objects which are larger than the size limits of other spaces live. Each object gets its own [`mmap`](https://en.wikipedia.org/wiki/Mmap)'d region of memory. Large objects are never moved by the garbage collector.
   3) 17 GB is an odd value. It's not a limitation for anything relating to V8 itself. And it didn't feel right in the context of things like map key limits, etc.
4) After tuning `vm.max_map_count` to an arbitrarily high value, the application was able to load all the new census data, consuming a maximum of around 27 GB of memory.

After all this, I persisted `vm.max_map_count` as a less-arbitrary value and scaled us back down to a 64 GB instance type. We're still not in the green, as we've been running into new issues with our database driver (e.g., #950), but this pull request puts us in a much better state.

### Checklist

## Testing Instructions

This branch is what @maurizi and I used to deploy the new census data, so we're testing the Terraform changes in production as we speak.

Using the last SHA we deployed to staging, make sure you can apply the Terraform plan:

<details>

```console
export GIT_COMMIT=67d842f
./scripts/infra plan
./scripts/infra apply
```
</details>

Open a shell on one of the container instances and see that `vm.max_map_count` is much higher than the default of `65536`:

<details>

```console
ssh districtbuilder-stg
[ec2-user@ip-10-0-0-38 ~]$ ssh 10.0.1.96
[ec2-user@ip-10-0-1-96 ~]$ cat /proc/sys/vm/max_map_count
192512
```

</details>

See that application health checks are passing:

<details>

```console
http https://app.staging.districtbuilder.org/healthcheck/
```

TBD.

</details>

See that the ECS service is stable:

<details>

TBD.

</details>


deployment/terraform/app.tf
```diff
@@ -156,7 +156,7 @@ resource "aws_ecs_task_definition" "app" {
   container_definitions = templatefile("${path.module}/task-definitions/app.json.tmpl", merge(
     local.shared_app_task_def_template_vars,
     {
-      max_old_space_size = local.fargate_app_memory * 0.75
+      max_old_space_size = local.fargate_app_memory * var.max_old_space_size_scale_factor
```
rbreslow, line 159: I parameterized this so we're not defining the same constant in multiple places.

deployment/terraform/task-definitions/app.json.tmpl
```diff
@@ -2,7 +2,6 @@
   {
     "name": "app",
     "image": "${image}",
-    "cpu": 0,
```
rbreslow, line 5: This is just clean up. Setting `cpu` to `0` at the container level is essentially the same as leaving it undefined. 

See: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment

ddohler (commented):
When running the `terraform` commands in the testing instructions, I see a much different output than what's described here (`Plan: 1 to add, 2 to change, 1 to destroy.`). However, I also see a couple `r5.4xlarge` instances already running in AWS, which is different than what the testing instructions say to expect, so there may have been some intermediate development happening that has changed the state of staging compared to this branch.

I had a couple comments that I think may benefit from further discussion, but I think they're basically tangential to getting this deployed. I would have approved except for the discrepancy between my Terraform output and what's described in the testing instructions; I want to make sure that's not unexpected.

maurizi:
I pushed this as a test branch to let Jenkins deploy it - if it allows us to load all of the 2020 data on staging I'd like to deploy it ASAP so we can do the same on production - and address @ddohler's comments after the fact.

rbreslow:
> I had a couple comments that I think may benefit from further discussion, but I think they're basically tangential to getting this deployed. I would have approved except for the discrepancy between my Terraform output and what's described in the testing instructions; I want to make sure that's not unexpected.

As long as the Terraform output makes sense for reconciling an intermediary deployment, I'm OK with that. Could you share what you see in a `<details>` block?

I want to continue the conversation, but I'm also mindful of the need to get _something_ out to remove resource limitations. If @maurizi can verify that this allows us to load the new data on staging, I'd like to get a 👍 while still following up and addressing all of your comments.

maurizi:
You can see the output here, it made sense to me (keeping the EC2 instance which was already a 128Gb instance, but updating the container):

http://urbanappsci.internal.azavea.com/view/districtbuilder/job/PublicMapping/job/districtbuilder/job/test%252Ffeature%252Fjrb%252Frm-task-hard-limits/1/console

Unfortunately somehow that _wasn't_ enough to load all of our 73 regions:
image

I'm going to try increasing the `max_old_space_size` per Derek's comment and see if that helps us.

We also seem to be much more consistently encountering errors downloading smaller layers (https://github.com/PublicMapping/districtbuilder/issues/939) so I'm going to work on a fix for that as well.

deployment/terraform/container_service.tf
```diff
@@ -157,28 +157,31 @@ data "aws_ec2_instance_type" "container_instance" {
...
   # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size
-  cpu    = var.container_instance_app_cpu
+  cpu    = local.container_instance_app_cpu
   memory = local.container_instance_app_memory
 
   container_definitions = templatefile("${path.module}/task-definitions/app.json.tmpl", merge(
     local.shared_app_task_def_template_vars,
     {
-      max_old_space_size = local.container_instance_app_memory * 0.75
+      max_old_space_size = local.container_instance_app_memory * var.max_old_space_size_scale_factor
```
maurizi, line 184: In order to play w/ setting `max_old_space_size_scale_factor` I had to add the following change (otherwise Node.js complained that the value wasn't an integer):
```suggestion
      max_old_space_size = floor(local.container_instance_app_memory * var.max_old_space_size_scale_factor)
```

deployment/terraform/container_service.tf
```diff
@@ -157,28 +157,27 @@ data "aws_ec2_instance_type" "container_instance" {
...
 resource "aws_ecs_task_definition" "app_container_instance" {
-  family       = "${var.environment}App_EC2LaunchType"
-  network_mode = "awsvpc"
-  # These are hard limits of CPU units and memory, specified at the task level.
+  family                   = "${var.environment}App_EC2LaunchType"
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["EC2"]
+  # The hard limit of memory to present to the task.
   # https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size
-  cpu    = var.container_instance_app_cpu
```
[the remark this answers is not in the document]
rbreslow: > Given that this application seems to be memory-bound rather than CPU-bound, a potential concern here (that I mentioned in Slack as well) is that increasing the number of cores available might increase memory pressure in a multi-threaded application (by increasing the number of threads).

That makes sense. I took the lever away because [ECS limits you](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html) to 10 vCPUs if you choose to specify a CPU limit. Since the instance has 16 vCPUs, we're essentially locking away some of that compute. Since there aren't any indications we're CPU bound, I added the lever back in 85dacca and figure we can remove it if we hit a wall at 10 vCPUs in the future.

> To be clear, I have no idea if this is actually a risk -- I'm not sure whether there even is any parallelism in the app, and even if there is, it may not be counting CPUs in order to choose how many threads to spawn. [...] I've seen the above failure mode occur in real life [...] my instinct would be to limit ourselves to adjusting memory only on this iteration.

Yeah, I totally agree. Although, knowing what we know now about `vm.max_map_count`, I don't think this was the culprit.

deployment/terraform/task-definitions/app.json.tmpl
rbreslow: > It seems that this is a floor, not a ceiling, so the task should be able to "float" to as much CPU usage as it wants, as long as there is no CPU contention from other tasks (which there won't be, because we're assigning 1 task per instance).

Yes! Good point. This subtly changes how the Docker Engine will ration CPU time, but since there's only ever one container in these workloads, we won't see any impact.

deployment/terraform/container_service.tf
```diff
@@ -157,28 +157,27 @@ data "aws_ec2_instance_type" "container_instance" {
...
   memory = local.container_instance_app_memory
 
   container_definitions = templatefile("${path.module}/task-definitions/app.json.tmpl", merge(
     local.shared_app_task_def_template_vars,
     {
-      max_old_space_size = local.container_instance_app_memory * 0.75
+      max_old_space_size = local.container_instance_app_memory * var.max_old_space_size_scale_factor
```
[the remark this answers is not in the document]
rbreslow: > [...] my reading of the Node docs is that the purpose of the "use 1.5GB out of 2GB" suggestion is to reserve sufficient space for system processes such that the Node app doesn't cause undue memory pressure on the host system but that's roughly what `container_instance_reserved_memory` is already doing, since it's based on an empirical assessment of baseline memory needs.

> Since we know that we already pushed this number to around 60GB on a 64GB instance today without problems (other than the fact that 60GB wasn't enough), it seems like it should be safe to go quite a bit higher here.

This makes sense, and I agree. I didn't think about how we're already fulfilling the spirit of the "use 1.5GB out of 2GB" suggestion by applying `container_instance_reserved_memory`.

> On the other hand, I may be misunderstanding something -- @maurizi noted that he had bumped `districtbuilder_state_count` to 60, which used up all but 135MB of RAM on the machine (as shown in the ECS console).

I don't think this is possible, mainly because there was a bug when you wrote this comment that would have prevented the application from using more than 17 gigs of memory. That may have been memory that was allocated to the task, rather than memory actually being utilized. 

> **Is it possible that we actually haven't tapped out the 64GB machine yet and that Node is artificially limiting itself to a lower value than what ECS is reserving on the host for the task?** [...]

> [...] (though I'd love to understand what's going on better once we're through this).

I don't fully understand the question. However, you posed it before the garbage collection errors, so I think it deserves a solid answer. Can we talk about it during our 1-on-1? :)

deployment/terraform/variables.tf
```diff
@@ -228,7 +228,7 @@ variable "image_tag" {
 }
 
 variable "container_instance_type" {
-  default = "r5.2xlarge"
+  default = "r5.4xlarge"
```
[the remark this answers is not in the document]
rbreslow: 🎣

deployment/terraform/container_service.tf
```diff
@@ -8,7 +8,22 @@ data "template_cloudinit_config" "container_instance_cloud_config" {
...
+      # largest difference between prediction vs. reality is 4791. Setting the
+      # intercept to 8 * 1024 appears to give us an adequate safety margin.
+      #
+      # Also, this method of determining a value for vm.max_map_count is only
+      # efficient if the container's memory usage equals the total memory of
+      # the host. So, there will always be a huge margin, but I feel that it's
+      # better than choosing an arbitrary constant.
+      #
+      # See: doc/DistrictBuilder_Memory_Usage_vs_Memory_Map_Areas.xlsx
+      vm_max_map_count = local.container_instance_app_memory * 3 + (8 * 1024)
```
rbreslow, line 26: <details>
<summary>Graph</summary>

Picture1

</details>

I recorded these values by splitting tmux into two panes: one watching the current memory map count of the container and another running `docker stats`.

```console
ps aux | grep node
sudo watch -d -n 1 "cat /proc/<PID>/maps | wc -l"
```

It'd be more accurate to determine what's going on under the hood (e.g., which objects are being `mmap`'d? why?), but this gives us an approximation applicable to our application's current profile.

maurizi:
I was able to apply this to staging by following the test instructions, and it successfully deployed a new version which became healthy / stable:

image

image

maurizi (approved):

deployment/terraform/container_service.tf
```diff
@@ -8,7 +8,22 @@ data "template_cloudinit_config" "container_instance_cloud_config" {
...
+      # an R^2 of 0.9017. The root-mean-square deviation is 3736, and the
+      # largest difference between prediction vs. reality is 4791. Setting the
+      # intercept to 8 * 1024 appears to give us an adequate safety margin.
+      #
+      # Also, this method of determining a value for vm.max_map_count is only
+      # efficient if the container's memory usage equals the total memory of
+      # the host. So, there will always be a huge margin, but I feel that it's
+      # better than choosing an arbitrary constant.
+      #
+      # See: doc/DistrictBuilder_Memory_Usage_vs_Memory_Map_Areas.xlsx
```
maurizi, line 25: I'd like to note that I found the attached doc very helpful in understanding how you arrived at this, though I also miss the wonderful ASCII chart you had here before

deployment/terraform/container_service.tf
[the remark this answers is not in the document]
ddohler: > Yeah, I totally agree. Although, knowing what we know now about vm.max_map_count, I don't think this was the culprit.

Yep, I agree with this as well, it seems that our memory requirements were quite a bit more modest than we thought.

deployment/terraform/container_service.tf
ddohler: This is great! I think an observational, empirical approach is perfectly fine here, given that we are storing large amounts of homogeneous data -- there's no reason to expect that the relationship is anything other than linear, and this is a quick and easy way to determine a reasonable number.

deployment/terraform/container_service.tf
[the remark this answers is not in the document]
ddohler: Sounds good, let's discuss offline -- I think the events of the past couple days have clarified my understanding of what was happening since I made this comment.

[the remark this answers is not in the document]
ddohler: For posterity, we discussed offline and agreed that this comment was correct but ultimately turned out to be irrelevant because the application wasn't limited by `max_old_space_size`.

deployment/terraform/variables.tf
```diff
@@ -299,6 +299,14 @@ variable "districtbuilder_state_count" {
   type = number
 }
 
+variable "max_old_space_size_scale_factor" {
+  # The official documentation mentions setting this to 1.5 GB on a machine that
+  # has 2 GB of memory.
+  # https://nodejs.org/api/cli.html#cli_max_old_space_size_size_in_megabytes
```
ddohler, line 305: It might be worth updating this default and/or comment to express that this value is currently 1.

ddohler (approved):
I haven't had time to actually perform the testing steps, but the proof, as they say, is in the functioning production deployment, and everything that's here looks good.

maurizi:
@rbreslow are we good to merge this now?

rbreslow:
> @rbreslow are we good to merge this now?

We weren't, because I didn't rebase some WIP commits and never added a changelog. I took half the day yesterday to focus on R&D and missed Derek's 👍 . Now, it's merged.

PR #944, opened 2021-08-16
Retry on topology errors

maurizi:
## Overview

On occasion we have encountered issues downloading the required JSON files for the `TopologyService` to load, and adding more states / increasing application resources seems to have increased the frequency of this.

Retrying on errors should enable us to get past any transitory errors w/o having to redownload / parse all the states we have already loaded

### Checklist

### Demo

### Notes

## Testing Instructions

- How to test this PR
- Prefer bulleted description

Closes #XXX

### DIFF
--- a/src/server/src/districts/services/topology.service.ts
+++ b/src/server/src/districts/services/topology.service.ts
@@ -77,7 +77,8 @@ export class TopologyService {
 
   private async fetchLayer(
     s3URI: S3URI,
-    archived: boolean
+    archived: boolean,
+    numRetries = 0
   ): Promise<GeoUnitTopology | GeoUnitProperties | void> {
     try {
       const [topojsonResponse, staticMetadataResponse] = await Promise.all([
@@ -116,7 +117,10 @@ export class TopologyService {
         this.logger.error("Invalid TopoJSON or metadata bodies");
       }
     } catch (err) {
-      this.logger.error(`Unable to download static files from for ${s3URI}: ${err}`);
+      this.logger.error(
+        `Failed to load topology for '${s3URI}' ${numRetries + 1} times, err ${err}`
+      );
+      this.fetchLayer(s3URI, archived, numRetries + 1);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/server/src/districts/services/topology.service.ts
```diff
@@ -116,7 +117,10 @@ export class TopologyService {
         this.logger.error("Invalid TopoJSON or metadata bodies");
       }
     } catch (err) {
-      this.logger.error(`Unable to download static files from for ${s3URI}: ${err}`);
+      this.logger.error(
+        `Failed to load topology for '${s3URI}' ${numRetries + 1} times, err ${err}`
+      );
+      this.fetchLayer(s3URI, archived, numRetries + 1);
```
rbreslow, line 123: This could recurse forever, so is the thinking that we'd rely on an ECS health check to kill the task? If this is on someone's local workstation, would we expect them to <kbd>Ctrl</kbd> + <kbd>C</kbd>? I'm trying to think through the consequences of having this be an infinite loop. It may be worth adding a `MAX_TOPOLOGY_RETRIES` like environment variable.

maurizi: > This could recurse forever, so is the thinking that we'd rely on an ECS health check to kill the task? 

Yes that was the thought.

In response to your comment I briefly explored using `process.exit(1)` to shutdown in response to a max # of retries, but that didn't actually shut down the docker container. In case of a mis-configuration the logs will at least be exceptionally noisy:
```
server_1    | [Nest] 73   - 08/16/2021, 2:37:45 PM   [TopologyService] Failed to load topology for 's3://fake-bucke
t/doesnt/exist/' 1 times, err AccessDenied: Access Denied +175ms
server_1    | [Nest] 73   - 08/16/2021, 2:37:45 PM   [TopologyService] Failed to load topology for 's3://fake-bucke
t/doesnt/exist/' 2 times, err AccessDenied: Access Denied +107ms
server_1    | [Nest] 73   - 08/16/2021, 2:37:45 PM   [TopologyService] Failed to load topology for 's3://fake-bucke
t/doesnt/exist/' 3 times, err AccessDenied: Access Denied +91ms
```

rbreslow (approved):
👍

PR #951, opened 2021-08-17
Fix CSV import

### LINKED ISSUE #945: Import CSV page errors
When attempting to import a CSV block equivalency file, I'm running into everything from general slowness to unresponsive behavior at various steps. It generally seems to be happening more often with state lower files, which do have a larger number of districts. 

Three types of unresponsive behavior:

- When importing the Florida Lower CSV, the page got stuck in the Upload step.
- When importing the Pennsylvania Lower CSV, the CSV was able to import but it got stuck on the Create Map step.
- When importing a Pennsylvania Congress CSV, the page was stuck on Create Map but later after navigating away from the page it did actually create the map.

[fl_lower.csv](https://github.com/PublicMapping/districtbuilder/files/7000507/fl_lower.csv)
[pa_lower.csv](https://github.com/PublicMapping/districtbuilder/files/7000509/pa_lower.csv)

kevinearldenny:
## Overview

Fix CSV import by allowing JSON payloads of up to 50mb, instead of the default which was originally 100kb

### Checklist

## Testing Instructions

- `./scripts/server`
- Attempt to upload the attached CSVs and create new projects for them. Note that PA requires 2020 data 

Closes #945

### DIFF
--- a/src/server/src/main.ts
+++ b/src/server/src/main.ts
@@ -4,6 +4,7 @@ import { NestFactory } from "@nestjs/core";
 import { AppModule } from "./app.module";
 import { BadRequestExceptionFilter } from "./common/bad-request-exception.filter";
 import { DEBUG } from "./common/constants";
+import * as bodyParser from "body-parser";
 
 async function bootstrap(): Promise<void> {
   const app = await NestFactory.create(AppModule, {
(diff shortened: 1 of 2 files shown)


src/server/src/main.ts
```diff
@@ -19,6 +20,8 @@ async function bootstrap(): Promise<void> {
   );
   app.useGlobalInterceptors(new ClassSerializerInterceptor(app.get(Reflector)));
   app.useGlobalFilters(new BadRequestExceptionFilter());
+  app.use(bodyParser.json({ limit: "50mb" }));
+  app.use(bodyParser.urlencoded({ limit: "50mb", extended: true }));
```
maurizi, line 24: This seems a little excessive - the problematic CSVs were only 7-6Mb
```suggestion
  app.use(bodyParser.json({ limit: "20mb" }));
  app.use(bodyParser.urlencoded({ limit: "20mb", extended: true }));
```

maurizi (approved):
image
image

PR #957, opened 2021-08-18
Disable Send to Plan Score button for incomplete projects

kevinearldenny:
## Overview

Disables the send to plan score button for any projects that have unassigned areas

### Checklist

- [ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

### Demo

image

## Testing Instructions

- Create an incomplete project and go to Evaluate Mode -> Competitiveness
- Expect: the Send to Plan Score button is disabled
- Complete the project, then return to Competitiveness page
- Expect: the Send to Plan Score button is now enabled again

Closes #906

PR #958, opened 2021-08-19
Allow retrieving the PlanScore for another users project.

maurizi:
## Overview

Fixes an issue I encountered in the course of testing #957 

Also refactors the `getProjectWithDistricts` method to remove redundant SQL WHERE clauses.

### Checklist

## Testing Instructions

- Attempt to submit PlanScore data for another users project. On `develop` this will fail w/ status code 401. On this branch it will return 200 (unless the PlanScore API is still down, in which case it will return 502).


kevinearldenny (approved):
LGTM 👍  Although the PlanScore API is still returning a 502 error, and it might be nice to show that on the frontend somewhere... but that's a separate issue it seems like

PR #967, opened 2021-08-20
Set custom bbox for projects in Alaska

### LINKED ISSUE #953: Alaska zooms to northern Europe
## Description
Creating a new map with Alaska initially zooms the map to Iceland

## Steps To Reproduce
1. Create a new map from Alaska

## Expected behavior
The map should zoom to Alaska initially.

## Actual behavior
The map zooms to Iceland.

## Your environment
MacOS, Firefox

## Additional context
Part of the Aleutian Island chain crosses the anti-meridian, and we've discovered issues on other projects with Mapbox GL and setting the map view when part of the bounding box crosses the anti-meridian, so I suspect something similar is happening here.

kevinearldenny:
## Overview

Sets a custom bbox for projects in Alaska 

### Checklist

### Demo

image

### Notes

I tested on a few different screen sizes and seems like the current bounding box works pretty well on larger and smaller screens.

## Testing Instructions

- Create a `region_config` for Alaska
- Create a project in Alaska 
- Close and re-open the project - expect project to zoom to Alaska, as opposed to Europe like `develop`

Closes #953

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -316,9 +316,11 @@ const DistrictsMap = ({
     panToggled &&
     !selectionInProgress &&
     (selectionTool === SelectionTool.Rectangle || selectionTool === SelectionTool.PaintBrush);
-
   // Conversion from readonly -> mutable to match Mapbox interface
-  const [b0, b1, b2, b3] = staticMetadata.bbox;
+  const [b0, b1, b2, b3] =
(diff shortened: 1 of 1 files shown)


maurizi:
There is a simpler / cleaner way to do this, which is what we did before for the 2010 version of this data - update the `bbox` in the `static-metadata.json` file on S3.

I would prefer to handle this problem that way.

PR #969, opened 2021-08-23
Speed up community maps

### LINKED ISSUE #934: Community Maps page is slow on production
This page performed well enough on staging, but on production with it's 18,000+ projects the query is far too slow.

We'll likely be able to solve this by adding an index to speed up the slow query.

maurizi:
## Overview

Adds indexes to speed up accessing the `project` table, as well as a script to generate test data for the purpose of verifying the indexes work.

Also cleans up the pagination footer component to better handle large number of pages.

### Checklist

### Demo

localhost_3003_ (5)

### Notes

We'll likely want to manually apply this migration to production, so that we can use `CREATE INDEX CONCURRENTLY ...` (which isn't allowed in a transaction, and so can't be run in a migration).

We should be able to safely add the index before deploying this fix to production.

## Testing Instructions
- `scripts/update`
- Ensure you have at least 1 region & 1 user in your database
- **At the end of the day,** run `scripts/manage create-random-projects 15000`, and then wait **5+ hours**.
- You should be able to load the `/maps` page in a reasonable amount of time.
- If you revert the latest migration, switch back to `develop`, restore access to community maps (https://github.com/PublicMapping/districtbuilder/commit/56b4e18a33a79149d863da2245f0a4d99abab2cd), and attempt to go to the community maps page, it should time out

Closes #934

### DIFF
--- a/src/server/migrations/1629680461750-project_index.ts
+++ b/src/server/migrations/1629680461750-project_index.ts
@@ -1,0 +1,19 @@
+import { MigrationInterface, QueryRunner } from "typeorm";
+
+export class projectIndex1629680461750 implements MigrationInterface {
+  name = "projectIndex1629680461750";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_110113f7f5d7d3b08d0c12f5b0" ON "project" ("updated_dt", "user_id")`
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_PUBLISHED_PROJECTS" ON "project" ("updated_dt" DESC, jsonb_array_length(project.districts->'features'->0->'geometry'->'coordinates'), "region_config_id") WHERE "archived" <> TRUE AND "visibility" = 'PUBLISHED'`
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP INDEX "IDX_110113f7f5d7d3b08d0c12f5b0"`);
+    await queryRunner.query(`DROP INDEX "IDX_PUBLISHED_PROJECTS"`);
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Disable Send to Plan Score button for incomplete projects [#957](https://github.com/PublicMapping/districtbuilder/pull/957)
 - Fix Send to Plan Score button for other users projects [#958](https://github.com/PublicMapping/districtbuilder/pull/958)
+- Restore Community Maps & fix API perf. [#969](https://github.com/PublicMapping/districtbuilder/pull/969)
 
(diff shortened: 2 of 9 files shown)


src/server/migrations/1629680461750-project_index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export class projectIndex1629680461750 implements MigrationInterface {
+  name = "projectIndex1629680461750";
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE INDEX "IDX_110113f7f5d7d3b08d0c12f5b0" ON "project" ("updated_dt", "user_id")`
+    );
+    await queryRunner.query(
+      `CREATE INDEX "IDX_PUBLISHED_PROJECTS" ON "project" ("updated_dt" DESC, jsonb_array_length(project.districts->'features'->0->'geometry'->'coordinates'), "region_config_id") WHERE "archived" <> TRUE AND "visibility" = 'PUBLISHED'`
```
ddohler, line 11: Can you explain a bit about how you ended up at this index definition? I'm looking at the query that's issued on the Community Maps page and the WHERE clause follows this form:
```sql
SELECT DISTINCT "distinctAlias"."project_id" as "ids_project_id", "distinctAlias"."project_updated_dt" FROM (
...
WHERE
  "project"."visibility" = $1 AND
  "project"."archived" = FALSE AND
  jsonb_array_length(project.districts->'features'->0->'geometry'->'coordinates')::integer = 0)
"distinctAlias"
ORDER BY "distinctAlias"."project_updated_dt" DESC, "project_id" ASC LIMIT 8 OFFSET 24 -- PARAMETERS: ["PUBLISHED"]
```

Since we're always querying for (in JS, not SQL) `features[0].geometry.coordinates.length === 0`, it seems to me that we could move that into the `WHERE` condition for the index so that the index only kicks in when we query for `length === 0`, which would cut down on index size and speed up this endpoint further. Is there a reason that wouldn't work (or that we might need to query for values of `length` that are non-zero)?

The other thing I'm wondering about is the inclusion of `region_config_id` -- I don't immediately see why including that in the index would be helpful, but I may be missing something.

ddohler (approved):
Everything works great, and I confirmed that the index allows the Community page to successfully load, whereas it doesn't without the index. I had a couple questions about the details of the index itself, but it's clearly doing the job it's supposed to do, so I think this is good either way.

src/server/migrations/1629680461750-project_index.ts
maurizi: > Is there a reason that wouldn't work (or that we might need to query for values of length that are non-zero)?

The `completed` flag that determines whether to add the filter to the WHERE clause is always used by the front-end, but it is optional on the API & Micah w/ Public Mapping intends to use the API to pull info about maps w/o passing the `completed` filter, so I wanted to ensure it was fast when querying for all published projects whether completed or not.

> The other thing I'm wondering about is the inclusion of region_config_id -- I don't immediately see why including that in the index would be helpful, but I may be missing something.

There is a dropdown on the Community Maps page that allows filtering by `region`.

ddohler: Got it, thanks for explaining!

PR #977, opened 2021-08-25
Fix PVI calculation

### LINKED ISSUE #970: Update PVI calculation
We should not be including other party votes when calculating PVI, only Dem / Rep:
https://github.com/PublicMapping/districtbuilder/blob/2a76d57e5e424df52c033fa9f57ecdd83921972d/src/client/functions.ts#L124-L131

maurizi:
## Overview

We were incorrectly including other party vote totals in calculating vote share for PVI, which should instead use the two-party vote share to calculate vote totals.

_Edit_ Additionally, we were calculating the 16 / 20 combined averages incorrectly, fixed in https://github.com/PublicMapping/districtbuilder/pull/977/commits/4f9224540feb152a37b97f4f6e19c7513f833ade

### Checklist

### Demo

On `develop` we incorrectly calculate PVI for AZ as R+8 when it should be ~R+4~ R+3.

Fixed PVI calculations, as well as omitting "other party" from the tooltip:
image

### Notes

Changing the tooltip for PVI means there will be a discrepancy between what we show in the PVI tooltip (two party vote) and what we show on the map tooltip / other voting metric tooltips, where we will include thrid parties when calculating vote total percentage,

## Testing Instructions

- `scripts/server`
- Load / create a project for a region with voting data (see https://docs.google.com/spreadsheets/d/1ugX-ku_rwooZZ5DAI5NuCpQCcVGT67fWYz2V22ICm8w/edit#gid=645349603). The PVI column should now calculate w/o other party votes counted.

Closes #970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed map outline ordering [#968](https://github.com/PublicMapping/districtbuilder/pull/968)
 - Fixed pinned metrics for read-only projects [#976](https://github.com/PublicMapping/districtbuilder/pull/976)
 - Fix duplicate button for projects in archived regions [#978](https://github.com/PublicMapping/districtbuilder/pull/978)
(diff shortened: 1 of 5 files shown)


KlaasH (approved):
:+1: Looks good!

PR #984, opened 2021-08-27
Use JSON configuration to control fields displayed in metrics table

### LINKED ISSUE #973: Handle extra demographic data fields in expandable metrics viewer and hover window
We're receiving extra demographic data fields from customers and currently, they don't display properly in the expandable demographics viewer. Instead of displaying all the fields, it only displays `Other` if that field is present. 
image

But you can see all the fields correctly in the label dropdown:
image

Also, the hover window doesn't display all the fields.

Here's a list of fields that we should aim to be able to support:
- Population
- White
- Black
- Hispanic
- Asian
- Native
- Pacific
- Other
- Multi

maurizi:
## Overview

In the first of what we anticipate to be a number of such requests, Dane County has asked to support both "Other" and "Multiracial" as demographics, in addition to the demographics we already have for WI.

To support this, I've removed our hardcoded list of fields we support to instead be based on what is configured in the `static-metadata.json` configuration file we create when preprocessing data.

We still have a hard-coded list that we use for sorting though, instead of relying solely on the sort order in `static-metadata.json` (any fields not on the hard-coded list, like "Multiracial", will sort to the end of the grouping).

### Checklist

### Demo

Note the "Other" and "Multiracial" columns:
Capture
Pinning works for the new column:
Capture2
Existing region still working as expected:
Capture3

## Testing Instructions

 - Using `scripts/dbshell`, run `TRUNCATE region_config CASCADE` to delete your regions / projects, then run `scripts/load/dev-data` to load new testing data
- Go to the testing organization http://localhost:3003/o/azavea and create a map for the Dane County template
- Use the double-arrow button in the sidebar to expand the metrics viewer
  - You should see columns for Native, Pacific, Other, and Multiracial
  - You should _not_ see a row for Multiracial in the map / sidebar tooltips

Closes #973

### DIFF
--- a/src/shared/functions.ts
+++ b/src/shared/functions.ts
@@ -104,7 +104,7 @@ export function getDemographicsMetricFields(staticMetadata: IStaticMetadata): Me
   // Need to loosen up the types here
   const order: readonly string[] = DEMOGRAPHIC_FIELDS_ORDER;
   // eslint-disable-next-line functional/immutable-data
-  data.sort(([, a], [, b]) => order.indexOf(a) - order.indexOf(b));
+  data.sort(([a,], [b,]) => order.indexOf(a) - order.indexOf(b));
   return data;
 }
 
--- a/src/shared/functions.ts
+++ b/src/shared/functions.ts
@@ -1,4 +1,13 @@
-import { UintArray, DemographicCounts, IStaticMetadata, IStaticFile } from "../shared/entities";
+import {
+  UintArray,
+  DemographicCounts,
+  IStaticMetadata,
+  IStaticFile,
+  MetricsList,
+  VotingMetricsList,
+  VotingMetricField
+} from "../shared/entities";
+import { CORE_METRIC_FIELDS, DEMOGRAPHIC_FIELDS_ORDER } from "./constants";
 
 // Helper for finding all indices in an array buffer matching a value.
 // Note: mutation is used, because the union type of array buffers proved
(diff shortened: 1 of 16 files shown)


KlaasH (commented):
What is the "hover window"? Is that not the map tooltip that shows up when you mouse over a region?  The issue says "and hover window" but this doesn't add the new category to the tooltips.  I would think they should be there.  It seems weird to list most-but-not-all of the demographic categories in the tooltips.

My other thought was that it would be nice to be able to apply an explicit sort order to the new field rather than having it sort to the end.

It seems more natural to me to put "Other" at the end.  I guess that makes sense if it means "any situation not covered by the other options" but sorting it before "Multiracial" would actually make sense if it means "Single race, but not one of the ones listed".

But regardless of which order makes sense, having the sort order be fully explicit seems better than making it mostly explicit plus a default behavior for the remainder.  It seems like the logic is mostly set up to just skip missing fields anyway, so I was thinking we could just add "multiracial" to `DEMOGRAPHIC_FIELDS_ORDER` in the position we want it, and let it show up only when it applies.  (And future additional categories could be handled the same way--add them to the list in the position we want them to show up if they're there.  That would work as long as we don't want the same categories sorting differently for different states/regions, which hopefully we wouldn't need to do.)

I tried that in my instance, but it crashed because there was no `multiracial` entry in `demographicsColors`.  So my two suggestions are connected...

maurizi:
> But regardless of which order makes sense, having the sort order be fully explicit seems better than making it mostly explicit plus a default behavior for the remainder. It seems like the logic is mostly set up to just skip missing fields anyway, so I was thinking we could just add "multiracial" to DEMOGRAPHIC_FIELDS_ORDER in the position we want it, and let it show up only when it applies. (And future additional categories could be handled the same way--add them to the list in the position we want them to show up if they're there. That would work as long as we don't want the same categories sorting differently for different states/regions, which hopefully we wouldn't need to do.)

> I tried that in my instance, but it crashed because there was no multiracial entry in demographicsColors. So my two suggestions are connected...

Good points. @dmcglone we do want "multiracial" in the tooltip? @jfrankl any opinions on a color? I can just pick something - we don't have any yellow demographic categories.

dmcglone:
> > But regardless of which order makes sense, having the sort order be fully explicit seems better than making it mostly explicit plus a default behavior for the remainder. It seems like the logic is mostly set up to just skip missing fields anyway, so I was thinking we could just add "multiracial" to DEMOGRAPHIC_FIELDS_ORDER in the position we want it, and let it show up only when it applies. (And future additional categories could be handled the same way--add them to the list in the position we want them to show up if they're there. That would work as long as we don't want the same categories sorting differently for different states/regions, which hopefully we wouldn't need to do.)
> 
> > I tried that in my instance, but it crashed because there was no multiracial entry in demographicsColors. So my two suggestions are connected...
> 
> Good points. @dmcglone we do want "multiracial" in the tooltip? @jfrankl any opinions on a color? I can just pick something - we don't have any yellow demographic categories.

Yes, I think it would be helpful to have it in the tooltip, given it seems to be important enough for them to include.

jfrankl:
> > > But regardless of which order makes sense, having the sort order be fully explicit seems better than making it mostly explicit plus a default behavior for the remainder. It seems like the logic is mostly set up to just skip missing fields anyway, so I was thinking we could just add "multiracial" to DEMOGRAPHIC_FIELDS_ORDER in the position we want it, and let it show up only when it applies. (And future additional categories could be handled the same way--add them to the list in the position we want them to show up if they're there. That would work as long as we don't want the same categories sorting differently for different states/regions, which hopefully we wouldn't need to do.)
> > 
> > 
> > > I tried that in my instance, but it crashed because there was no multiracial entry in demographicsColors. So my two suggestions are connected...
> > 
> > 
> > Good points. @dmcglone we do want "multiracial" in the tooltip? @jfrankl any opinions on a color? I can just pick something - we don't have any yellow demographic categories.
> 
> Yes, I think it would be helpful to have it in the tooltip, given it seems to be important enough for them to include.

I agree we should include these in the tooltip. For color, can we please use #D77EFF for "multiracial" and #5A563E for "other"? And I think it makes sense for "other" to always be last in the list.

maurizi:
@KlaasH updated in https://github.com/PublicMapping/districtbuilder/pull/984/commits/b82be6fc00e0e91ccfd5ddc8e34741564a3e3b96 to add `multiracial` to the two variables needed for it to show up on the tooltip.

I do think we might want to move the configuration for field order / color from TS to a JSON configuration file as well, if we get more requests like this in the future.

src/shared/functions.ts
```diff
@@ -81,3 +90,46 @@ export function getDemographicLabel(id: string) {
...
+  // eslint-disable-next-line functional/prefer-readonly-type
+  const data: (readonly [string, string])[] = staticMetadata.demographics.flatMap(file =>
+    CORE_METRIC_FIELDS.includes(file.id)
+      ? []
+      : [[file.id, getMetricFieldForDemographicsId(file.id)]]
+  );
+  // Need to loosen up the types here
+  const order: readonly string[] = DEMOGRAPHIC_FIELDS_ORDER;
+  // eslint-disable-next-line functional/immutable-data
+  data.sort(([, a], [, b]) => order.indexOf(a) - order.indexOf(b));
```
KlaasH, line 107:
```suggestion
  data.sort(([a,], [b,]) => order.indexOf(a) - order.indexOf(b));
```
The first item is the key, not the second. Since the field names aren't in the `order` list, sorting by that leaves the order unchanged.

KlaasH (approved):
:+1: Looks good!  One fix-up below that makes the columns in the sidebar table sort properly.

PR #987, opened 2021-08-27
Don't show voting data in map tooltip for regions w/o it

### LINKED ISSUE #972: Hide political data bar chart in Project when not available in Region data
## Description
We are displaying political data in the hover window for regions where we do not have political data. The app shouldn't display the political data bars in regions where political data is not present.
image

maurizi:
## Overview

Don't show political data for regions w/o it (like the Dane County hidden region)

### Checklist

### Demo

image

## Testing Instructions

- Create a map for a region w/o political data, like Dane County
  - Hover over a blockgroup on the map - the tooltip should not show poltical data
- Maps for regions with political data, such as DE 2020, should continue to show it in the tooltip

Closes #972


KlaasH (approved):
Was able to see the "before" for New York and confirm that the "after" works for that and a few other maps.  Looks good :+1:

PR #988, opened 2021-08-27
Hide copy/convert button for other users maps

### LINKED ISSUE #982: Don't show "Convert to 2020" modal on shared map
## Description
When viewing a shared map made by a different user, a user should not see the "Convert to 2020" modal. Only the original map maker should see that message.

## Screenshots
Screen Shot 2021-08-27 at 2 09 25 PM

maurizi:
## Overview

Hides the copy & convert to 2020 overlay / button for archived 2010 maps when  looking at another users map.
The button wouldn't work in this case anyways, so this should silence some Rollbar errors.

### Checklist

### Demo

Visible in the logged in browser, not visble in the not logged in incognito window:
image

## Testing Instructions

- Using `scripts/dbshell`, set a region for which you have projects to be `archived`
- If you go to a project of yours for that region, you should the copy / convert overlay (note that unless your archived region was for 2010 & you have a non-archived 2020 region configured for the same state, the convert button won't work)
- Go to that same project as another user, or while logged out. You should not see the overlay.

Closes #982

### DIFF
--- a/src/client/components/map/Map.tsx
+++ b/src/client/components/map/Map.tsx
@@ -190,6 +190,7 @@ interface Props {
   readonly evaluateMode: boolean;
   readonly isReadOnly: boolean;
   readonly isArchived: boolean;
+  readonly isThisUsersMap: boolean;
   readonly limitSelectionToCounty: boolean;
   readonly findMenuOpen: boolean;
   readonly findTool: FindTool;
@@ -285,6 +286,7 @@ const DistrictsMap = ({
   lockedDistricts,
   isReadOnly,
   isArchived,
+  isThisUsersMap,
   expandedProjectMetrics,
   limitSelectionToCounty,
   findMenuOpen,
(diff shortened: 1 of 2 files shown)


KlaasH (approved):
:+1: Does the thing.

It could also be done the other way around, i.e. calling the prop `isThisUsersMap` and reversing the tests. That seems more straightforward to me--fewer double-negatives--but it works either way.

src/client/components/map/Map.tsx
```diff
@@ -1186,7 +1188,11 @@ function mapStateToProps(state: State) {
     findMenuOpen: state.project.findMenuOpen,
     findTool: state.project.findTool,
     electionYear: state.project.electionYear,
-    showKeyboardShortcutsModal: state.project.showKeyboardShortcutsModal
+    showKeyboardShortcutsModal: state.project.showKeyboardShortcutsModal,
+    isThisUsersMap:
+      "resource" in state.user &&
+      "resource" in state.project.projectData &&
+      state.user.resource.id !== state.project.projectData.resource.project.user.id
```
KlaasH, line 1195:
```suggestion
      state.user.resource.id === state.project.projectData.resource.project.user.id
```

KlaasH (approved):
Thanks for switching!  One test didn't get flipped, as noted below.  But with that fixed up, this looks good :+1:

PR #902, opened 2021-08-09
Allow adding reference layers to project

kevinearldenny:
## Overview

- Add modal to support uploading a new reference layer
- Add `ReferenceLayer` entity to backend
- Add a new endpoint to parse and validate a geojson or csv upload
- Add new endpoint to save a `reference_layer` for a project
- Display all uploaded reference layers in a new pane in the project sidebar

### Checklist

- [x] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible

## Create form
image
image
image

## Displayed in sidebar
image

### Notes

Still need a bit of polish on the validation and csv support here, but figured I'd get a draft PR up here to begin reviewing since there is a lot here to look at.

## Testing Instructions

- `./scripts/update`
- Open a project, and click the + button in the bottom of the project sidebar on the left hand side to open the reference layer modal
- Attempt to upload the two attached geojson files; expect them to be saved to DB and record displayed in project sidebar

Closes #861

PR #999, opened 2021-09-03
Show / remove reference layers

### LINKED ISSUE #862: Remove reference layers from projects
Wireframes
image

AC
 - Can click on "..." menu next to reference layers, see "Remove from map" option in menu
 - Confirmation modal is displayed

### LINKED ISSUE #863: Display reference layers on map
AC
 - Saved reference layers should be displayed on the map when the checkbox in the sidebar is checked
 - Reference layers are not displayed by default (and we don't need to persist whether it is displayed or not to the database)

maurizi:
## Overview

This has two main features:
 - Shows reference layers that have been added on the map (
 - Allows deleting existing reference layers

Also includes some tweaks to what was added in #902
 - Allows not setting the label field when adding a reference layer
 - Hides the reference layer section when the metrics are expanded

### Checklist

### Demo

Showing point/polygon layers:
Toggle

Deleting layers:
Delete

Label option no longer required:
NoLabels

## Testing Instructions

- `scripts/update`
- Expand the reference layers section & add some CSV / geojson reference layers (I found state open data portals to be helpful resources when testing)
- Toggle the layers on/off using the checkboxes
- Try removing a reference layer using the "..." menu
- Reference layers should be visible when looking at another users map, but the "..." menu should not be

Closes #862 
Closes #863


src/client/components/map/Map.tsx
```diff
@@ -652,6 +652,9 @@ const DistrictsMap = ({
               source: getSourceId(layer.id),
               layout: {
                 "icon-image": "map-pin",
+                "icon-allow-overlap": true,
+                "icon-ignore-placement": true,
+                "text-optional": true,
```
maurizi, line 700: @jfrankl Wanted to note I made a conscious choice here to let point layer icons overlap (not the text) to prevent MapboxGL from collapsing adjacent icons together.

One of the main use cases for point reference layers we've discussed is incumbent locations, for which it seems quite important to show all the icons, so you know if you have more than 1 incumbent in the district you're drawing, even if they're quite close together.

src/client/components/ProjectReferenceLayers.tsx
```diff
@@ -45,20 +47,48 @@ const style: ThemeUIStyleObject = {
...
+      </Box>
+    )}
   </Flex>
 );
 
 const ProjectReferenceLayers = ({
-  referenceLayers
+  isReadOnly,
+  referenceLayers,
+  showReferenceLayers
```
kevinearldenny, line 87: Does it make sense to set `isExpanded` to true whenever there is at least one reference layer included on the map? When opening a map with reference layers, it's not particularly obvious that there are reference layers included:

image

Alternatively, perhaps switching the icon from a plus symbol (which I would associate with "adding a new layer" moreso than expanding the box) to something more traditionally associated with the idea of expandable content might make this more intuitive here.

@jfrankl , any thoughts here?

kevinearldenny (changes requested):
The core functionality looks good here 👍 

The only outstanding item I see here is making the UI more apparent that reference layers exist for a project when the reference layers box is not expanded. Perhaps we could switch out the `+` icon for something more appropriate here, or display the total count of reference layers next to the Reference Layers heading? Up to you whether we address this here or in a new issue.

src/client/components/map/Map.tsx
jfrankl: Great idea, this makes sense to me.

src/client/components/ProjectReferenceLayers.tsx
jfrankl: I think this is a good idea, to have the panel expanded by default when there are layers. Makes the feature much more discoverable, and it's easy to hide if needed.

maurizi: @kevinearldenny fixed in https://github.com/PublicMapping/districtbuilder/pull/999/commits/b306aa3143e5e01d72117a1e08de02cb647fe21c

kevinearldenny (approved):
LGTM 👍

PR #1009, opened 2021-09-15
Reduce tolerance for simplification for small regions

### LINKED ISSUE #994: Washington DC mini maps rendering isn't pretty
Minimaps on the Washington DC Community Maps page have a rendering issue, see screenshot below:

image

kevinearldenny:
## Overview

Set the tolerance to .001 instead of .005 for small units with an area than less than 1 (?) for the simplification used to improve performance for mini-maps that inadvertently caused rendering issues for DC and other small regions.

### Checklist

### Demo

image

## Testing Instructions

- Create and complete projects for DC and RI
- Look at them in the community maps viewer; expect the mini-maps to render properly

Closes #994


maurizi (approved):
👍 LGTM.

FYI I think the unit of the mysterious `1` is degrees squared (which is generally not the right unit to calculate area in, but for a simple heuristic like this its fine).

PR #1021, opened 2021-09-28
Update copy to remove reference to line data

### LINKED ISSUE #1010: Reference layer help text says we support lines, but our validation only allows point & polygon
We should either change the copy or enable lines to be uploaded as reference layers

maurizi:
## Overview

We don't support this quite yet, so this copy is incorrect

Closes #1010


maurizi:
Trivial, merging now that CI has passed.

PR #1029, opened 2021-09-29
Fix map CSV import

### LINKED ISSUE #1027: Regression in Import Project screen
On production/staging, imports are not working, with network tools showing the following error being returned:

image

maurizi:
## Overview

This regression was caused by the refactor in a13ff4 - by no longer explicitly listing fields we allowed extra fields to be included in the POST data.

### Checklist

## Testing Instructions

- Export one of your maps, and attempt to import it. On `develop` the import will fail, on this branch it should succeed

Fixes #1027


KlaasH (approved):
:+1: Looks good!

PR #1034, opened 2021-10-05
Add support for VAP/CVAP fields, additional data fields

### LINKED ISSUE #1031: Add custom demographic fields in metrics viewer
Clients are requesting additional/custom demographic fields in their dataset accessible in the metrics viewer. Currently, it will only display population, white, black, asian, hispanic, native, pacific, and other (+ political/compactness, etc.). We need to be able to support those additional fields. Ideally, this would be through a config file that defines the fields from the geojson.

Test data for San Bernardino, CA is here: s3://global-districtbuilder-dev-us-east-1/regions/US/CA/2021-09-30T13:15:11.777Z/

maurizi:
## Overview

San Bernadino requested a large number of additional data fields, which we needed to make a few tweaks to support:
 - VAP / CVAP fields need to use the VAP / CVAP population when calculating percentages
 - We needed a way to order these fields in the expanded metrics viewer
 - Majority race calculations needed to be updated to not use the extra fields

### Checklist

### Demo

localhost_3003_projects_556f6c44-f70c-42a6-9dc8-e45080d57857

### Notes

I specifically chose **not** to update `process-geojson` to add a way to persist these changes into `static-metadata.json` as part of the command, since the configuration ended up somewhat complex. Rather, we have a default that should work for most regions (display the same fields as are in the tooltip / used for the majority race calculation), and we can _manually_ edit the JSON file to add `demographicGroups` when needed.

I have already updated the `static-metadata.json` at `s3://global-districtbuilder-dev-us-east-1/regions/US/CA/2021-09-30T18:00:31.844Z/static-metadata.json` to include a groups definition.

While I was changing things in the `IStaticMetadata` type, I also removed the currently no longer used `labels` field (https://github.com/PublicMapping/districtbuilder/commit/a6339e00dda8fb356f6af8193533ebf933e21f32), and I also fixed a React warning that popped up from a misuse of Theme-UI had been lurking around unnoticed (https://github.com/PublicMapping/districtbuilder/commit/1c547fe93a54aa212fa21a17caa6496089e158c8)

## Testing Instructions

- Set up a region using the preprocessed data mentioned in the issue:
   - `INSERT INTO region_config VALUES (DEFAULT, 'San Bernadino', 'US', 'CA', 's3://global-districtbuilder-dev-us-east-1/regions/US/CA/2021-09-30T18:00:31.844Z/', DEFAULT, DEFAULT, DEFAULT, '2020');`
- The expanded metrics viewer should show all fields, in the correct order (first group, majority race col, all other groups) for a map in the above region (on `develop` they will not be in the correct order)
- The majority race calculation should show Hispanic for the region as a whole (on `develop` it shows a non-race field)
- There shouldn't be regressions in a map for a region w/o `demographicsGroups` specified

Closes #1031

### DIFF
--- a/src/client/components/DemographicsChart.tsx
+++ b/src/client/components/DemographicsChart.tsx
@@ -19,6 +19,7 @@ const DemographicsChart = ({
 }: {
   readonly demographics: { readonly [id: string]: number };
 }) => {
+  // Only showing hard-coded core metrics here for space reasons, so we can hard-code population as well
   const percentages = mapValues(
     demographics,
     (population: number) =>
--- a/src/client/components/DemographicsTooltip.tsx
+++ b/src/client/components/DemographicsTooltip.tsx
@@ -70,6 +70,7 @@ const DemographicsTooltip = ({
   readonly isMajorityMinority?: boolean;
   readonly abbreviate?: boolean;
 }) => {
+  // Only showing hard-coded core metrics here for space reasons, so we can hard-code population as well
   const percentages = mapValues(
     demographics,
     (population: number) =>
--- a/src/client/components/ProjectListFlyout.tsx
+++ b/src/client/components/ProjectListFlyout.tsx
@@ -46,8 +46,7 @@ const ProjectListFlyout = (props: FlyoutProps) => {
         sx={{
           ...{ variant: "buttons.ghost", fontWeight: "light", "& > svg": { m: "0 !important" } },
           ...style.menuButton,
-          ...invertStyles(props),
-          ...props
+          ...invertStyles(props)
         }}
         className="project-list-flyout-menu"
       >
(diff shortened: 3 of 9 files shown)


maurizi:
> Hide link to Competitiveness Evaluate page for regions w/o political data #866 may have regressed at some point? I was able to see the Competitiveness link on San Bernadino even though it doesn't have any political data.

I'll look into this & probably open a small PR to fix it, good catch!

> Is the format of static-metadata.json described anywhere? If we're going to be relying on hand-edits to it in order to enable advanced features, it will probably be helpful to have some documentation about it.

There are types for it in `src/shared/` but there isn't any documentation. Any opinions on where we might put this? As a doc comment above the type? As a `.md` file in `docs/`?

> The S3 path (US/CA/<timestamp>/) for San Bernadino is the same as for any other city in California, except for the timestamp. It might be worthwhile to spend a bit of time thinking about adjusting our storage scheme to better handle nested regions like this. Not critical, because the robots don't care. 🙂 But a more human-friendly path might be helpful in troubleshooting situations.

I don't think we rely on the specific folder structure in S3 anywhere, so such a refactor should be a matter of moving things in S3 in conjunction w/ updating the path to them in the database.

Issue #960, opened 2021-08-19
PG sessions blocking in idle state waiting on ClientRead


rbreslow:
TBD.

maurizi:
After some trial and error, I think I've identified the issue in question causing this error - we're blocking the event loop when running the TopoJSON mergeArcs function (or other CPU intensive things like imports / exports).

Using the load testing script in https://github.com/PublicMapping/districtbuilder/pull/1046 I was able to recreate the idle transaction / slow to respond health checks and cause the staging site to temporarily 504.

By switching the merge function to a worker thread pool (https://github.com/PublicMapping/districtbuilder/commit/c214bac64111615e7a213972d512a7d765c68821) I was able to increase throughput & completely reduce the slowness in the healthcheck endpoint. Unfortunately we can't use that branch as-is (it takes far too long to run `deserialize` for large regions) but this is a great starting point to work from.

rbreslow:
I thought I'd share the query that I initially used to determine that sessions were idle (`state_change`) waiting on ClientRead (`wait_event`). I'm not sure if I left it anywhere, but it seems appropriate to keep it in this issue.

```sql
SELECT client_addr,
    query_start,
    state_change,
    wait_event,
    EXTRACT(epoch FROM (current_timestamp - query_start)) AS duration,
    query
FROM pg_stat_activity
WHERE wait_event IS NOT NULL
    AND backend_type = 'client backend'
	AND query = 'SELECT 1';
```

Awesome! I'm so happy that the load testing yielded fruit. I'm also excited to hear whether or not you can scale down the application now and if you continue any exploration into the data loading at startup.

PR #1046, opened 2021-10-22
Add HAR files & load test harness to replicate healthcheck timeouts

### LINKED ISSUE #1015: Add load test to replicate production health check timeouts
We'd like to attempt to replicate https://github.com/PublicMapping/districtbuilder/issues/960 - setting up a repeatable load test would be a good first step.

maurizi:
## Overview

We had an existing `k6` based load-testing suite, but it was outdated and needed to be updated, so I replaced what was there with a harness script similar to what we've used in other projects (OAR, Raster Foundry, etc.), capable of loading a .HAR file with any set of archived requests (generated from the network tab of the browser developer tools).

The .HAR files I uploaded use a local developer environment, but the domain is overridable via an environment variable (overiding like this is required to use `k6` through Docker regardless).

### Checklist

### Demo

image

### Notes

 - I spent some time trying to support testing both the healthcheck & project endpoints from 1 script, but in the end found it easier to run two instances of `k6` in separate tabs, one running the health check .HAR & the other running the project editing .HAR file
 - There is support in `k6` for creating virtual users by duplicating requests, but I couldn't use this for the project `PATCH` requests, as we really wanted to simulate concurrent access to separate projects, which is why I re-invented a tiny variant of this virtual user support myself.

## Testing Instructions
- Ensure you have a PA project setup and note the UUID (use this for `PROJECT_UUIDS` in the example below)
- Open the network tab in the developer tools and find the `Authorization` header for an API request, and note the value (use the value after `Bearer ` in the examples below)
- Using the directions in the `load-testing/README.md`, in three separate shells run the following:
  - `scripts/server` (make sure to wait for all topology layers to finish loading)
  - `JWT_AUTH_TOKEN="..." HAR_FILE=patch-local-pa-project.har PROJECT_UUIDS="..." REQ_ORIGIN="http://host.docker.internal:3003" docker-compose run --rm k6`
   - `JWT_AUTH_TOKEN="..." HAR_FILE=healthcheck.har REQ_ORIGIN="http://host.docker.internal:3005" docker-compose run --rm k6 run /scripts/harness.js --iterations=4`
- You should see the above load tests both succeed
- Duplicate your PA project a number of times, and include those additional UUIDS as part of `PROJECT_UUIDS`, and rerun the load testing scripts. You should be able to reliably cause the healthchecks to fail under sufficient load (10 concurrent projects being edited was enough for me)

Closes #1015


maurizi:
@ddohler I put this to use in diagnosing #960, so I don't think you need to test this yourself, but could I get a quick review of the changes here?

load-tests/harness.js
```diff
@@ -0,0 +1,124 @@
...
+          "is status 200": (r) => r.status === 200,
+        });
+      });
+
+      // If there's another batch of requests, sleep for the appropriate duration
+      if (nextBatch) {
+        sleep(Math.floor((nextBatch - batch) / 1000));
+      }
+    }
+  });
```
rbreslow, line 123: It's worth surfacing some subtleties here (so that they're out of my head :)!). This version of the load testing harness runs all requests within a single virtual user (VU), which you can think of as a single thread of execution. And, `http.batch` [runs synchronously](https://community.k6.io/t/does-code-in-groups-after-http-calls-run-synchronously/1384).

If one of the requests in the `http.batch` call decides to hang or time out, we'll never make it to the next batch of requests. This behavior can be problematic because it can give the application too much breathing room.
Real traffic keeps coming and coming, even when other requests are still in the queue. We ran into an issue with OAR where initial load tests could not replicate the real failure scenario.

The [OAR load test](https://github.com/open-apparel-registry/open-apparel-registry/blob/6034c548da0bd46866e433f7ebd9b2a93c8238be/load-tests/zoom_rio_de_janerio_with_contributor_filter.js#L66-L94) has some slight differences. The number of VUs corresponds to the number of batches of requests. Each VU sleeps until its turn, fires off one `http.batch` call, and finishes—sort of like a relay race. The benefit there being the load test will continue to replay traffic even while other VUs/`http.batch` invocations haven't received responses. The drawback being you lose the ability to scale the load test with K6 parameters.

ddohler (approved):
I didn't try to use this but given that it allowed us to replicate our production health-check issues, it's good enough for now and should be a good starting point for any future diagnosis we need to do. Thanks @rbreslow for adding that additional context!

PR #1065, opened 2021-11-16
Add Map Configuration modal to project screen

### LINKED ISSUE #1025: Map configuration modal
We'd like to have a place to edit various settings that as of now can only be edited during project creation:
image
image

AC:
From new modal, can edit / save:
 - name
 - population deviation
 - multi-member settings
 
 The wireframes also show "description" & "district names" - those can be skipped.

maurizi:
## Overview

Allows editing name, pop. deviation, and multi-member settings via a new modal on the Project screen

### Checklist

### Demo

image

image

image

## Testing Instructions

- Create a new map or go to an existing map and use the "..." menu button in the header next to the project name to open the Map Configuration modal
  - Saved changes should be reflected (sidebar updating, name changing, etc)
  - Validation errors should be displayed in the form (e.g. deviation > 100)
  - The "..." menu should _not_ be shown for a public map that is not yours (or while you are not logged in)

Closes #1025

### DIFF
--- a/src/client/components/MultiMemberForm.tsx
+++ b/src/client/components/MultiMemberForm.tsx
@@ -1,45 +1,71 @@
 /** @jsx jsx */
-import { jsx, Styled, Input } from "theme-ui";
+import React from "react";
+import { jsx, Box, Styled, Input } from "theme-ui";
 
 interface Props {
   readonly totalPopulation: number;
   readonly numberOfMembers: readonly number[];
+  readonly errors?: readonly string[];
   // eslint-disable-next-line functional/no-mixed-type
   readonly onChange: (numberOfMembers: readonly number[]) => void;
 }
 
-const MultiMemberForm = ({ totalPopulation, numberOfMembers, onChange }: Props) => {
-  const totalReps = numberOfMembers.reduce((total, numberOfReps) => total + numberOfReps, 0);
+const MultiMemberForm = ({ totalPopulation, numberOfMembers, errors, onChange }: Props) => {
+  const totalReps = numberOfMembers.reduce(
+    (total, numberOfReps) => (Number.isNaN(numberOfReps) ? total : total + numberOfReps),
+    0
+  );
   const popPerRep = Math.floor(totalPopulation / totalReps);
   return (
-    <Styled.table sx={{ margin: "0", width: "100%" }}>
-      <thead sx={{ bg: "muted" }}>
-        <tr>
-          <td>Districts</td>
-          <td>Number of reps</td>
-          <td sx={{ textAlign: "right" }}>Target population</td>
-        </tr>
-      </thead>
-      <tbody>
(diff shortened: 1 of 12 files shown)


src/client/components/MultiMemberForm.tsx
```diff
@@ -28,6 +28,7 @@ const MultiMemberForm = ({ totalPopulation, numberOfMembers, onChange }: Props)
               <Input
                 sx={{ maxWidth: "200px" }}
                 value={numberOfReps}
+                pattern="[0-9]+"
```
KlaasH, line 31: I noticed that if I start typing something that's not a number in one of these boxes, it immediately changes to NaN, and all the target population values also go to NaN.
image

It won't let me save the form that way, so it's still doing it's job as validation, but it's not the most graceful failure mode.

src/client/components/ProjectDetailsModal.tsx
```diff
@@ -0,0 +1,324 @@
...
+        isMultiMember,
+        valid: true
+      }
+    : {
+        name,
+        populationDeviation,
+        numberOfMembers,
+        isMultiMember,
+        valid: false
+      };
```
KlaasH, line 99: Not sure it's actually better, but this could be rewritten to put the condition inline in the object, since the value of `valid` is the only thing actually affected by the condition.
```suggestion
  return {
        name,
        populationDeviation,
        numberOfMembers,
        isMultiMember,
        valid: name && populationDeviation !== null && numberOfMembers !== null
  };
```

KlaasH (approved):
:+1: Looks good!
I found one quirk in the multi-member validation, but it wasn't actually introduced by these changes.

src/client/components/ProjectDetailsModal.tsx
maurizi: If I try this out, I get type errors.

I think our being verbose here helps the TS compiler see that we're returning either a `ValidForm` or an `InvalidForm`, but it may be getting confused for some other reason.

KlaasH: Doh, yeah, I didn't try it out, just read it, squinted at it to see what was different, then wrote this down.  It does seem like in a perfect world the type checker would be able to recognize that this still comes out to `ValidForm | InvalidForm`, but yeah, not today.

src/client/components/MultiMemberForm.tsx
maurizi: @KlaasH I reworked how we display invalid values and added server-side validation in https://github.com/PublicMapping/districtbuilder/pull/1065/commits/46c0a76cb2a8adbeea241994a222daf111b06a6d, care to take another look?

KlaasH (approved):
:+1: Looks good!  When I try to set any of the numbers to "potato", it just sighs and blanks the field until I start acting right, rather than getting flustered and showing NaN.

PR #1068, opened 2021-11-18
Update Competitiveness metric on Evaluate 

### LINKED ISSUE #1050: Updates to Competitiveness metric Evaluate screen
Due to some confusion about the PVI metric and because it is the same as the statewide PVI when a plan is complete, some small tweaks to the competitiveness metric should take place.

On the Evaluate panel for the Competitiveness metric, where we are currently displaying the plan PVI score, we should instead display the partisan breakdown of the plan's districts, i.e. "5D/4R" or "12D/5R"
image

Similarly, on the Competitiveness Evaluate mode screen, we should remove the "Partisan Voting Index (PVI): R+4" and simply instead display the overall district breakdown, as above: "5D/4R", etc.
image

rachelekm:
## Overview

Update the Competitiveness metric on the Evaluate panel to instead display PVI breakdown by districts, according to the following format:

- By partisan breakdown i.e. “4 R / 5 D” or “5R / 1E (Even) / 2D” 
- Colored according to party color within Heading of `CompetitivenessMetricDetail`
- “N/A” if districts' PVI data is undefined

### Checklist

### Demo

Expected display within `CompetitivenessMetricDetail`:
Screen Shot 2021-11-18 at 12 31 28 PM

Expected display of Competitveness metric within `ProjectEvaluateView`:
Screen Shot 2021-11-18 at 12 31 38 PM

### Notes

Since the competitiveness metric now uses `pviByDistrict` (derived from `pviBuckets` passed down through props) instead of a competitiveness value as well as implements `formatPviByDistrict` for party labelling and coloring, we no longer need `value` and `party` to be updated for this metric. However, I’ve kept the `selectEvaluationMetric` store dispatch as part of  `ProjectEvaluateSidebar` to update `electionYear` since that value is still used throughout the project. As far as I can tell, there are no side effects to getting rid of `party` and `value` in the competitiveness metric and the resulting `EvaluateMetric`, but it would be worth double-checking.

## Testing Instructions

- Create/import a map and assign some districts. Confirm the Competitiveness metric on the Evaluate panel as well as on the Competitiveness Evaluate mode screen displays the new format as expected.
- Create an empty map or create/import a map with no PVI district data and confirm "N/A" displays as the Competitiveness metric

Closes #1050

### DIFF
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -22,7 +22,6 @@ import {
   ChoroplethSteps,
   DistrictGeoJSON,
   ElectionYear,
-  Party,
   DistrictsGeoJSON,
   ReferenceLayerGeojson,
   PviBucket
@@ -75,14 +74,6 @@ export const getPartyColor = (party: string) =>
 export const getMajorityRaceDisplay = (feature: DistrictGeoJSON) =>
   feature.properties.majorityRace && capitalizeFirstLetter(feature.properties.majorityRace);
 
-export function formatPvi(party?: Party, value?: number): string {
-  return value !== undefined && party !== undefined
-    ? `${party.label}+${Math.abs(value).toLocaleString(undefined, {
-        maximumFractionDigits: 0
-      })}`
-    : "N/A";
-}
-
 /* Creates array of party-labelled pvi bucket counts as strings */
 export function formatPviByDistrict(
   pviBuckets: readonly (PviBucket | undefined)[] | undefined
@@ -90,12 +81,10 @@ export function formatPviByDistrict(
   const partyLabels = ["R", "E (Even)", "D"];
   // Count by partyLabels
   const bucketCounts = pviBuckets?.reduce(
-    // eslint-disable-next-line
-    (allBuckets: { [key: string]: number } | undefined, bucket: PviBucket | undefined) => {
+    (allBuckets: { readonly [key: string]: number } | undefined, bucket: PviBucket | undefined) => {
       const name =
         bucket &&
         partyLabels.find(label => bucket.name.includes(label) || label.includes(bucket.name));
-      // eslint-disable-next-line
       return name
         ? allBuckets && name in allBuckets
           ? { ...allBuckets, [name]: allBuckets[name] + 1 }
@@ -110,7 +99,7 @@ export function formatPviByDistrict(
     partyLabels
       .map((label: string) => {
         return bucketCounts[label]
-          ? `${Math.abs(bucketCounts[label]).toLocaleString(undefined, {
+          ? `${bucketCounts[label].toLocaleString(undefined, {
               maximumFractionDigits: 0
             })} ${label}`
           : undefined;
--- a/src/client/functions.ts
+++ b/src/client/functions.ts
@@ -24,7 +24,8 @@ import {
   ElectionYear,
   Party,
   DistrictsGeoJSON,
-  ReferenceLayerGeojson
+  ReferenceLayerGeojson,
+  PviBucket
 } from "./types";
 
 export function areAnyGeoUnitsSelected(geoUnits: GeoUnits) {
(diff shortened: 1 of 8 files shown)


src/client/functions.ts
```diff
@@ -74,12 +74,40 @@ export const getPartyColor = (party: string) =>
...
-    : "N/A";
+/* Creates array of party-labelled pvi bucket counts as strings */
+export function formatPviByDistrict(
+  pviBuckets: readonly (PviBucket | undefined)[] | undefined
+): readonly string[] | undefined {
+  const partyLabels = ["R", "E (Even)", "D"];
+  // Count by partyLabels
+  const bucketCounts = pviBuckets?.reduce(
+    // eslint-disable-next-line
+    (allBuckets: { [key: string]: number } | undefined, bucket: PviBucket | undefined) => {
```
maurizi, line 85: This will clean up the lint:
```suggestion
    (allBuckets: { readonly [key: string]: number } | undefined, bucket: PviBucket | undefined) => {
```

src/client/functions.ts
```diff
@@ -74,12 +74,40 @@ export const getPartyColor = (party: string) =>
...
+    },
+    {}
+  );
+  // Create string with partyLabels label
+  const bucketCountsStrings =
+    bucketCounts &&
+    partyLabels
+      .map((label: string) => {
+        return bucketCounts[label]
+          ? `${Math.abs(bucketCounts[label]).toLocaleString(undefined, {
```
maurizi, line 104: Is there a situation where `bucketCounts` can be negative? I would expect a count to be non-negative.
```suggestion
          ? `${bucketCounts[label].toLocaleString(undefined, {
```

src/client/types.d.ts
```diff
@@ -83,7 +83,7 @@ export interface PviBucket {
 }
 
 export interface EvaluateMetricWithValue extends BaseEvaluateMetric {
-  readonly type: "fraction" | "percent" | "count" | "pvi";
+  readonly type: "fraction" | "percent" | "count" | "pvi" | "pvibydistrict";
```
maurizi, line 86: With the changes in this PR, the `"pvi"` type is no longer used:
```suggestion
  readonly type: "fraction" | "percent" | "count" | "pvibydistrict";
```

src/client/types.d.ts
```diff
@@ -72,7 +72,7 @@ export type ElectionYear = "16" | "20" | "combined";
 
 export interface Party {
   readonly color: string;
-  readonly label: "D" | "R";
+  readonly label: "D" | "R" | "E (even)";
 }
```
maurizi, line 76: I don't think we use the `Party` interface anymore and can remove it:
```diff
diff --git a/src/client/types.d.ts b/src/client/types.d.ts
index d6fb0c7..51d3532 100644
--- a/src/client/types.d.ts
+++ b/src/client/types.d.ts
@@ -70,11 +70,6 @@ export interface BaseEvaluateMetric {
 
 export type ElectionYear = "16" | "20" | "combined";
 
-export interface Party {
-  readonly color: string;
-  readonly label: "D" | "R" | "E (even)";
-}
-
 export interface PviBucket {
   readonly name: string;
   readonly label: string;
@@ -83,10 +78,9 @@ export interface PviBucket {
 }
 
 export interface EvaluateMetricWithValue extends BaseEvaluateMetric {
-  readonly type: "fraction" | "percent" | "count" | "pvi" | "pvibydistrict";
+  readonly type: "fraction" | "percent" | "count" | "pvibydistrict";
   readonly value?: number;
   readonly total?: number;
-  readonly party?: Party;
   readonly hasMultipleElections?: boolean;
   readonly electionYear?: ElectionYear;
   readonly populationPerRepresentative?: number;
```

maurizi (changes requested):
This is looking quite good, just a few changes I suggested but its working well in testing 🎉

maurizi (approved):

PR #1077, opened 2021-11-30
Change "county" in County Split copy

### LINKED ISSUE #1057: Change County splits Evaluate copy
image

When using a different layer than County as the top level geo, this text doesn't make sense anymore. It could be parameterized to replace "County" with whatever the top level geo is, but it might just be easier to change the text to:
`Splits occur when a jurisdiction is split between two or more districts. Some states require minimizing splits, to the extent practicable.`

ktohalloran:
## Overview

The copy describing county splits wasn't always reflecting the right terminology that matched the top level geo of the map. This PR changes the copy to use the same word as the "County Splits" title.

### Checklist

### Demo

Screen Shot 2021-11-30 at 1 26 38 PM

## Testing Instructions

- Run `./scripts/server`
- Create a Dane County map using the Azavea's project template
- Click "Evaluate". You should see "Ward Splits" at the bottom under "Optional", as well as the word "ward" where there was the word "county".
- Create another map of anywhere but Dane County and click "Evaluate". The sidebar should now say "County Splits" and use the word "county".

Closes #1057

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,55 +6,77 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-### Added
-### Changed
-### Fixed
 
+### Added
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


maurizi (approved):
looks good 👍

PR #1078, opened 2021-12-02
Change copy under Reference Layers on other users' maps

### LINKED ISSUE #1032: Reference layer modal copy shouldn't say "Add" on other user's maps
When viewing another user's map, the reference layer modal displays the standard text "Add a polygon...". You can't add reference layers to other people's maps (which is expected - if someone wants to add a reference layer, they should copy the map into their own account). This text should be changed to "This map has no reference layers. Copy the map to your account to add reference layers."

image

ktohalloran:
## Overview

When viewing another user's map, the Reference Layer modal showed the same language as when viewing your own map about adding a layer. This PR changes that language to reflect the map's readonly status.

### Checklist

### Demo

Screen Shot 2021-12-02 at 7 43 24 AM

New language when viewing another person's map.

## Testing Instructions

- Run `./scripts/server`
- Open another user's map
- Click on the Reference Layers modal at the bottom of the sidebar. It should say "This map has no reference layers. Copy the map to your account to add reference layers."

Closes #1032

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix Reference Layer copy on other user's maps to reflect map's readonly status [#1078](https://github.com/PublicMapping/districtbuilder/pull/1078)
+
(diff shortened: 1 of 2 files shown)


maurizi (approved):
👍

PR #1079, opened 2021-12-02
Increase VirtualBox memory in Vagrantfile

ktohalloran:
## Overview

Some computers were killing the server while trying to run the VM due to the allotted memory for the VM being too small. This PR increases the allotted memory.

### Checklist

### Notes

I'm not sure what the ideal increase in memory is; this was enough for my computer to let the server run without seeming like an unnecessary amount of memory.

## Testing Instructions

- Run `vagrant up`, then `vagrant ssh`, and then `./scripts/server`
- If you open up the VirtualBox UI you should see that your VM's base memory has increased to 8192 MB
- Let the server start up and let it run for a few minutes. If your server was repeatedly being killed (like mine) with the original memory allotment, it should run without being killed now


maurizi (approved):
👍 looks good to me, didn't test

PR #1088, opened 2021-12-21
Update server and management node dependencies

### LINKED ISSUE #210: Use eslint in manage CLI project
## Overview

We've switched to `eslint` for both the server and client, but it looks like the CLI project still has `tslint` references.

## Acceptance criteria

The manage CLI project no longer has any references to `tslint`, and is being linted properly with `eslint`.

### LINKED ISSUE #891: Update TypeORM / Nest.JS / related dependences
The version of TypeORM we are using doesn't fully support default values for array columns, needing very verbose syntax for the default & causing it to erroneously regenerate migrations for any array columns with defaults (e.g. the `pinned_metric_fields` column on `Project` / `ProjectTemplate`).

I attempted to upgrade TypeORM to version `0.2.36` to fix this, which seemed to work and fixed the issue w/ spurious migrations being generated, but caused `nextjsx/crud` to break by generating incorrect SQL in the WHERE clause for the `filter` option for the `toggleFeatured` endpoint.

maurizi:
## Overview

Updates Node, Nest.JS (and related plugins like `nestjs/crud`), TypeORM, `eslint`, and `prettier` for the `server` and `manage` modules.

The main intent here was to update our version of TypeORM, which was generating spurious migrations related to the `pinned_metric_fields` column whenever generating new migrations (requiring editing the new migrations to remove those lines), but doing so pulled on a long chain of dependencies that needed to be updated - the main one being that updating Nest.JS moved us from TypeScript 3 -> 4 and Node 12 -> 16 (I think we could have jumped to only 14, but I choose to go to the most recent LTS), which then required updating a number of developer tools like `eslint` and `prettier`.

### Checklist

## Testing Instructions

- `scripts/update`
- `scripts/server`
  - Test out the application and ensure there haven't been any regressions
- Spot check the `scripts/manage` commands for regressions
  - Process & publish geojson for a small region
  - `./scripts/manage create-random-projects 1 all`
  - etc.

Closes #891 
Closes #210


src/manage/.prettierrc
```diff
@@ -1,3 +1,5 @@
 {
-  "printWidth": 100
+  "printWidth": 100,
```
ddohler, line 2: It looks like we have two of these files and two of the `.eslintrc.js` files as well -- could we use a single, shared file for one or both?

ddohler (commented):
I haven't been able to test yet because my machine is still busily chugging away building an OTP graph, but this looks good -- the number of changes was surprisingly limited considering how many things got upgraded and how many major versions we jumped, and I only have one small comment after reading through the changes. I am seeing that Jenkins says there are still some linting errors that need to be resolved, however.

ddohler (changes requested):
Overall this is looking good! I tried a bunch of different stuff in the app and most things are working. However, I did manage to find some requests that result in errors:

- Registering a new user; raises `TypeError: Cannot destructure property 'templateName' of 'precompile(...)' as it is undefined.` and the user sees an error dialogue, although the relevant database rows do seem to get inserted.
- Moving geographies from one district into another; raises `error: invalid input syntax for type uuid: "string"`
- Adding pinned metrics; raises the same error as moving geographies.

There may be others but everything else that I tried did work, so hopefully any problems that I've missed have the same root causes as one of the above and will be resolved if we fix those.

src/manage/.prettierrc
maurizi: We have _3_ actually - one for the `server`, one for `manage` & one for `client` (which wasn't changed in this PR).

I think ideally we would consolidate, but that seems like an effort in itself so I didn't want to do that as part of this PR.

maurizi:
@ddohler thanks for the careful testing! I've addressed those issues in https://github.com/PublicMapping/districtbuilder/pull/1088/commits/4de69e7ca91f5501a589884e067195a958efafea & https://github.com/PublicMapping/districtbuilder/pull/1088/commits/3d603c4daf814a0c04ff44ee66ee31ae1bca3441

ddohler (approved):
Looks good! I confirmed the issues I saw were fixed, and I tried a few other new things that I didn't test the first time around. Everything is working nice and smoothly now, as far as I can see. 👍

Issue #602, opened 2021-03-03
Upgrade Nest.js and related dependencies


maurizi:

pcaisse:
We are motivated to upgrade TypeORM in particular due to this nasty bug (intended behavior?) where if you filter on a non-existent field using it would just remove the WHERE clause entirely and not throw an error, including on updates/deletes: https://github.com/typeorm/typeorm/issues/3416

For a minimal example, copy/paste this into a file at `src/manage/src/commands/test-where-clause-bug.ts`:
```
import { Command } from "@oclif/command";
import { createConnection } from "typeorm";

import { User } from "../../../server/src/users/entities/user.entity";
import { connectionOptions } from "../lib/dbUtils";

export default class TestWhereClauseBug extends Command {
  async run(): Promise<void> {
    const connection = await createConnection(connectionOptions);
    const userRepo = connection.getRepository(User);

    /* tslint:disable:no-object-mutation */
    const result = await userRepo.find({ userId: "07f9966f-facf-42ce-a4ee-48320007f1cc" }); // actual field name is `id`
    this.log("result", result);
  }
}
```

Then run
```
$ ./scripts/manage test-where-clause-bug
```
and you will see all users listed like so:
```
Creating district-builder-2_manage_run ... done
Version 9 of Highlight.js has reached EOL and is no longer supported.
Please upgrade or ask whatever dependency you are using to upgrade.
https://github.com/highlightjs/highlight.js/issues/2877
query: SELECT "User"."id" AS "User_id", "User"."email" AS "User_email", "User"."name" AS "User_name", "User"."isEmailVerified" AS "User_isEmailVerified", "User"."hasSeenTour" AS "User_hasSeenTour", "User"."passwordHash" AS "User_passwordHash" FROM "user" "User"
result [
  User {
    id: '07f9966f-facf-42ce-a4ee-48320007f1cc',
    email: 'pcaisse@azavea.com',
    name: 'Peter Caisse',
    isEmailVerified: false,
    hasSeenTour: true,
    passwordHash: '$2b$10$LE5K5Zm9AJdAPIuBppEo6eEsBQB8XMZ6v1mbvoCYF3ylfqxI39mJi'
  },
  User {
    id: '5cccbc68-06a7-4b6e-a05f-732063bc81fa',
    email: 'test@azavea.com',
    name: 'Test User',
    isEmailVerified: false,
    hasSeenTour: false,
[...]

PR #1099, opened 2022-01-07
Change topojson serialization format

### LINKED ISSUE #1095: Investigate issues loading PA LRC region in staging

maurizi:
## Overview

Upgrading to Node.js 16 in #602 caused a regression in data loading - the maximum size a of string was was reduced from ~1Gb to ~512Mb, which meant we could no longer load our two largest regions by using `JSON.parse(buffer.toString())`

This PR takes a multi-step approach to tackling this problem:
 - I've switched to serialzing the TopoJSON using the v8 binary serialization format: https://github.com/PublicMapping/districtbuilder/commit/c2a0dcb7ff876364e90f534b087b5c91ea9757bc, https://github.com/PublicMapping/districtbuilder/commit/d17738ea29fee0f69c1a90df1e79efec4669dc3e & https://github.com/PublicMapping/districtbuilder/commit/dc52da87c585429013c92c9bd4b7833d107c3f2a
 - Added a command to download the existing TopoJSON files (I used streaming JSON so we can load the now too large files w/o needing to downgrade Node temporarily) https://github.com/PublicMapping/districtbuilder/commit/f8262c278b49a46eb3c1e7e943207ac546b580ec
    - I've already applied this command on all the files used on staging as well as those referenced by `load-dev-data`, but I still need to update production, and any one-off regions will no longer work unless updated
   
After testing this branch on staging, I encountered a lot of resource contention and a large number of layers failed to load the first time, so I added https://github.com/PublicMapping/districtbuilder/commit/807dc9d57910850ca2acb3dd6ba5500b7999d419 & https://github.com/PublicMapping/districtbuilder/commit/e8ea4f103eb3120ea85728ab92e8dbc0e4628c29 to make the `TopologyService` more robust.

### Checklist

### Demo

Updated healthcheck, showing which layers are loading / pending.
image

### Notes

I was hopeful this would decrease our load times, but it hasn't done so to any meaningful degree. We did buy a lot of headroom in terms of TopoJSON file size - the max file size for [`v8.deserialize` is ~4GB](https://nodejs.org/api/v8.html#v8serializevalue).

## Testing Instructions

Try running `./scripts/manage serialize-topojson ...` locally, and make sure you can run the server

I tested this against staging, and was able to get staging to load successfully.

Closes #1095


KlaasH (approved):
Looks good :+1: 
I deleted the `topo.buf` for the NJ config I have in my database, then recreated it and confirmed that it loaded and I could make a map.

Re performance: I found [this comparison of Javascript encoding tools](https://github.com/Adelost/javascript-serialization-benchmark), and [this PR that adds v8.serialize](https://github.com/Adelost/javascript-serialization-benchmark/pull/5) to the mix.  It seems like the tools that are notably faster than `v8.serialize` require a predefined schema, which is probably more trouble than it would be worth for our purposes.

src/server/src/districts/services/topology.service.ts
```diff
@@ -25,25 +61,27 @@ export class TopologyService {
...
-        // Next we load all remaining (unarchived) regions
-        // eslint-disable-next-line
-        for (const s3URI of activeURIs) {
-          this._layers[s3URI] = this.fetchLayer(s3URI, false);
-        }
+        // Load regions a number of regions in parallel, adding as each one completes
+        await asyncLoop(
+          sortedRegions.map(region => () => {
+            const promise = this.fetchLayer(region.s3URI, region.archived);
+            this._layers = { ...this._layers, [region.s3URI]: promise };
```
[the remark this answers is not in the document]
maurizi: Yeah I tried / failed at this as well - good to know it wasn't just me 😅

src/server/src/districts/services/topology.service.ts
```diff
@@ -25,25 +61,27 @@ export class TopologyService {
...
+          !STATE_ORDER.includes(region.regionCode),
+          STATE_ORDER.indexOf(region.regionCode)
+        ]);
 
-        // Next we load all remaining (unarchived) regions
-        // eslint-disable-next-line
-        for (const s3URI of activeURIs) {
-          this._layers[s3URI] = this.fetchLayer(s3URI, false);
-        }
+        // Load regions a number of regions in parallel, adding as each one completes
```
maurizi, line 76:
```suggestion
        // Load a number of regions in parallel, adding another as each one completes
```

src/manage/src/commands/serialize-topojson.ts
```diff
@@ -0,0 +1,124 @@
...
+export default class SerializeTopojson extends Command {
+  static description = `reprocess topojson files into binary format
+  
+  Pass a list of s3_uri paths to reprocess, e.g.
+  serialize-topojson s3://bucket-name/regions/US/PA s3://other-bucket-name/regions/US/DE
+`;
+
+  static strict = false;
+
+  // Streaming reader for GeoJSON files. Works on files over 1GB, but is slow.
```
maurizi, line 20:
```suggestion
  // Streaming reader for GeoJSON files. Works on files over 512MB, but is slow.
```

src/server/src/districts/services/topology.service.ts
```diff
@@ -11,8 +14,41 @@ import { GeoUnitProperties } from "../entities/geo-unit-properties.entity";
 import _ from "lodash";
 import { getObject, s3Options } from "../../common/s3-wrapper";
 
+const MAX_RETRIES = 5;
+// Loading a topojson layer is a mix of I/O and CPU intensive work,
+// so we can afford to have more concurrent than we have cores, but not too many
```
maurizi, line 19:
```suggestion
// so we can afford to have more layers loading than we have cores, but not too many
```

PR #1100, opened 2022-01-10
Switch to a more performant way of calculating maximum district ID

### LINKED ISSUE #1098: Import project not working for some regions
From testing, it seems to be larger regions (or more complex maps?) - I was able to import a DE map, but not a PA or NJ map.

The error in rollbar is: `Maximum call stack size exceeded` @ `/home/node/app/server/src/districts/controllers/districts.controller.ts` line 144 

This is a regression, so my assumption is this was triggered by the Node upgrade, though the likely fix is to refactor the code to use less recursion.

maurizi:
## Overview

Upgrading Node.js on the backend uncovered a lingering performance issue in the CSV import in our use of `Math.max`, which we are using to detect the maximum district ID in the CSV. This caused the import to fail for larger states.

Per MDN, using `Math.max(...array)` on a large array is not recommended, and instead they suggest using reduce:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max#getting_the_maximum_element_of_an_array

Switching to the form suggested by MDN for large arrays fixed the issue.

### Checklist

### Notes

I searched the codebase for similar problematic uses of `Math.min` or `Math.max` and this was the only one case where we were using it on a very large array.

## Testing Instructions

- Attempt to import a map for a larger state like PA or NJ. On `develop` it will fail, on this branch it will succeed.

Closes #1098


ktohalloran (approved):
Works great! 👍

PR #1101, opened 2022-01-11
Run reindex on monthly schedule

### LINKED ISSUE #1086: Add cron job to run concurrent REINDEX on project TABLE on a regular basis
Part of our access patterns involve repeated updates to columns used in a database index, which can lead to [index bloat](https://wiki.postgresql.org/wiki/Index_Maintenance), for which the PostgreSQL docs recommend [routine reindexing](https://www.postgresql.org/docs/current/routine-reindex.html).

We can use AWS ECS task scheduling to take care of running a task for us periodically (from anecdotal experience running `REINDEX` once every few weeks has seemed more than sufficient).

We don't yet have any standalone commands that can be run in our AWS environment (the `oclif` commands in the `manage` directoryhave thus far been designed to be run locally only).

The easiest route here is probably to add a new line to the `scripts` section in `src/server/package.json`, referencing a a new JS / TS file, which then runs `REINDEX .... CONCURRENTLY` - which we should then be able to configure as a task, much like how we run `typeorm` migrations.

This script _could_ be a standalone Nest application: https://docs.nestjs.com/standalone-applications - but given that we're just running a SQL query we might not need Nest

maurizi:
## Overview

Our largest table gets many frequent updates, leading over time to index bloat.
I have been manually running `REINDEX TABLE CONCURRENTLY project`, but we would prefer to do this automatically, not manually.

This PR aims to achieve that by creating a [stand-alone Nest.js application](https://docs.nestjs.com/standalone-applications) which runs `REINDEX`, and then calling that via a scheduled ECS task.

### Checklist

### Notes

 - This is both our first scheduled ECS task as well as our first stand-alone Nest.js application, so I didn't have any internal examples to draw on. Other than the documentation, I also used this [blog post](https://vthub.medium.com/running-ecs-fargate-tasks-on-a-schedule-fd1ca428e669) for inspiration.
 - ~This builds on top of #1099, I'll rebase this and change the base branch the PR is pointing to once that's merged~
 
## Testing Instructions

I'm not quite sure how to test this end-to-end.

I've run the `reindex` script locally, as well as ran `scripts/infra plan` and saw the new resources I expect in the plan output, though I haven't run `scripts/infra apply` yet (nor do I know of a great way to verify this works other than waiting until February)

Closes #1086

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add project-level info to geojson 'metadata' property [#1076](https://github.com/PublicMapping/districtbuilder/pull/1076)
 - Add instructions on configuring PG Admin and on releasing new region data [#1081](https://github.com/PublicMapping/districtbuilder/pull/1081)
 - Add blog link to Resource menu [#1083](https://github.com/PublicMapping/districtbuilder/pull/1083)
+- Added cron task to REINDEX project table monthly [#1101](https://github.com/PublicMapping/districtbuilder/pull/1101)
 
 ### Changed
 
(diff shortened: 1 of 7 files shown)


nanotubing (approved):
I reviewed the reindex related code in this commit, and it seems reasonable to me, though agreed that this is hard to test.

You could potentially create a very small test table and submit a job to reindex it on an hourly basis? That may give you a test job that repeats frequently enough to let you troubleshoot a potential issue, but which puts a trivial load on the server

PR #1104, opened 2022-01-20
Update client dependencies

### LINKED ISSUE #1087: Update client dependencies to maintain parity with server
Update the following to match the versions used in the frontend:
 - TypeScript
 - Node
 - `eslint`
 - `prettier`

Consider updating / update if required for the new TS version:
 - React / CRA

maurizi:
## Overview

Updates the following `client` dependencies:
 - Node.js
 - `eslint`
 - `prettier`

Updating Node required updating:
 - Theme-UI
 - Create React App
   - this lead to me also updating `react-papaparse` & swapping `workerize-loader` for Comlink

### Checklist

### Notes

We did acquire some new deprecation warnings when the React dev server first starts up - as far as I can tell, those are from CRA / one of it's dependencies, not something we're depending on directly.

## Testing Instructions

- `scripts/update`
- Test out the application and look for any regressions

Closes #1087


src/client/actions/organizationJoin.ts
```diff
@@ -5,15 +5,15 @@ export const joinOrganization = createAction("Join organization")<{
   readonly organization: OrganizationSlug;
   readonly user: UserId;
 }>();
-export const joinOrganizationSuccess = createAction("Join organization success")<
```
ddohler, line 8: Is prettier still respecting `.prettierrc`? It seems like these lines have all been shortened.

ddohler (approved):
Looks great! I noticed one file where it seems like prettier may have started using a different line length, which can sometimes be an indication that prettier has lost track of `.prettierrc`, but it could also be different behavior in the newer version (prettier doesn't make guarantees about line length as far as I recall). 

I tried to test pretty extensively and didn't find anything that looked out of the ordinary, nice job! 👏

ddohler:
Forgot to mention in the review that the commits here should be squashed before merging (and Jenkins will need to be appeased as well).

PR #1110, opened 2022-01-26
Add scrolling to map config modal for viewing multiMember list

### LINKED ISSUE #1089: Make Map configuration modal body scroll
When the browser window isn't tall enough, Map configuration modal doesn't display the multimember district list. Seen here on 1280x600 zoomed to 100% in Chrome:
image
When zooming to 80%, multimember district list with scroll bar appears:
image

To fix, the Map configuration modal body elements (Map name, Districts, Population deviation tolerance) should all scroll. The header (Map configuration, Save and cancel buttons) should be sticky.

ktohalloran:
## Overview

The list of multi-member districts was not appearing in the Map configuration modal when the screen height was too short. This PR adds a minimum height to the `multiMemberContainer` to ensure that it always displays regardless of screen size and adds scrolling to the modal body while leaving the title and save/cancel buttons in place.

### Checklist

### Demo
Screen Shot 2022-01-26 at 10 19 50 AM

Screenshot taken from a 1280x680 window zoomed to 100% in Chrome.

### Notes

The added `modalBodyContainer` was given a max height of 100vh-16rem; this amount was determined via experimentation to be what looked best in conjunction with the modal styling when both a normal size and when too short. 

## Testing Instructions

- Run `./scripts/server` and navigate to one of your map project pages.
- Open the Map Configuration modal and shrink the screen to be shorter than the modal's height. It should resize to fit the new height and a scroll bar should appear that allows you to see the modal's body.
- Select "Use Multi-member Districts". The list should appear without changing the dimensions of the modal. 
- Verify the scrolling works as expected.

Closes #1089

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add instructions on configuring PG Admin and on releasing new region data [#1081](https://github.com/PublicMapping/districtbuilder/pull/1081)
 - Add blog link to Resource menu [#1083](https://github.com/PublicMapping/districtbuilder/pull/1083)
 - Added cron task to REINDEX project table monthly [#1101](https://github.com/PublicMapping/districtbuilder/pull/1101)
+- Add scrolling to Map configuration modal for best viewing multi-member district list [#1110](https://github.com/PublicMapping/districtbuilder/pull/1110)
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


maurizi:
@ktohalloran as part of this, we should also _remove_ the scroll from the multi-member section itself.

Sorry for not making it clear in the original issue, but in general a scrolling element inside another scrolling element is something to try to avoid.

ktohalloran:
@maurizi thanks for taking a look at this--I thought the two scrollbars looked funky so I should have inferred that the one inside needed to go away. Thanks for clarifying--I've made that change so it should be ready for re-review.

maurizi (approved):
looking good 👍

PR #1119, opened 2022-02-01
Fix error deleting reference layers

### LINKED ISSUE #1106: Regression in deleting reference layers
On `develop`, attempting to delete a reference layer causes an error that makes the page crash. Upon reload the reference layer is not deleted.

This issue does not exist on production - this is likely a regression from the recent dependency upgrades.

rachelekm:
## Overview

Creates a fix for the reference layers DELETE endpoint so that the page no longer crashes and a reference layer can be successfully deleted. The `ReferenceLayersController` used the `deleteOneBase` method by the `@Crud()` decorator to generate a DELETE endpoint, which ultimately did not match the api call from the frontend:
```
server_1    | [Nest] 39  - 02/01/2022, 9:52:47 PM     LOG [RouterExplorer] Mapped {/api/reference-layer/:id/:projectId, DELETE} route
```
The inconsistency is most likely a regression error after recent Node updates. Attempting to call at this generated endpoint also incorrectly queries the database and still returns a 404 status error. The new DELETE endpoint now matches the frontend call as well as the expected params for the TypeORM entity:
```
server_1    | [Nest] 39  - 02/01/2022, 9:52:47 PM     LOG [RouterExplorer] Mapped {/api/reference-layer/:id, DELETE} route
```

### Checklist

### Demo
Deleting a reference layer
db_prdemo_deletereflayers_small

### Notes

I still decided to use the `deleteOneBase` method since that seemed like the easiest fix, but instead of overriding it I'm creating a new route that will in turn call it. However, this means there will still be a `/api/reference-layer/:id/:projectId` endpoint generated, just no longer used. If called, it will respond with a 404 status.

## Testing Instructions

- `scripts/update`
- Expand the reference layers section and add a csv/geojson file (I used this [geojson data](https://www.pasda.psu.edu/uci/DataSummary.aspx?dataset=1210))
- Toggle layers on, confirm they display, then select to remove layer from map using the "..." menu
- Confirm the layers are successfully deleted and the page does not crash

Closes #1106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fix 404 error when deleting reference layers [#1119](https://github.com/PublicMapping/districtbuilder/pull/1119)
 - Fix Reference Layer copy on other user's maps to reflect map's readonly status [#1078](https://github.com/PublicMapping/districtbuilder/pull/1078)
 - Fix Equal Population status so red "X" doesn't show if within deviation threshold [#1084](https://github.com/PublicMapping/districtbuilder/pull/1084)
 - Fix disappearing user data on Community Maps page refresh [#1090](https://github.com/PublicMapping/districtbuilder/pull/1090)
--- a/src/server/src/reference-layers/controllers/reference-layers.controller.ts
+++ b/src/server/src/reference-layers/controllers/reference-layers.controller.ts
@@ -46,7 +46,6 @@ import { ProjectVisibility } from "../../../../shared/constants";
     },
     projectId: {
       type: "string",
-      primary: true,
       field: "id"
     }
   },


maurizi (changes requested):
Good call on looking at the `RouterExplorer` logs - totally failed to take note of the change there! And this fix does seem to be working.

That said, I _think_ I've found a better way to handle this, that gets Nest CRUD to make correct routes to start with:
```diff
diff --git a/src/server/src/reference-layers/controllers/reference-layers.controller.ts b/src/server/src/reference-layers/controllers/reference-layers.controller.ts
index de7f7eef..10852958 100644
--- a/src/server/src/reference-layers/controllers/reference-layers.controller.ts
+++ b/src/server/src/reference-layers/controllers/reference-layers.controller.ts
@@ -47,7 +47,6 @@ import { ProjectVisibility } from "../../../../shared/constants";
     },
     projectId: {
       type: "string",
-      primary: true,
       field: "id"
     }
   },
@@ -183,9 +182,7 @@ export class ReferenceLayersController implements CrudController<ReferenceLayer>
 
   @UseInterceptors(CrudRequestInterceptor)
   @UseGuards(OptionalJwtAuthGuard)
-  // deleteOneBase generates `/:id/:projectId` route
-  // create new DELETE route at `/:id`
-  @Delete("/:id")
+  @Override()
   async deleteOne(
     @Param("id") id: ReferenceLayerId,
     @ParsedRequest() req: CrudRequest
```

What I think is happening here is that we shouldn't have had `projectId` as a `primary` param in the first place, though it could also be that the meaning of what a `primary` param is has changed in the more recent version of `nestjsx/crud`?

maurizi (approved):

rachelekm:
Thanks @maurizi, this is a much better fix and I implemented that change. Looking at the docs, multiple `primary` params to generate routes were introduced [here](https://github.com/nestjsx/crud/pull/565/files#diff-1b3296236a76cc2e63c59fed4251b70162913dd6dc759629f6aad1bf6e6fabfcL297-L299), before we added reference layers at all, so you're right in that there must have been some change in how `primary` params are defined in `nextjs/crud`. From a quick glance this is our only controller with multiple `params` so this shouldn’t affect our other routes using the `@Crud()` decorator.

PR #1127, opened 2022-02-08
Fix incorrect majority race assignment in map and evaluate

### LINKED ISSUE #1111: Incorrect assignment of majority race value for majority minority districts
The majority-minority Evaluate viewer is incorrectly assigning the majority race value, as well as drawing the map choropleth with the wrong value.

Seen here in the Alhambra region on staging:
image
The district is 48% Hispanic and the Majority Race should be **Minority Coalition**, but it is classified as Hispanic. The choropleth is also showing the district as having a greater than 65% Hispanic population which is incorrect.
Map here: https://app.staging.districtbuilder.org/projects/8c47597f-c3b0-4ccc-a21d-b05db62f95b5

ktohalloran:
## Overview

The Majority-Minority evaluate screen was identifying and drawing the map choropleth incorrectly for the majority race. This issue was also seen in the `ProjectSidebar` when majority race was pinned.

### Checklist

### Demo

Screen Shot 2022-02-08 at 9 55 30 AM
The screen now correctly identifies that the first two districts listed are minority-coalition districts over 65%, rather than a majority Hispanic county. It also correctly identifies that the Asian populations in districts 3 and 4 make up less than 65% and are therefore filled with a lighter color.

## Testing Instructions

- Start the server and navigate to a project screen
- Create a few new districts or use existing districts that match the following criteria: one with a majority race of over 65%, one with a majority race of under 65%, one with a majority-minority coalition over 65%, one with a majority-minority coalition that does not exceed 65%
- Navigate the project evaluate screen and verify that all 4 districts are displaying the proper majority race and that the choropleth matches the percentage of that race
- Change the population to CVAP or VAP and navigate to the project evaluate screen to verify it still assigns as expected

Closes #1111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix import performance on larger regions [#1100](https://github.com/PublicMapping/districtbuilder/pull/1100)
 - Only show All, VAP & CVAP options for population choice [#1122](https://github.com/PublicMapping/districtbuilder/pull/1122)
 - Fix hot reloading on environments using Vagrant [#1126](https://github.com/PublicMapping/districtbuilder/pull/1126)
+- Fix incorrect assignment of majority race on majority-minority evaluate screen [#1127](https://github.com/PublicMapping/districtbuilder/pull/1127)
 
 ## [1.13.0] - 2021-11-30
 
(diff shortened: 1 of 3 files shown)


maurizi (approved):
Looks good 👍 

I think was caused because I moved some duplicative code into a helper method, but we were inconsistent in whether `percent` meant something in the range of 0 -1 or 0 - 100.

I searched in the codebase for other instances where we may have been inconsistent like that but I couldn't find any.

Issue #1129, opened 2022-02-10
Scheduled REINDEX rule not applied to production


maurizi:
The recent production release `1.14.0` included changes meant to run a REINDEX query on a monthly schedule (https://github.com/PublicMapping/districtbuilder/pull/1101).

However, the rule created on AWS only seems to apply to the staging site, and not production: https://console.aws.amazon.com/events/home?region=us-east-1#/eventbus/default/rules/reindex-scheduled-ecs-event-rule
![image](https://user-images.githubusercontent.com/4432106/153414719-41f72a3c-6832-41d4-b2e0-309ce65d39fb.png)

maurizi:
I think the issue here is that were supposed to parameterize the rule name by environment? (i.e. `reindex-scheduled-ecs-event-rule-staging` / `reindex-scheduled-ecs-event-rule-prod`).

I'm not sure on that though.

BryanQuigley:
Despite my PR.. I'm still going to look into pg_cron:
 * Needs pg_cron extension added to avavea postgis container
 * Might need terraform upgrade as I've found an open issue (https://github.com/hashicorp/terraform-provider-aws/issues/12025) - reproducing or confirming it's fixed might be a good first step
 * needs migration to actually set it up

PR #1130, opened 2022-02-10
parameterize rules so both staging and prod can exist

BryanQuigley:
## Overview

This should allow us to tell the difference between prod and staging for reindexing.

### Checklist

## Testing Instructions

- this has to just be deployed everywhere?

Closes: #1129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+ - parameterize REINDEX rules so both staging and prod can exist [#1130](https://github.com/PublicMapping/districtbuilder/pull/1130)
 
 ## [1.14.0] - 2022-02-09
 
--- a/deployment/terraform/scheduled_tasks.tf
+++ b/deployment/terraform/scheduled_tasks.tf
@@ -1,11 +1,11 @@
 resource "aws_cloudwatch_event_rule" "reindex_task" {
-  name                = "reindex-scheduled-ecs-event-rule"
+  name                = "reindex-scheduled-ecs-event-rule-${var.environment}"
   // Run monthly at 5AM UTC (midnight EST)
   schedule_expression = "cron(0 5 1 * ? *)"
 }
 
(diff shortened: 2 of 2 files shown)


BryanQuigley:
This looks mostly correct to me. I don't like that both staging and prod are trying to rename the same rules, but I think it will go like:
1. Staging is deployed as above
2. Prod when deployed ends up creating all new infra. (as above it's renaming anything with staging in it to prod *or* it's changing the name to (known after apply). )

ddohler:
> Prod when deployed ends up creating all new infra. (as above it's renaming anything with staging in it to prod or it's changing the name to (known after apply). )

I _think_ this sounds right. Right now both staging and production are referencing the same ARN in their Terraform state, so when we alter that resource, the plan updates based on the current state and both plans try to alter the same resource. But once staging is deployed, the ARN in the production state will no longer point to any existing resource (because the staging deploy will have changed it). Then I think a subsequent production deploy _should_ detect that and create the resource from scratch instead. But I'm operating from limited experience here, and I'm not sure how optimistic `terraform plan` is about whether the state of the world matches what it has stored in `state` -- it might be necessary to run something like `terraform apply --refresh-only` to update the production state before running the production deploy.

Alternatively, another route might be to do an intermediate deploy to remove the resource first, and then add it back in with the proper parameterization. That would still encounter the same issue with states potentially getting out of sync across the shared resource, but I'm assuming that Terraform is more likely to shrug and carry on if a thing it wants to delete doesn't exist than if a thing it wants to alter doesn't.

maurizi (approved):
> Prod when deployed ends up creating all new infra. (as above it's renaming anything with staging in it to prod or it's changing the name to (known after apply). )

As we discussed, a simple path forward on testing this is to merge it and let CI redeploy ~production~ staging, and then we can double-check if applying this to production will result in a new rule.

Before merging - don't forget to update the `CHANGELOG` with a reference to this PR

BryanQuigley:
Confirmed on production that with the rename on staging new rules/targets will be created.
# aws_cloudwatch_event_rule.reindex_task will be created
 # aws_cloudwatch_event_target.reindex_task will be created

PR #1132, opened 2022-02-10
Stop loading archived topology

### LINKED ISSUE #1118: Stop loading archived topology layers
On application startup we load archived 2010 topology layers in order to keep their feature properties in memory, to support converting to 2020 maps, CSV exports, as well to load county names (or other top-level geo-unit names) on the Evaluate page.

This still takes a bit of memory to support, so we'd like to stop needing to load even this smaller dataset into memory. We are willing to drop support for converting and exporting to CSV, but ideally we would retain the ability to show county names in Evaluate mode

AC:
 - Remove Convert button for maps in archived regions
 - Hide any unsupported export options from export menu
 - Refactor County Split evaluate detail to not require full topology properties loaded on the server (or if this is too difficult, hide the section)
 - Stop loading `archived` layers in `TopologyService`

rachelekm:
## Overview

This PR stops loading archived topology layers as part of `TopologyService`. While we initially retained the topology properties and let the larger geometries get garbage collected, loading the `topo.json` and keeping the properties in memory still takes up space that would be worth freeing up. Archived, read-only maps relied on topology properties for the following:

- Converting from 2010 to 2020 census data
- Exporting to csv
- Copying to account
- Fetching sorted county names used in Evaluate’s County Split metric

Now, users with archived map projects should no longer be able to convert census data, export to csv, or copy an archived, read-only project to their account. An error message will display if a user tries to import a csv with an archived region. However, archived region county names were added as a constant variable on the frontend and should display as before.

### Checklist

### Demo
db_prdemo_archivedtopo_small

### Notes

- In addition to deleting the button to allow users to convert their 2010 census data to 2020 data, I got rid of its endpoint and any related code that was isolated to this functionality. This included the `Crosswalk` service/entity to map new blocks from old—however, I did not get rid of the crosswalk s3 folder `/crosswalk-files/` as this seemed like an action more difficult to reverse should we bring converting census data back.

- The County Split metric’s county names for archived regions are now displayed by referencing a `key: value` pair in the `archivedCountyNames` constant. The values were created for each archived region in the production database using the below command to mimic the topology properties resource generated from the backend (and preserve sorting): 

```
aws --profile district-builder s3 cp {S3_URI}/topo.json - | jq -r '[.objects.county.geometries | .[] | {"name": .properties.name}]' | pbcopy
```

- I had to update our server tests since we now can’t import and process csv files for archived regions. Ideally, the test would use a region with 2020 census data, but I ran into some trouble with the tests still failing after updating the test `region.s3URI` and the `de.csv` data. I can successfully import the same csv through the client and, for the time being, using 2010 data and defining the test region as `archived=false` seems to do the trick, so this may be a separate issue if we want to update the data for server tests. 

## Testing Instructions

- `scripts/server`

- Create a project for `DE`, since we have both 2010 and 2020 census data in our dev database

- Using `scripts/dbshell`, update `ARCHIVED = TRUE` for the region using 2020 data

- Your project should now have a small info box at the top-left of your map alerting you that you are using archived data

- Confirm the following:

  - The info box does not include a “Convert to 2020” button

  - On “export” an option to export to csv should no longer exist

  - There should no longer be a “Copy to account” button

  - On Evaluate confirm there is no toast error message 

  - Navigate to the County Split Evaluate detail and confirm that county names are displaying appropriately

  - Navigate back to the homepage and select the left-hand options menu for your archived project and confirm there is no option to export to csv
  
- Ensure the above features are still available for projects with non-archived regions

Closes #1118


src/server/src/districts/services/topology.service.ts
```diff
@@ -70,16 +70,18 @@ export class TopologyService {
           }),
           {}
         );
+        // Filter out archived regions
```
rachelekm, line 73: I wanted to chain the `filter` and `sortBy` lodash methods but couldn't quite figure it out without `sortedRegions` losing its structure, so ended up leaving these separate

src/server/src/districts/services/topology.service.ts
```diff
@@ -141,17 +139,16 @@ export class TopologyService {
...
           topology,
           { groups: geoLevelHierarchy.slice().reverse() },
           staticMetadata,
           demographics,
           voting,
           geoLevels
         );
-        // For archived read-only regions, we get the properties of the topology (which are used for exports)
-        // and let the much larger TopoJSON geometries get garbage collected
-        return archived ? GeoUnitProperties.fromTopology(geoUnitTopology) : geoUnitTopology;
```
maurizi, line 154: I think we can get rid of the `GeoUnitProperties` class in it's entirety now.

src/server/src/districts/services/topology.service.ts
```diff
@@ -70,16 +70,18 @@ export class TopologyService {
           }),
           {}
         );
+        // Filter out archived regions
+        const filteredRegions = _.filter(regionConfigs, ["archived", false]);
```
maurizi, line 74: Sometimes you need explain what you're doing in a comment because it isn't immediately obvious what the code does, but usually when I see myself writing a "this is what the code does" comment I try and see if renaming some variables / rewriting the code to be more self-explanatory would remove the need for it.

```suggestion
        const unarchivedRegions = _.filter(regionConfigs, region => !region.archived]);
```

src/client/components/evaluate/ProjectEvaluateSidebar.tsx
```diff
@@ -65,9 +68,13 @@ const ProjectEvaluateSidebar = ({
...
-      store.dispatch(
-        regionPropertiesFetch({ regionConfigId: project.regionConfig.id, geoLevel: geoLevel })
-      );
+      !isArchived
+        ? store.dispatch(
+            regionPropertiesFetch({ regionConfigId: project.regionConfig.id, geoLevel: geoLevel })
+          )
+        : store.dispatch(
+            regionPropertiesFetchSuccess(archivedCountyNames[`${project.regionConfig.regionCode}`])
+          );
```
maurizi, line 77: This is a nice simple way to handle this - I like it 😄

maurizi (changes requested):
This is looking really good! Great job 👍 

The major thing that I think still needs to happen here is that there is more code we should be able to remove - the `GeoUnitProperties` class was only used for archived regions but I still see it referenced in a few places in the code.

rachelekm:
Thanks, @maurizi! Implemented these changes and there shouldn't be any more references to the `GeoUnitProperties` class.

maurizi (approved):
Working great 👍

PR #1145, opened 2022-02-16
Fix bounds for AK mini-map

### LINKED ISSUE #992: Alaska mini-map bounds are off
image

# Notes
 For the main map we use the `bounds` specified in `static-metadata.json` for the region. We don't currently download the `static-metadata.json` for the mini-map display _or_ use it server-side, but we _do_ have it available in-memory on the server as part of the `TopologyService` so it might make sense to attach a `bounds` property to the data returned by the project list pages.

rachelekm:
## Overview

This PR fixes the incorrect bounds generated for the Alaska mini-map displayed on the home screen. This error was initially caused by an issue with the `@turf/bbox` module—the output for `bbox` included incorrect longitudes coordinates for Alaska’s bounding box, assumed to be an issue with the geojson spanning the anti-meridian. For a more reliable fix, this PR instead uses the static bounds specified in `static-metadata.json`, similar to the main map for a project.

### Checklist

### Demo

Screen Shot 2022-02-16 at 11 39 46 AM

### Notes

The notes of Issue #992 mention that, since the `static-metadata.json` is available in-memory as part of `TopologyService`, it might be worth creating a `bounds` property that can be sent back for mini-map generation. However, we recently got rid of archived topology layers within `TopologyService`, which means this would not be an available option for any projects with archived regions. To accommodate archived projects, the `s3URI` property is attached to the data returned to project list pages instead, which is then used by `fetchStaticMetadata` in a memoized function for each project. While `static-metadata.json` is small, if we want to free up some space we could potentially create the `bounds` property using `TopologyService` by default and return `s3URI` instead for projects that are archived?

## Testing Instructions

- `scripts/server`
- Create a project for `AK`
- Confirm the mini-map bounds are correct on home screen and the community maps page

Closes #992

### DIFF
--- a/src/client/api.ts
+++ b/src/client/api.ts
@@ -1,5 +1,6 @@
 import axios, { AxiosResponse } from "axios";
 import { saveAs } from "file-saver";
+import memoize from "memoizee";
 
 import {
   CreateProjectData,
@@ -22,7 +23,8 @@ import {
   ReferenceLayerId,
   CreateReferenceLayerData,
   IProjectTemplate,
-  CreateProjectTemplateData
+  CreateProjectTemplateData,
+  IStaticMetadata
 } from "../shared/entities";
 import {
   DistrictsGeoJSON,
@@ -402,6 +404,16 @@ export async function fetchOrganizationFeaturedProjects(
       });
   });
 }
+
+export const fetchMemoizedStateBbox = memoize(
+  async (region: IRegionConfig): Promise<IStaticMetadata["bbox"]> => {
+    const staticMetadata = await fetchStaticMetadata(region.s3URI);
+    return staticMetadata.bbox;
+  },
+  {
+    normalizer: (args: [region: IRegionConfig]) => JSON.stringify(args[0].id)
+  }
+);
 
 export async function exportOrganizationProjectsCsv(slug: OrganizationSlug): Promise<void> {
   return new Promise((resolve, reject) => {
--- a/src/client/components/map/ProjectDistrictsMap.tsx
+++ b/src/client/components/map/ProjectDistrictsMap.tsx
@@ -2,13 +2,11 @@
 import { Box, jsx, Spinner, ThemeUIStyleObject } from "theme-ui";
 import MapboxGL from "mapbox-gl";
 import React, { useEffect, useRef, useState } from "react";
-import bbox from "@turf/bbox";
 
-import { BBox2d } from "@turf/helpers/lib/geojson";
-
-import { ProjectNest } from "../../../shared/entities";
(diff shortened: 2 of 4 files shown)


src/client/components/map/ProjectDistrictsMap.tsx
```diff
@@ -35,13 +33,21 @@ const ProjectDistrictsMap = ({
 }) => {
   const mapRef = useRef<HTMLDivElement>(null);
   const [mapLoaded, setMapLoaded] = useState<boolean>(false);
+  const [bounds, setBounds] = useState<[number, number, number, number] | null>(null);
+  project.regionConfig &&
+    !bounds &&
+    fetchMemoizedStateBbox(project.regionConfig).then((bboxData: IStaticMetadata["bbox"]) => {
+      // Conversion from readonly -> mutable to match Mapbox interface
+      const [b0, b1, b2, b3] = bboxData;
+      setBounds([b0, b1, b2, b3]);
```
maurizi, line 42: This would work as well to go from read-only -> mutable, and is less brittle (it doesn't need to know what the structure of `bboxData` is:
```suggestion
      setBounds([...bboxData]);
```

src/client/api.ts
```diff
@@ -403,6 +405,16 @@ export async function fetchOrganizationFeaturedProjects(
   });
 }
 
+export const fetchMemoizedStateBbox = memoize(
+  async (region: IRegionConfig): Promise<IStaticMetadata["bbox"]> => {
+    const staticMetadata = await fetchStaticMetadata(region.s3URI);
+    return staticMetadata.bbox;
+  },
+  {
+    normalizer: (args: [region: IRegionConfig]) => JSON.stringify(args[0].name)
```
maurizi, line 414: `RegionConfig.name` is not required to be unique (https://github.com/PublicMapping/districtbuilder/blob/334331cddc6d2c9860699917fc02ada3ccbd4a74/src/server/src/region-configs/entities/region-config.entity.ts#L22-L23) so we shouldn't rely on that for the `memoize` cache key. Using `id` should work though:
```suggestion
    normalizer: (args: [region: IRegionConfig]) => JSON.stringify(args[0].id)
```

src/client/api.ts
```diff
@@ -403,6 +405,16 @@ export async function fetchOrganizationFeaturedProjects(
   });
 }
 
+export const fetchMemoizedStateBbox = memoize(
```
maurizi, line 408: Great use of `memoize` here 💡

maurizi (changes requested):
Just a few small cleanups, but I tested this and it's working well

maurizi (approved):

PR #1146, opened 2022-02-17
Use separate node_modules cache per executor in CI

### LINKED ISSUE #979: Jenkins builds intermittently failing
Our Jenkins builds will sometimes fail when running `scripts/cibuild` with the following error:
```
$ tsc
error TS2688: Cannot find type definition file for 'papaparse'.
error Command failed with exit code 1.
```

Rerunning the build seems to have fixed the issue each time this has happened.

maurizi:
## Overview

This aims to fix perennial issues where concurrent builds would cause mysterious Jenkins failures where dependencies that had been installed would then go "missing" later in the build

I believe our issue was caused by caching `node_modules` as a Docker volume on the host to enable faster re-builds.
In development this was fine, but on CI any concurrent builds could stomp on each others NPM dependencies. 

Using the `EXECUTOR_NUMBER` as part of the volume path is intended to avoid this, by giving each executor it's own set of `node_modules` folders to use.

### Checklist

## Testing Instructions

I'm going to `ssh` into Jenkins after it builds this PR, and check to see if the new cache directory has the executor number in it.

I'm not going to try to recreate the concurrent build issue, because that sounds like more trouble than it's worth for a simple fix like this - just going to keep an 👀 out in case this doesn't fix it.

Closes #979

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Duplicated maps keep reference layers [#1137](https://github.com/PublicMapping/districtbuilder/pull/1137)
 - Fix bounds for AK mini-map [#1145](https://github.com/PublicMapping/districtbuilder/pull/1145)
 - Fix TopologyService health check[#1148](https://github.com/PublicMapping/districtbuilder/pull/1148)
+ - Fix concurrent Jenkins builds breaking node dependencies [#1146](https://github.com/PublicMapping/districtbuilder/pull/1146)
 
 ## [1.14.0] - 2022-02-09
 
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -2,9 +2,19 @@ version: "2.4"
 services:
   server:
     image: "districtbuilder:${GIT_COMMIT:-latest}"
+    volumes:
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER}-server-node-modules:/home/node/app/server/node_modules"
+
+  client:
+    image: node:16-slim
+    volumes:
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER}-client-node-modules:/home/node/app/node_modules"
 
   manage:
     image: "districtbuilder-manage:${GIT_COMMIT:-latest}"
+    volumes:
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER}-server-node-modules:/home/node/app/server/node_modules"
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER}-manage-node-modules:/home/node/app/manage/node_modules"
 
   shellcheck:
     image: koalaman/shellcheck:stable
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -23,52 +23,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
-        # Ensure container images are current
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            build --pull
-
-        # Clean dist directory
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps server \
-            clean
-
-        # Update frontend, Yarn dependencies and build static asset bundle
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps client \
-            bash -c "yarn install && yarn compile && yarn build"
-
-        # Update backend, Yarn dependencies and build server
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps --entrypoint "bash -c" server \
-            "yarn install && npm rebuild && yarn build"
-
-        # Update manage, Yarn dependencies and build
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps manage \
-            bash -c "yarn install && yarn build"
-
-        # Bring up PostgreSQL and NestJS in a way that respects
-        # configured service health checks.
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            up -d database server
-
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm server \
-            migration:run
+        CI=1 ./scripts/update
 
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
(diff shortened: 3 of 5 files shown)


maurizi:
On CI, after running this a few times I see these new directories in `/var/cache/`:
```
azavea@urbanappsci:~$ /var/cache/district-builder-
district-builder-0-client-node-modules/ district-builder-2-client-node-modules/
district-builder-0-manage-node-modules/ district-builder-2-manage-node-modules/
district-builder-0-server-node-modules/ district-builder-2-server-node-modules/
```
There was an existing set of cached `node_modules` without the executor number, which I deleted and then checked to see that they were not recreated when I reran the CI build.

maurizi:
@KlaasH assigning you bc as I noted this might also be an issue for PFB.

I don't think you need to test anything, but if you have any shell wizardry to make this less verbose I'm interested

maurizi:
It feels like we should be threading `docker-compose.ci.yml` through everywhere in CI anyways, but maybe I'm making this overly complex....

I think we could do this instead and it would also work
```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index fdad8429..2c5ad1ad 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,7 @@ services:
     volumes:
       - ./src/server:/home/node/app/server
       - ./src/shared:/home/node/app/shared
-      - /var/cache/district-builder-server-node-modules:/home/node/app/server/node_modules
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER:-}server-node-modules:/home/node/app/server/node_modules"
       - $HOME/.cache/district-builder:/usr/local/share/.cache/yarn
       - $HOME/.aws:/root/.aws:ro
     working_dir: /home/node/app/server
@@ -51,7 +51,7 @@ services:
     volumes:
       - ./:/home/node/app
       - ./src/server/dist/server/src/static:/home/node/app/build
-      - /var/cache/district-builder-client-node-modules:/home/node/app/node_modules
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER:-}client-node-modules:/home/node/app/node_modules"
       - $HOME/.cache/district-builder:/usr/local/share/.cache/yarn
     working_dir: /home/node/app
     command: yarn start
@@ -73,8 +73,8 @@ services:
       - ./src/server:/home/node/app/server
       - ./src/shared:/home/node/app/shared
       - ./src/manage:/home/node/app/manage
-      - /var/cache/district-builder-server-node-modules:/home/node/app/server/node_modules
-      - /var/cache/district-builder-manage-node-modules:/home/node/app/manage/node_modules
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER:-}server-node-modules:/home/node/app/server/node_modules"
+      - "/var/cache/district-builder-${EXECUTOR_NUMBER:-}manage-node-modules:/home/node/app/manage/node_modules"
       - $HOME/.cache/district-builder:/usr/local/share/.cache/yarn
       - $HOME/.aws:/root/.aws:ro
     working_dir: /home/node/app/manage
```

I guess it raises the question of why we have `docker-compose.ci.yml` ?

KlaasH (approved):
This makes sense and seems like a straightforward and efficient way of keeping Jenkins from stepping on its own toes.

> It feels like we should be threading docker-compose.ci.yml through everywhere in CI anyways

Yeah, I agree with that. There are a few different things happening in `docker-compose.ci.yml`, so it seems like an oversight/bug that some of the `cibuild` commands weren't using it.

It is kind of a bummer to reproduce most of `scripts/update` in `cibuild` with just that one change.  One alternative could be to pass a variable, like what `cibuild` does further down when calling `test`, and have `update` look for that and set its config args accordingly:
```diff
diff --git a/scripts/cibuild b/scripts/cibuild
index 338c48a..6992f3e 100755
--- a/scripts/cibuild
+++ b/scripts/cibuild
@@ -23,52 +23,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
-        # Ensure container images are current
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            build --pull
-
-        # Clean dist directory
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps server \
-            clean
-
-        # Update frontend, Yarn dependencies and build static asset bundle
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps client \
-            bash -c "yarn install && yarn compile && yarn build"
-
-        # Update backend, Yarn dependencies and build server
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps --entrypoint "bash -c" server \
-            "yarn install && npm rebuild && yarn build"
-
-        # Update manage, Yarn dependencies and build
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm --no-deps manage \
-            bash -c "yarn install && yarn build"
-
-        # Bring up PostgreSQL and NestJS in a way that respects
-        # configured service health checks.
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            up -d database server
-
-        docker-compose \
-            -f docker-compose.yml \
-            -f docker-compose.ci.yml \
-            run --rm server \
-            migration:run
+        CI=1 ./scripts/update
 
         # Build tagged container images
         GIT_COMMIT="${GIT_COMMIT}" docker-compose \
diff --git a/scripts/update b/scripts/update
index 34ffe0a..c5c0208 100755
--- a/scripts/update
+++ b/scripts/update
@@ -17,32 +17,38 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        if [[ -n "${CI}" ]]; then
+            CONFIG_ARGS=( "-f" "docker-compose.yml" "-f" "docker-compose.ci.yml" )
+        else
+            CONFIG_ARGS=( "-f" "docker-compose.yml" )
+        fi
+
         # Ensure container images are current
-        docker-compose build --pull
+        docker-compose "${CONFIG_ARGS[@]}" build --pull
 
         # Clean dist directory
-        docker-compose \
+        docker-compose "${CONFIG_ARGS[@]}" \
             run --rm --no-deps server \
             clean
 
         # Update frontend, Yarn dependencies and build static asset bundle
-        docker-compose \
+        docker-compose "${CONFIG_ARGS[@]}" \
             run --rm --no-deps client \
             bash -c "yarn install && yarn compile && yarn build"
 
         # Update backend, Yarn dependencies and build server
-        docker-compose \
+        docker-compose "${CONFIG_ARGS[@]}" \
             run --rm --no-deps --entrypoint "bash -c" server \
             "yarn install && npm rebuild && yarn build"
 
         # Update manage, Yarn dependencies and build
-        docker-compose \
+        docker-compose "${CONFIG_ARGS[@]}" \
             run --rm --no-deps manage \
             bash -c "yarn install && yarn build"
 
         # Bring up PostgreSQL and NestJS in a way that respects
         # configured service health checks.
-        docker-compose \
+        docker-compose "${CONFIG_ARGS[@]}" \
             -f docker-compose.yml \
             up -d database server
```

The `CONFIG_ARGS` syntax isn't exactly graceful (I googled up how to actually do it properly), but I like that it avoids making a modified copy of the build steps from `update` that's hard to compare (and therefore hard to keep in sync) with the original.

Marking this "Approve", though, because it's fine either way :+1:

PR #1148, opened 2022-02-17
Fix TopologyService health check

### LINKED ISSUE #1147: Fix TopologyService health check
We recently updated `TopologyService` so that we no longer support loading layers for archived regions, as part of PR #1132. However we did not update the health check, which consequently broke in staging. We need to refactor health check to expect loaded layers for unarchived regions only.

rachelekm:
## Overview

`TopologyService` loads layers by first setting the default value of the key:value pair in `this._layers` to `null` for each region, which later updates to a promise to fetch layers. However, archived regions are filtered out before this loop so their values are left as the default `null`, which fails the health check and consequently broke staging. This PR updates the way in which `TopologyService` filters out regions with archived layers so that the initial regions that populate `this._layers` no longer contain archived regions.

### Checklist

## Testing Instructions

- `scripts/server`
- Wait for all region data to download in the server
- Call the health check endpoint in the browser: `localhost:3005/healthcheck`
- Confirm successful health check

Closes #1147

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+
 - Add modal to allow organization admins to create templates from projects [#1107](https://github.com/PublicMapping/districtbuilder/pull/1107)
 
 ### Changed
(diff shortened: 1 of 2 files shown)


maurizi (approved):
🎉

PR #1150, opened 2022-02-21
Add color picker to reference layer fly out

### LINKED ISSUE #998: Reference layer color picker
Allow selecting from a preset list of colors:
image

AC:
 - New field added to `ReferenceLayer` entity to store color choice. This should be an `enum`, with the actual RGB value for the colors defined in the frontend code, for flexibility in changing the styling in the future.
 - New option added to reference layer flyout menu to set the color, which should persist to the database

ktohalloran:
## Overview

This PR adds the ability to change a reference layer's color. This change is made via the reference layer flyout and is saved to the database.

### Checklist

### Demo

color_picker_demo

## Testing Instructions

- Start the server and open a project with existing reference layers. Verify that those existing layers display as they should and that the color picker works to change their colors.
- Create new layer, toggle it on, change its color, and verify that both the color of the icon and the layer itself change to the new color. Also verify that the selection marker in the flyout has moved to the correct color.
- Verify that on color change the layers stay in alphabetical order that no flickering occurs.

Closes #998


src/shared/constants.ts
```diff
@@ -54,6 +54,15 @@ export enum ReferenceLayerTypes {
...
 }
 
+export enum ReferenceLayerColors {
+  Green = "#17c42f",
+  Orange = "#f3ad37",
+  Purple = "#7549f6",
+  Blue = "#0018ff",
+  Pink = "#eaa0bd",
+  Red = "#f62756"
+}
```
maurizi, line 64: We _don't_ want this. This ends up with the database storing `#17c42f` or other hex codes in the `color` column: https://github.com/PublicMapping/districtbuilder/pull/1150/files#diff-41f1bd333faff04c6cfda50343542e0c2ae21be090242270c30477bd025c45a6R7

Per the issue description:
> This should be an enum, with the actual RGB value for the colors defined in the frontend code, for flexibility in changing the styling in the future.

We should change this enum to not reference hex codes like so, and then recreate the migrations:
```suggestion
export enum ReferenceLayerColors {
  Green = "GREEN",
  Orange = "ORANGE",
  Purple = "PURPLE",
  Blue = "BLUE",
  Pink = "PINK",
  Red = "RED"
}
```

We can move the hex codes to `src/client/constants/colors.ts`; since the server should never need/use the color hex codes, putting them in `src/client` instead of `src/shared` ensures it _can't_ use them.

Something like this should work to get the color code from the enum value:
```typescript
const REFERENCE_LAYER_COLOR_CODES = {
  [ReferenceLayerColors.Green]: "#17c42f",
  [ReferenceLayerColors.Orange]: "#f3ad37",
  [ReferenceLayerColors.Purple]: "#7549f6",
  [ReferenceLayerColors.Blue]: "#0018ff",
  [ReferenceLayerColors.Pink]: "#eaa0bd",
  [ReferenceLayerColors.Red]: "#f62756"
};
```

src/server/src/reference-layers/entities/update-reference-layer.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @IsOptional()
+  readonly project: ProjectIdDto;
+  @IsNotEmpty({ message: "Please enter a layer type" })
+  @IsOptional()
+  readonly layer_type: ReferenceLayerTypes;
+  @IsOptional()
+  readonly label_field?: string;
+  @IsNotEmpty({ message: "Please enter a value field" })
+  @IsOptional()
+  readonly layer: ReferenceLayerGeojson;
```
maurizi, line 22: We don't want to allow updates to these fields - there are no UI mechanisms to update them, and in any case we would never want to allow the user to change things like project ID, layer content or layer type without more careful validation than merely checking that it exists - otherwise a bug in our client code or a malicious user of the API could cause us to have invalid data stored in the database.

Since we're using `PATCH` we're only sending the `layer_color` field anyway.

```suggestion
```

src/server/src/reference-layers/entities/update-reference-layer.dto.ts
```diff
@@ -0,0 +1,26 @@
...
+  @IsNotEmpty({ message: "Please choose a layer color" })
+  @IsOptional()
+  readonly layer_color: ReferenceLayerColors;
```
maurizi, line 25: See the comment above in create dto re: `@isEnum`

src/client/components/map/Map.tsx
```diff
@@ -734,6 +735,22 @@ const DistrictsMap = ({
...
+        ) {
+          fetchOneReferenceLayer(layer.id).then(result => {
+            map.setPaintProperty(
+              getRefLayerLayerId(layer.id),
+              result.layer_type === ReferenceLayerTypes.Polygon ? "line-color" : "icon-color",
+              result.layer_color
+            );
+          });
+        }
+      });
```
maurizi, line 753: If I look at the developer network tools, changing the color triggers 4 API requests, each of which return the updated geojson.

It looks like that is this:
 - The initial PATCH, which returns the full geojson
 - Once for this line here https://github.com/PublicMapping/districtbuilder/pull/1150/files#diff-e83fa137620a096633593c2ac8c432be0fb237d97c96157a7ab3e4a03c5246f8R53
 - And then _twice_ for this call to `fetchOneReferenceLayer`
image

I think we can simplify things a good deal here:
 - Delete the above section
 - Delete the call to `projectReferenceLayersFetch` in `ReferenceLayerFlyout`
 - Move the `patchReferenceLayer` call into a new action/reducer, and use the data returned by the PATCH call to update the redux store - no further API calls needed
 
 As a bonus, moving the call to `patchReferenceLayer` into the `projectData` reducer along with the other reference layer related ones means it has ready access to the current project id from the redux store, so you don't need to thread it all the way down to `ReferenceLayerFlyout`

maurizi (changes requested):
Functionally this is looking really great & it matches the wireframes well. Great job 🎉 

I've left a few comments on things I think need to be updated before this is merged, primarily backend related but also one frontend issue re: how we update the layer colors on the map after saving them.

ktohalloran:
Ok, now that Jenkins built this again successfully, this is ready for re-review! Thanks for all the feedback and the assistance in making the edits along the way.

maurizi (approved):
Tested it out one last time and it's working great 👍

PR #1151, opened 2022-02-22
Update images

BryanQuigley:
## Overview

postgres minor bump 12.2 -> 12.4
node:16 now bumped to bullseye - this will likely happen with node:16-slim
images in the June timeframe anyway, but let's be explicit so no surprises

### Checklist

## Testing Instructions
Normal test cases should be fine.


maurizi (approved):
👍

PR #1152, opened 2022-02-22
Switch PlanScore integration from simple to multi-step API

### LINKED ISSUE #1131: Update PlanScore integration
When we implemented the PlanScore integration, we used what is now their "simple" API, which is now documented to have a size limit of 5MB.

We should update our code to instead use their multi-step API: https://github.com/PlanScore/PlanScore/blob/main/API.md#multistep-interaction

maurizi:
## Overview

The PlanScore backend has a file size limitation for direct upload, but allows any size file if you instead upload your file directly to S3 using a signed URL as part of a multi-step API.

Updating our app to make use of this more complex API made things quite a bit more complex, and also a bit more robust:
 - Our check for whether its safe to call the PlanScore API wasn't quite right (there can be unassigned areas, but every districts needs something assigned to it)
 - We weren't making use of their ability to poll for when the report is complete - previously we made the link available even when PlanScore wasn't done processing it.

### Checklist

### Demo

send-to-planscore

### Notes

I initially attempted to do this in a "stateless" fashion, hitting our backend for API requests to PlanScore but posting to S3 directly from the browser, which would have eliminated the need to do polling on the client.

Unfortunately I couldn't get this to work, due to both CORS issues and complications with using the callback approach for the S3 upload.

## Testing Instructions

- `scripts/update`
- If you don't already have it configured, contact @maurizi for the DistrictBuilder PlanScore token, then add it to your `.env` file
- On a map with some districts assigned, go to the Evaluate page, and then the competitiveness detail. You should see a "Send to PlanScore" button.
  - If any of your districts don't have any area assigned to them yet, it should be disabled
  - Once enabled, clicking it should move the button to a loading state
  - Loading will take... a while... but should eventually finish (if it doesn't, you should eventually see an error once it times out)
  - Once done loading, the "View on PlanScore" link should take you directly to the completed report
- If you refresh the page, the "View on PlanScore" link should still be present. If you make some changes and save, it should switch back to "Send to PlanScore"

Closes #1131


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -504,33 +510,98 @@ export class ProjectsController implements CrudController<Project> {
...
+      contentType: "application/json",
+      filename: `${project.name}.geojson`
+    });
+    const s3Response = await axios.post(s3Uri, form, {
+      headers: {
+        ...form.getHeaders(),
+        "Content-Length": `${form.getLengthSync()}`
+      },
+      // API docs say to expect 302, but in practice I've seen 303, checking for either to be safe
+      validateStatus: status => status === 303 || status === 303,
```
KlaasH, line 565:
```suggestion
      validateStatus: status => status === 302 || status === 303,
```

KlaasH (approved):
:+1: Looks good!  Submitting, viewing on PlanScore, not being able to submit when the map isn't done, and resetting when the map changes all worked as expected.
I saw one typo, noted below, which might not matter in practice (maybe they just always send a different status from what the docs say) but might as well be fixed.

Issue #1142, opened 2022-02-14
Change serialization to use topobuf


maurizi:
### Context

Work on the concurrency ADR showed that we can get significant improvements in file size by using a Protobuffer based file format - I used an outdated version of `geobuf` to do this - we'll be forking that in #1141 as `topobuf`

#### Notes

We changed from JSON -> the `v8` serialization format in https://github.com/PublicMapping/districtbuilder/pull/1099, which should be able to serve as a template for this PR

AC:
 - Update `process-geojson` and the backend server code to use `topobuf` instead of `v8.serialize` / `v8.deserialize`
 - Update `serialize-topojson` to be able to read `v8` files and write `topobuf` files
   - Use the updated `serialize-topojson` script to make new `.pbf` files for every region in:
     - Staging
     - Production
     - `scripts/load-dev-data`

maurizi:
Well this was an all-around disappointment.

Testing in staging showed using topobuf instead of v8 deserialization consumes more memory:
(left side is using v8, right is using pbf)

![image.png](https://images.zenhubusercontent.com/5cb8df898a3a7251ef0bdcff/3b81d249-b263-47b5-8173-09b93406070b)

This is probably something we could fix in the topobuf library at the expense of some CPU cycles... but we're slower than v8 deserialization already 😞 

topobuf:
```
districtbuilder-server-1  | cache-miss San Bernadino: 227.267ms
districtbuilder-server-1  | cache-miss Dane County: 634.601ms
districtbuilder-server-1  | cache-miss Delaware: 826.569ms
districtbuilder-server-1  | cache-miss New Jersey: 3.451s
districtbuilder-server-1  | cache-miss Wisconsin: 5.165s
districtbuilder-server-1  | cache-miss Pennsylvania: 14.393s
districtbuilder-server-1  | cache-miss Illinois: 10.048s
```

v8:
```
districtbuilder-server-1  | cache-miss San Bernadino: 137.446ms
districtbuilder-server-1  | cache-miss Dane County: 327.031ms
districtbuilder-server-1  | cache-miss Delaware: 597.973ms
districtbuilder-server-1  | cache-miss New Jersey: 3.232s
districtbuilder-server-1  | cache-miss Wisconsin: 5.475s
districtbuilder-server-1  | cache-miss Pennsylvania: 10.225s
districtbuilder-server-1  | cache-miss Illinois: 8.007s
```

File size is much much smaller with topobuf, but that is about the only thing it has going for us.
[...]

PR #1154, opened 2022-02-23
Prewarm cache on startup

maurizi:
## Overview

This is a small improvement I made in the course of working on #1142 - prewarming the cache on startup ensures that after a deploy we don't cause users to go from a hot to a cold cache for the map they're working on.

Also updates the `lru-cache` types to use the new definitions from DefinitelyTyped.

### Checklist

### Demo

N/A

## Testing Instructions

- `scripts/server`


ddohler (approved):
Works great -- my first edit after a cold start was much snappier than previously.

My one question is how much this will slow down the application's startup times. If it's a lot, it might be worth adding a warming list of just the states representing say, the top 80% of maps to strike a balance between startup time and responsiveness. But if our efforts to reduce startup times have paid off enough that it's not really a problem then we should just give people the best experience we can.

maurizi:
> My one question is how much this will slow down the application's startup times.

Looking at the Cloudwatch logs current application startup times in staging are 3 minutes (we _really_ improved load times with the recent changes).

The cache pre-warming strategy in https://github.com/PublicMapping/districtbuilder/pull/1154/commits/393d4ada7a72f2f76d1f4e2b037002589b648c37 doubled load times to 7 minutes.

I just pushed changes in https://github.com/PublicMapping/districtbuilder/pull/1154/commits/05082eb04aea690d95c78c1fbde65c1ec222c0ba, and with that the pre-warming now doesn't cause any noticeable increase in load times (still about 3 minutes), and made the code a bit cleaner as well.

ddohler (approved):
Wow, that's so much better! I think 3 minutes is totally acceptable--that's getting close to being on the order of how long it takes AWS to bring up a new machine and place an ECS task on it, so further improvements are likely to see diminishing returns.

I read through the new changes but didn't attempt to test since it seems like you've been exercising this pretty thoroughly in staging.

PR #1172, opened 2022-03-17
Update caching based on load test results

### LINKED ISSUE #1166: Investigate cache issues & load test 32Gb instances

maurizi:
## Overview

I did a variety of load tests on the application, using `r5.xlarge`, `r5.2xlarge`, and `r5a.2xlarge` instances.

This revealed a few problems:
 - Cache eviction wasn't working right, because the size calculation function is required to return an integer (fixed in https://github.com/PublicMapping/districtbuilder/pull/1172/commits/9b4ac7d2b22cd47769bb29f6151c3a8e8e5b3185)
 - The `vm.max_map_count` kernel parameter wasn't high enough for the new worker setup, which can load large states multiples times concurrently - I doubled it (https://github.com/PublicMapping/districtbuilder/pull/1172/commits/21b3b532863c5edd2c0bade9a3c6e7ab5614fe0e)
 - We don't have enough headroom right now to use `r5.xlarge` easily - I had to adjust the cache settings to the point where on startup when loading all 50 states into the cache, about 30 got evicted by the time it was done
   - This _might_ be solveable? see the next bullet point
 - We have a **memory leak**, occurring when adding states to the cache and/or when evicting states from the cache (it doesn't occur if I load test the same state repeatedly, only when states are added/removed from the worker cache)
   - This may be driven by objects created by native code (likely `Buffer`s) that are not on the normal JS heap, as the heap dumps I created showed a much smaller heap size than expected (in the order of megabytes instead of gigabytes).
     - If the above is the case, according to my research the only way to find it through tools are
        1. hope there is a related on-heap object also being leaked that we can find with standard Node heap profiling tools 🤞  or
        2. compile Node and any binary dependencies we have with debug symbols, and use Valgrind, a C/C++ leak detection tool 🔥 
 - It is possible for _only a worker thread_ to be killed by running out of heap space, but the `Pool` class doesn't detect this or offer an easy way to fix it 😢 I've attempted to recover from this scenario in https://github.com/PublicMapping/districtbuilder/pull/1172/commits/5c414281ee6a93442b4d0dc426680129e526e153 and https://github.com/PublicMapping/districtbuilder/pull/1172/commits/787c9f5a6777fa1fc04afa85423c6c97deec7fe1

As part of investigating / working on this I made a few other improvements:
 - Added _many_ more load test HAR files https://github.com/PublicMapping/districtbuilder/pull/1172/commits/c040f10f3b7ec77070ab567a35b5264f62d89b2a
 - Adjusted how we calculate the size of the worker cache https://github.com/PublicMapping/districtbuilder/pull/1172/commits/b651f10e5407d4a557b71aba129b071cf9b1be11
 - limited the number of workers a state can be sent to https://github.com/PublicMapping/districtbuilder/pull/1172/commits/d4359860373f629efca0e81d92cbdf6378f7ef47
   - this is intended to keep a surge of interest in the site for 1 particular state from hogging all site resources or spiking memory too much when loading TopoJSON
 - Switched from random allocation of states to workers to using round-robin allocation https://github.com/PublicMapping/districtbuilder/pull/1172/commits/aea8b49c4396836fee3dd4767a17b4844cf0df23, with the intent of making the application behavior more predictable (certainly it made my load tests more predictable).
   
Since I was unable to solve the memory leak issues, I added a memory check to our healthcheck https://github.com/PublicMapping/districtbuilder/pull/1172/commits/8a0109723be9e7cce73401b0018ed9a0e5ae28a8 - that should let us restart tasks somewhat gracefully _before_ it gets abruptly killed by the OoM killer

### Checklist

### Notes

Some other things I looked into:
 - Based on suggestions on a Node Github issue, I tried swapping out the default Node allocator for `jemalloc`, which is more proactive at returning memory to the operating system, but this didn't seem to make a substantive difference in our memory leak issues

## Testing Instructions

To test these changes and determine the right parameters to set, I've been deploying this branch to staging and then running a suite of load tests against it:
```
export JWT_AUTH_TOKEN="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJiMmIxZDIzMi04Y2RiLTRiOTYtODEyZi1mYjdhMmViZDZlMjEiLCJpZCI6ImIyYjFkMjMyLThjZGItNGI5Ni04MTJmLWZiN2EyZWJkNmUyMSIsIm5hbWUiOiJNaWtlIiwiZW1haWwiOiJtbWF1cml6aUBhemF2ZWEuY29tIiwiaXNFbWFpbFZlcmlmaWVkIjp0cnVlLCJoYXNTZWVuVG91ciI6dHJ1ZSwiaWF0IjoxNjQ3MzkwOTUwLCJleHAiOjE2NDc5OTU3NTB9.GL7PUT84Z9e4P7zCRv0usCuFeLoy1s-YfSYqRamuKx8"
HAR_FILE=patch-staging-texas-project.har PROJECT_UUIDS="5b2aaa92-a129-498a-b763-72b0d9319c51 d381a3c9-d198-4d7d-adc5-2dc4ee4ca02a 58aae758-669a-4a8b-8673-e16956875c63 1b7495fa-b963-4754-a2d7-6c5cc6666c80" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-ca.har PROJECT_UUIDS="d295864b-f608-42ce-b028-5ecb5c71c177 5b4ffa11-a556-44b4-b7a6-c3dbf7465ff0 c9dfd3fb-2141-43ba-a45a-69597305a1b0 a5c0f9f7-d3b4-4509-92b4-f9847b5c5af6" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-pa.har PROJECT_UUIDS="990c2314-e965-4874-b2ba-a56f8211beee aa753f04-93ab-49d1-a736-017e6a881f29 7b6ded96-8655-4400-95e6-4eed81872a47 7b6ded96-8655-4400-95e6-4eed81872a47" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-nc.har PROJECT_UUIDS="dd52d6ad-6255-4d19-9fcd-bbd528f91ef6 7214bfe2-5cec-46d4-a7a8-797e392a21fb e10798b6-df80-4820-a8be-7d85b8f10b38 e10798b6-df80-4820-a8be-7d85b8f10b38" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=patch-staging-nj-project.har PROJECT_UUIDS="1618b6f4-1038-4844-9b43-a0845cd44c41 26cece9c-f7b5-46ea-b208-b7ac1bc26004 84bb91b5-8141-42ce-bed8-3cbc9b1a6466 3e10cca9-4aa4-48f0-b55c-82b2448b18ac 66346135-6e12-4da3-942e-db89ab3fbb79" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-fl.har PROJECT_UUIDS="fb7b4a61-6730-4abf-9d09-6dc7066aaf94 eb553a05-6df1-4a31-bd3b-164ea3dd5b46 96622600-c35f-47cc-90c8-11afdb18489b a1e8333d-7641-4e72-94e7-0fc101e342f9" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-ny.har PROJECT_UUIDS="ce20d613-1e11-44cd-b93e-7b6ab41db5e4 283a028a-64e2-4c5a-864d-3bdf2da6b8f7 e0c94622-0b44-407e-91fc-5407561882e5 c4b5cca3-2630-4d8c-bb2f-ce3e59a280b7" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-il.har PROJECT_UUIDS="0c3d903f-398a-4dac-ac5c-5c5813df3996 fdfe8643-eb0b-4332-972a-04a89f3190db 0f4a9e6a-5a1d-4178-9094-62eed37f6117 e6de61e8-be5b-4ffc-8ea0-ee4ccc88e674" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-tn.har PROJECT_UUIDS="ae51eae0-bbf7-4d20-9322-29510b0ae415 5aa8e6f3-7449-4167-8bc0-93d97137f78c 17fa0f0f-3123-45d0-9e34-c8616559c101 d74a3b9a-520c-4a9e-8a68-35ce403b218f" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-va.har PROJECT_UUIDS="8323bf2f-429b-4b4d-a709-1ed2b35d18c5 5f054614-4d8b-46f7-b610-174e723c9ff1 c677d233-9d4e-4e47-89be-198f8f86d38d 454d2ad2-cbd0-48db-8675-ccd698d39207" docker-compose run --rm k6 \
&& sleep 10 && HAR_FILE=staging-mo.har PROJECT_UUIDS="7671bf94-6ff6-4251-a965-6f180fa581bf 2c0a68fa-0aeb-4afa-b982-1014f95cd161 69c5f30d-2565-490e-93a5-f90ca072d412 d720fd55-2c3d-4359-a4ce-3cba088dbf35" docker-compose run --rm k6
```

Closes #1166


src/server/src/healthcheck/healthcheck.controller.ts
```diff
@@ -23,9 +26,11 @@ export class HealthcheckController {
     const timeout = process.env.TYPEORM_HEALTH_CHECK_TIMEOUT
       ? Number(process.env.TYPEORM_HEALTH_CHECK_TIMEOUT)
       : undefined;
+    const maxRss = os.totalmem() * 0.9;
     return this.health.check([
-      async () => this.db.pingCheck("database", { timeout }),
-      () => this.topoLoaded.isHealthy("topology")
+      () => this.db.pingCheck("database", { timeout }),
+      () => this.topoLoaded.isHealthy("topology"),
+      () => this.memory.checkRSS("memory", maxRss)
```
ddohler, line 33: One potential danger of having a self-destruct trigger like this is that it could become a source of instability if we start running with less headroom for some reason and instances start hitting the self-destruct threshold soon after launch. On other projects we've tried to deal with similar issues by giving tasks a fixed (but variable) lifespan upon launch so that they self-destruct after a certain period of time. The approaches could potentially be combined, too, so that the memory check only becomes active after a time limit.

Another thing that might be helpful would be to have this notify Rollbar if it ever fails so that we would hopefully have some warning if we were creeping closer to the threshold (something to the effect of "Huh, we've had 7 tasks self-destruct for memory reasons in the past two days but it used to be super rare").

I don't think we necessarily need to do all or any of these right now, but if there's a cheap way to make this a bit more resistant to false positives I think it'd be good to take it.

P.S. Another possibility comes to mind: does clearing out a worker reclaim its leaked memory? If so, then perhaps we could assign lifespans at the worker level, which might allow us to cycle them individually without dropping requests or having the task as a whole be marked unhealthy.

ddohler (approved):
Looks good! I didn't run the load tests myself, but I did verify that everything still works well in local development. I had a couple comments for consideration but I think this will do the job.

src/server/src/healthcheck/healthcheck.controller.ts
maurizi: > P.S. Another possibility comes to mind: does clearing out a worker reclaim its leaked memory? If so, then perhaps we could assign lifespans at the worker level, which might allow us to cycle them individually without dropping requests or having the task as a whole be marked unhealthy.

From looking at the memory in-use metrics in AWS & correlating that w/ the couple of "worker died" rollbar errors we've seen, it certainly looks like a large chunk of memory gets cleaned up when the thread is killed.

That could potentially work around our memory leak issues pretty handily - I'm going to take a shot at something that makes use of that & see if it addresses the memory leaks.

maurizi:
@ddohler can you take another look at this? I refactored code based on your suggestion of clearing out workers to reclaim memory - I've removed `lru-cache`, and now the worker pool tracks region sizes and terminates workers when their cache would exceed the max size.

Based on my load tests & with some tweaks to the cache size, I was able to get all 50 states loaded into the cache on a 32Gb instance! Metrics show it topping out at about 85% of memory in use.

image

Even with it working on 1st load w/ all states cached, I'm a _little_ unsure if we'll still want to drop down to 32Gb instances.
The new setup where we terminate threads to reclaim memory _works_, but as a cache eviction strategy it's far from ideal, since - we no longer evict least-recently-used, its quite likely that during concurrent use someone will encounter a cache miss due to someone else triggering a worker termination.

On 64Gb instances we have 8 workers & lots of headroom, so this isn't a huge deal, but on 32Gb instances we're quite close to full on startup & have only 4 workers, so the _very first_ large state that encounters a cache miss will throw out a quarter of our cached data. But maybe in practice it wouldn't be a big deal? Especially since most times of heavy traffic are likely to be focused on a single region.

ddohler (approved):
🎉 Everything looks good! I did local testing with additional log output to verify expected functionality. I think a TSM on the load testing strategy here would be super interesting.

src/server/src/worker-pool.ts
```diff
@@ -1,77 +1,208 @@
...
-      : // If we have no available pools, use the most recent for this region
+      : // Next check if any workers are available, if we haven't hit our limit for this region
+      lastUsed.length < MAX_PER_REGION && allSettledQueues.length > 0
+      ? getBestFit(allSettledQueues)
+      : // If we have no available workers, use the most recent for this region
       lastUsed.length > 0
       ? lastUsed[0]
-      : // Lastly, just use anything
-        (_.sample(workerPools) as WorkerPool);
+      : // Lastly, just use anything, choosing workers round-robin
```
maurizi, line 107:
```suggestion
      : // Lastly, just use anything
```

src/server/src/worker.ts
```diff
@@ -454,11 +488,25 @@ async function getTopologyProperties(
   );
 }
 
+async function getTopologyMetadata(
+  regionConfig: IRegionConfig,
+  staticMetadata: IStaticMetadata
+): Promise<TopologyMetadata> {
+  // Computes the length of the districtsDefinition array
+  // For most regions, this is the number of counties in the state
```
maurizi, line 496:
```suggestion
```

PR #1180, opened 2022-03-24
update to postgres 13 and postgis 3.1

BryanQuigley:
## Overview

One of the downtime's today seemed to line up with an autovacuum.  Postgres 13 has improvements to autovacuuming (and other size improvements)

It also appears to "just work" - and should be a relatively painless upgrade.  Also switched to upstream postgis image.


BryanQuigley:
Closing for now

PR #1185, opened 2022-03-31
Fix archived layers crash

### LINKED ISSUE #1176: Organizations with unarchived templates should not be listed on the Import / Create project forms
After archiving the PA LRC's on project template, the organization still appears as an option on both the Import & Create project screens, however attempting to use the organization will show a spinner forever in the case of the Import or display an empty section for the Create page.

In both cases, we shouldn't be showing the organization at all if they have no templates that can be used.

maurizi:
## Overview

This fixes a multi-part bug:
 - We were incorrectly displaying inactive project templates / templates for archived regions on the Create & Import screens
 - If you tried to create a project for an archived region, we would not detect or validate that (we do for import)
 - Causing the server to try and load a region after application startup would instead cause it to crash, which was triggered by the previous bug, since trying to use the template would then cause it to try and load the archived region

### Checklist

### Notes

We had a `// @ts-ignore` above one of the offending lines I had to change in https://github.com/PublicMapping/districtbuilder/pull/1185/commits/64c378ce0887f03ec29b74dd02b2794cebfb8137, which had that line not been ignored would have revealed the type error. 😡 

## Testing Instructions

- Using `scripts/dbshell`, set one of your project template regions to be archived (from the `scripts/load-dev-data` regions I used PA).
  - On `develop`, if you attempt to create a project using that template, it will cause your server to crash
  - On this PR it should not be shown, and if you attempt to use the API to create a project for that template (to test, load the page with the region unarchived, then archive the region after the page loads) it should return an error and not crash the server
- Using `scripts/dbshell`, set the project template whose region you archived to be inactive
  - On `develop` you will still see an option to create a template using this project on both the Import & Create project screens. Attempting to create a project using the template from the Create project screen should cause your server to crash
  - On this PR, you should not see the template as an option
- Using `scripts/dbshell`, set _all_ of an organization's project templates to be inactive
  - On `develop` you should still see that organization as an option to create/import a project with
  - On this PR it should be filtered out
- After starting your server, use `scripts/dbshell` to add a new project region. Then try to create a project for this region.
   - On `develop` this should cause your server to crash
   - On this PR this should succeed

Closes #1176

### DIFF
--- a/src/client/screens/CreateProjectScreen.tsx
+++ b/src/client/screens/CreateProjectScreen.tsx
@@ -18,7 +18,14 @@ import {
 } from "theme-ui";
 
 import { DEFAULT_POPULATION_DEVIATION } from "../../shared/constants";
-import { CreateProjectData, IChamber, IProject, IRegionConfig } from "../../shared/entities";
+import {
+  CreateProjectData,
+  IChamber,
+  IOrganization,
+  IProject,
+  IRegionConfig,
+  OrganizationNest
+} from "../../shared/entities";
 
 import { regionConfigsFetch } from "../actions/regionConfig";
 import { userFetch } from "../actions/user";
@@ -48,6 +55,13 @@ const validate = (form: ProjectForm) =>
   form.regionConfig !== null
     ? ({ ...form, valid: true } as ValidForm)
     : ({ ...form, valid: false } as InvalidForm);
+
+function filterProjectTemplates<O extends OrganizationNest | IOrganization>(org: O): O | undefined {
+  const projectTemplates = org.projectTemplates.filter(
+    template => template.isActive && !template.regionConfig.archived
+  );
+  return projectTemplates.length > 0 ? { ...org, projectTemplates } : undefined;
+}
 
 interface ProjectForm {
   readonly name: string;
@@ -174,6 +188,9 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps) 
   });
   const { data } = createProjectResource;
   const [totalPopulation, setTotalPopulation] = useState<number | null>(null);
+  const organizationsWithTemplates =
+    "resource" in user &&
+    user.resource.organizations.filter(o => filterProjectTemplates(o) !== undefined);
 
   const onDistrictChanged = (e: React.ChangeEvent<HTMLInputElement>) => {
     const chamber =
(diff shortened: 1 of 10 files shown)


src/client/screens/CreateProjectScreen.tsx
```diff
@@ -174,6 +188,9 @@ const CreateProjectScreen = ({ regionConfigs, user, organization }: StateProps)
   });
   const { data } = createProjectResource;
   const [totalPopulation, setTotalPopulation] = useState<number | null>(null);
+  const organizationsWithTemplates =
+    "resource" in user &&
+    user.resource.organizations.filter(o => filterProjectTemplates(o) !== undefined);
```
ddohler, line 193: Something about the way this is structured was difficult for me to understand; I think it's because `filterProjectTemplates` is kind of acting as both a transformer and a filter at once. However, in a few minutes of thinking I wasn't able to come up with anything obviously better; the closest I got was to have `filterProjectTemplates()` return

```typescript
return org.projectTemplates.filter(
    template => template.isActive && !template.regionConfig.archived
  );
```

which I felt made the organization filter here clearer, but it didn't work as nicely as I was hoping down on line 301. Maybe you can think of something nicer, but otherwise this doesn't matter too much.

ddohler (approved):
Looks good! I tested and confirmed that trying to create a map with an archived region on `develop` resulted in 🔥 . The one thing that didn't quite go as expected was when I loaded the "Create Project" page on this branch, archived the PA region, and then tried to continue with creating a PA project. The testing instructions (and the diff, I thought) imply that doing this should have led to an error, but instead it chugged for a bit, loaded the region, and then created the project successfully (albeit in read-only mode because the ~project~ region was archived). I don't think anything about this really needs to change -- this seems like a really rare situation and even if someone does manage to stumble down this path it'll look a little strange but not wrong.

ktohalloran:
I haven't finished reviewing it yet (my server was doing something weird) but I understand this is a time sensitive issue so feel free to move forward without my review. That being said, I did want to ask, while the template does not exist on the "Create Project" page on this branch, it's still visible on the organization page and still has an active "Use this template" button. Clicking on it just results in a forever spinner with no immediate crashes, but should the template still be listed on the org page, too?

maurizi:
> That being said, I did want to ask, while the template does not exist on the "Create Project" page on this branch, it's still visible on the organization page and still has an active "Use this template" button. Clicking on it just results in a forever spinner with no immediate crashes, but should the template still be listed on the org page, too?

Was that with the region archived & the template still active? That setup doesn't really make much sense anyways, and we don't have any templates on production set up that way, so I'm not super worried about it.

PR #1186, opened 2022-03-31
[WIP] Add Lambda Function to Send DB Server Alarm to Slack

aaronxsu:
## Overview

More to come once all done.

Marked as WIP since:
1. need to find out why existing cloud watch alarms send all state (`OK`, `IN ALARM`, `INSUFFICIENT DATA`) changes to SNS global topic no matter from what state
2. need to find out a way to fake an alarm on staging to test that the lambda function sends notification to slack correctly

### Checklist

### Demo

### Notes

## Testing Instructions

- More to come once all done

Closes #XXX


deployment/terraform/lambda-functions/alert_db/src/alert_db.py
```diff
@@ -0,0 +1,60 @@
...
+def handler(event, context):
+    if SLACK_BOT_WEBHOOK is not None:
+        for record in event["Records"]:
+            sns = record["Sns"]
+            message = json.loads(sns["Message"])
+            alarm_name = message["AlarmName"]
+            new_state = message['NewStateValue']
+            reason = message['NewStateReason']
+
+            if alarm_name in DB_ALARM_NAMES:
```
aaronxsu, line 38: Only send alarms when they are DB server related. Add this check since it seems the global SNS topic listens to a bunch of other stuffs as well.

deployment/terraform/variables.tf
```diff
@@ -392,3 +392,12 @@ variable "aws_cloudfront_canonical_user_id" {
...
   type    = string
 }
+
+variable "aws_lambda_service_role_policy_arn" {
+  default = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+  type    = string
+}
+
+variable "slack_bot_webhook_url" {
+  type = string
```
aaronxsu, line 402: I already updated the staging tfvars for this webhook url. This is to remind me to update production tfvars once this is approved and merged.

deployment/terraform/lambda.tf
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  tags = {
+    Project     = var.project
+    Environment = var.environment
+  }
+}
+
+resource "aws_sns_topic_subscription" "alert_database_server_alarms_global_events" {
+  topic_arn = aws_sns_topic.global.arn
```
aaronxsu, line 32: This lambda function subscribes to the global SNS topic, which is used by the existing CloudWatch DB server alarms- but this global SNS topic is used by other alarms too. The lambda function itself will filter down to only DB server related alarms before sending Slack message.

This also means that the lambda function is triggered anytime there is a thing sent to this global SNS topic in this app. Not sure if this is what we wanted. I subscribed to this only bc the existing alarms send to it. I am happy to discuss and make changes if need be: say, to create another SNS topic, change existing CloudWatch alarm actions, and subscribe lambda to this new SNS topic, etc.

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
```diff
@@ -0,0 +1,60 @@
...
+                slack_message = {
+                    "text": f":alert::alert::alert: {alarm_name} state is "
+                    f" now {new_state}: {reason}\n"
+                    f"```\n{message}```"
+                }
```
maurizi, line 43: This looks great for when an alarm starts, but let's give a nicer/shorter message when the alarm goes away
```suggestion
                slack_message = {
                    "text": f":white_check_mark: {alarm_name} has recovered"
                    if new_state == "Ok"
                    else f":alert::alert::alert: {alarm_name} state is "
                    f" now {new_state}: {reason}\n"
                    f"```\n{message}```" if new_state
                }
```

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
```diff
@@ -0,0 +1,60 @@
...
+                post_to_slack_resp = requests.post(
+                    SLACK_BOT_WEBHOOK,
+                    json.dumps(slack_message).encode('utf-8')
+                )
+                try:
+                    post_to_slack_resp.raise_for_status()
+                except Exception as e:
+                    print("Response content:\n{}".format(
+                        post_to_slack_resp.content))
+                    raise e
```
maurizi, line 60: I'm curious why you didn't choose to use Node.js for this lambda function, given the rest of the application is a Node project?

This is already written up and working, and given how disconnected this is from the rest of the app I think it's fine to leave it in Python, but I'd prefer to not introduce new languages to the project without a reason to do so - in the future .

maurizi (approved):
I didn't test this myself, but it seems to be working based on the Slack messages posted a few days ago.

BryanQuigley (approved):
LGTM.  I still think it does make sense to just expose all alerts that we have defined, not just DB.

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
maurizi: What are the other alarms we have set up? It might make sense to enable more than just the database ones, though I don't want to be too noisy such that we stop paying attention to them (which they might be, since we've been ignoring them so far).

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
aaronxsu: This is a good question. I wrote a NodeJS lambda function before here: https://github.com/azavea/raster-foundry-deployment/tree/develop/core/terraform/lambda-functions/send-groundwork-metrics-to-slack - this is definitely something feasible. 

The issue is about the NodeJS version not being pinned in Alpine Linux even though the Alpine Linux version itself is pinned, which is the base image being used for our Terraform image: https://github.com/PublicMapping/districtbuilder/blob/develop/docker-compose.ci.yml#L26
```
cat /etc/os-release
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.12.0
PRETTY_NAME="Alpine Linux v3.12"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

For example, the last time I developed the above mentioned lambda function using NodeJS, the NodeJS version used in that Alpine Linux was `10.x`, but now Alpine Linux upgrades the default NodeJS package to `12.x`- even though the Alpine Linux version itself did not change. This is something hard to keep track of. Although for small lambda function in NodeJS script like this, it may not make much difference, and we could just use the corresponding `runtime` version on the Terraform side. I chose Python this time just for the sake of reducing the potential maintenance overhead.

As you mentioned, this is on the infra side of the code, and does not have much to do with application code. So I think this should be just fine.

In the future when we migrate away from using Alpine Linux as the base Terraform image, we shall consider using NodeJS+TS, which is tbh something I'd rather write in for this.

What do you think?

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
aaronxsu: I am not sure about the other alarms we set up before as they don't seem to be managed by Terraform. Note: this PR does not set up any alarm actions, as they are already defined by previous/existing infra settings, and they all dispatch actions to the global SNS topic.

If having more fine-grained SNS action dispatching from cloudwatch is of our interest instead of dispatching everything to the global SNS topic, I think we should open up an issue for that.

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
aaronxsu: Agreed. I updated this based on your suggestions here: https://github.com/PublicMapping/districtbuilder/pull/1186/commits/a98aa05161e410214bb01fcb921bb17a1925afcf

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
maurizi: That makes sense as a reason to stick w/ Python for now.

aaronxsu: Sounds good. Another approach I'd say is to move the build and zip of a lambda function outside of `./scripts/infra` so that we are not subject to this limitation and thus we could switch to use NodeJS. We may also want to update the Jenkinsfile as well for CI/CD, which should also be feasible I think. I'd be happy to switch, since I think we still have a chance to do that at this point!

BryanQuigley: if we reuse this for other projects we are going to want it in python..

aaronxsu:
@BryanQuigley I hear you, and I do personally think that exposing all alerts we have defined sounds like a good idea. I think we may open up an issue to plan for an approach for this, since, I also agree with what @maurizi brought up in a comment above that it could potentially send alerts with an amount more than we anticipate that would potentially be buried and ignored in the channel. 

Please let me know what both of you think!

BryanQuigley:
My vote is to try it with all alerts and then open cards to quiet them if there is too much.  You've taken the most look at the alerts, so if it seems like we should quiet them first - then let's do that.

deployment/terraform/lambda-functions/alert_db/src/alert_db.py
aaronxsu: I investigated a bit more and updated some notes in here: https://github.com/PublicMapping/districtbuilder/pull/1186#issuecomment-1090598613 - please lemme know what you think!

maurizi:
I took a look at the existing alarms myself:
image

The database related alarms look valuable enough to me to enable.

The two "CPU reservation" alarms are actually mechanisms that control our ECS deployments - when we're exceeding the desired number of ECS containers as part of a deploy the alarm activates, and its action is to reduce the # of containers by one every 30 minutes. So I don't think we need to send slack messages for those.

PR #1187, opened 2022-04-01
Fix worker routing based on size

maurizi:
## Overview

After monitoring the application logs after the deploy yesterday, I noticed unexpected behavior with how workers were being chosen - specifically we didn't seem to be avoiding picking workers that were too full for the region we're loading, leading to more cache evictions than expected, and caused cache utilization to mostly trend downwards over time:

These are the logs from production for last night, filtered to just show worker thread recreation. Notice the `rss` values are trending downward over time, and while they do go up on occasion they never go back above 50% of total system memory.
image
image
image

This caused me to take another look at `getBestFit`, whereupon I noticed we were incorrectly picking the largest worker *even if* it wouldn't fit, leading to much lower cache utilization over time.

Fixing this bug led to better utilization of the cache (and OoM errors), so I did some more load testing, after which I reduced the max cache size again, as well as increased the max `rss` before we fail the healthcheck 90% -> 95%.

New load tests show much better cache utilization, with `rss` values sometimes dipping down to as low as 17Gb when multiple workers were recreated close together, but mostly staying above 20Gb:
image

### Checklist

### Demo

### Notes

## Testing Instructions

- How to test this PR
- Prefer bulleted description

Closes #XXX


ddohler (approved):
Good catch, this makes sense and looks good (I didn't try to test)!

Issue #1189, opened 2022-04-05
Jenkins e2e tests sometimes hang indefinitely


maurizi:
Tests on Jenkins are _sometimes_ hanging_ indefinitely.

Test output shows the following:
```
$ jest --config ./test/jest-e2e.json --watchAll=false
[Nest] 29  - 04/05/2022, 4:07:48 PM   DEBUG [worker-pool] Object:
{
  "dockerMemLimit": "12 GB",
  "hostmem": "23.52 GB",
  "maxCacheSize": "115.2 MB"
}

FAIL test/districts.controller.e2e-spec.ts (181.575 s)
  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

      at Timeout._onTimeout (../node_modules/threads/dist/master/spawn.js:35:53)

  ● DistrictsController › import plan from block equivalency CSV › should return a districts definition

    Timeout: Did not receive an init message from worker after 10000ms. Make sure the worker calls expose().

[...]

PR #1195, opened 2022-04-07
Lower cache size and improve strategy for choosing workers

maurizi:
## Overview

Since releasing `1.16.1` our production tasks have become much better at utilizing worker cache - _too_ good in fact, as we've seen several tasks run out of memory and get killed.

I determined a cache level that should always be safe from workers running out of memory and is able to hold all 50 states (13Gb on a 32Gb instance), however while this was only a few GB lower than the previous amount I found it was no longer able to reliably start the application without at least one worker running out of memory and needing to be recreated.

After some research (see https://en.wikipedia.org/wiki/Bin_packing_problem#Offline_algorithms) I determined the best way to ensure full cache utilization on startup would require knowing the region size ahead of time, so I added a new `layer_size_in_bytes` field to `RegionConfig` that stores the size instead of calculating it on the fly, and added a command to set it that we can run for production (I already ran it on staging when testing this branch).

### Checklist

### Demo

Cache size logged:
image

After loading all 50 states, loading a 2nd CA region was enough to tip us over the edge and cause the 1st worker to be recreated - but note that for 2 / 4 workers we have _exactly_ `3.12Gb` of space being used - the maximum.
image

### Notes

I'm hopeful the changes in https://github.com/PublicMapping/districtbuilder/commit/d50c88485b4c5955b5e1fa267bf7ec32024866c6 will also address #1189 but I'm not sure

## Testing Instructions

- `scripts/update`
- You should be able to start your server locally

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added submit map button [#1179](https://github.com/PublicMapping/districtbuilder/pull/1179)
 
 ### Changed
+- Improved caching setup more & lowered cache size [#1195](https://github.com/PublicMapping/districtbuilder/pull/1195)
 
 ### Fixed
 
--- a/src/manage/src/commands/publish-region.ts
+++ b/src/manage/src/commands/publish-region.ts
@@ -3,10 +3,15 @@ import { IArg } from "@oclif/parser/lib/args";
 import S3 from "aws-sdk/clients/s3";
 import cli from "cli-ux";
 import { createReadStream } from "fs";
+import { readFile } from "fs/promises";
 import { join } from "path";
 import readDir from "recursive-readdir";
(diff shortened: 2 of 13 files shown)


ddohler:
Heads up that I won't be able to get to this until I'm back on the 18th. Feel free to merge after @rachelekm 's review if you would like.

ddohler:
Actually, scratch that -- I'm reassigning to @BryanQuigley for review while I'm out.

rachelekm (approved):
Cool! This looks good to me and worked well in my dev environment. For the sake of testing workers I added a couple regions and set them all to have a `layerSizeInBytes` of 100 MB and it functioned as expected:
Screen Shot 2022-04-08 at 4 16 15 PM
Screen Shot 2022-04-08 at 4 16 00 PM

BryanQuigley (approved):
LGTM.

Didn't measure it precisely, but memory usage feels less jumpy.

PR #1202, opened 2022-04-13
Add compatibility with newer cgroups

### LINKED ISSUE #1196: cgroup memory check incorrect (and needs case for newer installs)
## Description
In staging even though memory limit is set to default (about 31 GB), cat /sys/fs/cgroup/memory/memory.limit_in_bytes
9223372036854771712 still shows 9223372036854771712 (~unlimited)
 
It should be using:/sys/fs/cgroup/memory/memory.max_usage_in_bytes
(which shows 31810248704)

And on newer installs /sys/fs/cgroup/memory.max is the right answer.

## Steps To Reproduce
1. ./scripts/setup on system running cgroup2 
2. Fails with /sys/fs/cgroup/memory/memory.limit_in_bytes not found.

BryanQuigley:
This checks for the existence of the cgroupv2 memory limit and if that
doesn't exist falls back to the cgroupv2 limit.  This also aligns us
with the limit that works on our current staging environment.

## Testing Instructions

It should fail to start if this doesn't work.

Closes #1196

### DIFF
--- a/src/server/src/worker-pool.ts
+++ b/src/server/src/worker-pool.ts
@@ -20,7 +20,7 @@ const TASK_TIMEOUT_MS = 90_000;
 // in use maxing out at around 80%
 const dockerMemLimit = existsSync("/sys/fs/cgroup/memory.max")
   ? Number(readFileSync("/sys/fs/cgroup/memory.max", { encoding: "ascii" }))
-  : Number(readFileSync("/sys/fs/cgroup/memory/memory.max_usage_in_bytes", { encoding: "ascii" }));
+  : Number(readFileSync("/sys/fs/cgroup/memory/memory.limit_in_bytes", { encoding: "ascii" }));
 
 const hostmem = os.totalmem();
 const totalmem = Math.min(hostmem, dockerMemLimit);
--- a/src/server/src/worker-pool.ts
+++ b/src/server/src/worker-pool.ts
@@ -1,5 +1,5 @@
 import { Logger } from "@nestjs/common";
-import { readFileSync } from "fs";
+import { readFileSync, existsSync } from "fs";
 import _ from "lodash";
 import os from "os";
 import Queue from "promise-queue";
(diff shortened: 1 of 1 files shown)


src/server/src/worker-pool.ts
```diff
@@ -18,9 +18,10 @@ const TASK_TIMEOUT_MS = 90_000;
 // Remaining amount is split amongst each worker for topology data
 // This strategy seems to work for any amount of host memory and targets total memory
 // in use maxing out at around 80%
-const dockerMemLimit = Number(
-  readFileSync("/sys/fs/cgroup/memory/memory.limit_in_bytes", { encoding: "ascii" })
-);
+const dockerMemLimit = existsSync("/sys/fs/cgroup/memory.max")
+  ? Number(readFileSync("/sys/fs/cgroup/memory.max", { encoding: "ascii" }))
+  : Number(readFileSync("/sys/fs/cgroup/memory/memory.max_usage_in_bytes", { encoding: "ascii" }));
```
maurizi, line 23: Based on the reading I've done, the `max_usage_in_bytes` will "show max memory usage recorded": https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt, which is not the same as what `limit_in_bytes` returns.

And indeed, the value returned by `/sys/fs/cgroup/memory/memory.max_usage_in_bytes` doesn't match the `mem_limit` set in our `docker-compose.yml`, but reverting back to `limit_in_bytes` does https://github.com/PublicMapping/districtbuilder/blob/57b66247386eeb677206d87b806882e0284ce5d2/docker-compose.yml#L47-L48

Tested locally:
```shell
$ docker-compose run --rm --no-deps --entrypoint /bin/bash server -c "cat /sys/fs/cgroup/memory/memory.max_usage_in_bytes"
9113600
$ docker-compose run --rm --no-deps --entrypoint /bin/bash server -c "cat /sys/fs/cgroup/memory/memory.limit_in_bytes"    
12884901888
```

```suggestion
  : Number(readFileSync("/sys/fs/cgroup/memory/memory.limit_in_bytes", { encoding: "ascii" }));
```

BryanQuigley: You are correct - max_usage_in_bytes is not the correct value.

memory.limit_in_bytes appears to be set to unlimited in the actual AWS run container though.. - regardless that doesn't really matter in staging as the totalmem is the full memory.

maurizi (approved):
Looks good to me, though I don't know how to test the cgroups 2 part of this.

Issue #1171, opened 2022-03-16
Support CVAP new data for all 50 states


maurizi:
- Brian ran into an issue running `process-geojson` that we need to fix
    > I updated districtbuilder, but for some reason it's now throwing an error with tippecanoe during process-geojson when it tries to produce vectortiles about requiring GLIBC 2.29, when I have GLIBC 2.31. I'm not sure how to begin troubleshooting that.
 - Need to QA on staging / release data to production

maurizi:
CVAP & VAP data is now available on production.

PR #1206, opened 2022-04-28
Updates to support CVAP data

maurizi:
## Overview

When trying to add CVAP and VAP data to all 50 states, it predictably increased file size.
To work around this, I removed the `demographics` and `voting` properties from the TopoJSON in #1197 

For reasons unknown to me, even after dropping these fields the memory used by the server seems to be much higher than for the previous TopoJSON data, to the point that the server would get killed by running out of memory before loading even half of all states.

To determine a fix, I did several things:
 
 Added a script to load full state data locally, to ensure the cache settings I had would work for a local development instance or a small server like we'd like to use for staging https://github.com/PublicMapping/districtbuilder/commit/03ca0e9bad00a960b9098bc66d9c7ba82781174d
 - This revealed some race conditions that were only present when loading many states from disk at once (something that doesn't happen in production) which I fixed with https://github.com/PublicMapping/districtbuilder/commit/336bece12806cd1b369c9fc37ca55b532bc49656 and https://github.com/PublicMapping/districtbuilder/commit/ba6721f8635ca8d1adde42bab3d96c4d14975b07
 - Having that many states loaded at once makes it really painful if you accidentally run the server without realizing it, so I changed some of our scripts to use the `--no-deps` flag for `docker-compose run` https://github.com/PublicMapping/districtbuilder/commit/f56dc35f0c9c60f32b317a8287e78b31d4d83ef6

I tried playing with the max cache size for a bit, but didn't get very far here - I would have had to reduce it down far enough that we could no longer load all 50 states on a 32Gb instance.

I decided to try using a different serialization format again, which had much better success, and very different performance characteristics from when I last looked at things (https://github.com/PublicMapping/districtbuilder/issues/1142#issuecomment-1049286088), presumably due to dropping a lot of data fields.

You can see [this Gist for the benchmarks I wrote](https://gist.github.com/maurizi/dc78d915c7aace64c33aa554934f541b) - the summary is that `v8` deserialization is no longer performing well for us, and we would be better served by either `JSON.stringify` or something custom like `topobuf`.
Since after dropping unneeded data fields we're now below the 500Mb limit on string length for all of the production TopoJSON files, I switched us back to using `JSON.stringify`. This is primarily for the sake of simplicity, though `topobuf` is only slightly slower and does have lower memory requirements and a larger maximum file size, so we could switch to that in the future to support a larger region like the custom PA for the PA LRC organization that exceeds the file size limitations of `JSON.stringify`.

While I was waiting for CI to provision repeatedly I also worked on a refactor to `worker-pool.ts` to turn it into a NestJS service to be more in line with code standards on the rest of the server: https://github.com/PublicMapping/districtbuilder/commit/f25de63212c1fb4e3029699445bab40abbe5aaec

### Checklist

### Demo

Staging is running with CVAP & VAP data right now 😃 

## Testing Instructions

- `scripts/update`
- `scripts/load-full-state-data`
- Your server should be able to start (and should load relatively quickly)
- Remove your cached TopoJSON data and restart your server `sudo rm -rf /var/cache/district-builder-server-topology/*` - it should still load, though slower as it will need to download the files again

Connects to #1171


ddohler (commented):
I was able to start up the server and "load" all 50 states locally (though obviously the workers ran out of memory several times and had to be restarted).

One thing I noticed is that it seems like workers may be getting killed due to OoM errors more often than I would expect -- I had an existing Illinois plan, and I tested exporting it to CSV, uploading the CSV, and then creating a map from it, in a sequence without any other steps in between. I saw OoM errors in the console at each of those steps, but I was expecting to only see OoM errors for the first one, because after that the Illinois data should have been loaded into a worker and it wouldn't need to get killed. Here is the log output I saw (limited to WorkerPoolService):

```
# Export to CSV
server_1    | [Nest] 40  - 05/03/2022, 1:24:49 PM     LOG [WorkerPoolService] Terminating worker 1 due to OoM, worker sizes: 0: 485.59 MB, 1: 485.59 MB, total rss: 2.04 GB
server_1    | [Nest] 40  - 05/03/2022, 1:24:49 PM   DEBUG [WorkerPoolService] Created worker 1

# Upload CSV
server_1    | [Nest] 40  - 05/03/2022, 1:24:49 PM     LOG [WorkerPoolService] Terminating worker 1 due to OoM, worker sizes: 0: 485.59 MB, 1: 485.59 MB, total rss: 2.04 GB
server_1    | [Nest] 40  - 05/03/2022, 1:24:49 PM   DEBUG [WorkerPoolService] Created worker 1
server_1    | [Nest] 40  - 05/03/2022, 1:24:53 PM     LOG [WorkerPoolService] Terminating worker 0 due to OoM, worker sizes: 0: 485.59 MB, 1: 485.59 MB, total rss: 2.1 GB
server_1    | [Nest] 40  - 05/03/2022, 1:24:54 PM   DEBUG [WorkerPoolService] Created worker 0
server_1    | [Nest] 40  - 05/03/2022, 1:24:57 PM   DEBUG [worker-1] Downloaded layer s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z/, size: 485.59 MB, rss: 2.49 GB
server_1    | [Nest] 40  - 05/03/2022, 1:25:02 PM   DEBUG [worker-0] Downloaded layer s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z/, size: 485.59 MB, rss: 3.23 GB
server_1    | [Nest] 40  - 05/03/2022, 1:25:07 PM     LOG [WorkerPoolService] Terminating worker 1 due to OoM, worker sizes: 0: 485.59 MB, 1: 485.59 MB, total rss: 3.28 GB
server_1    | [Nest] 40  - 05/03/2022, 1:25:08 PM   DEBUG [WorkerPoolService] Created worker 1
server_1    | [Nest] 40  - 05/03/2022, 1:25:16 PM   DEBUG [worker-1] Downloaded layer s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z/, size: 485.59 MB, rss: 3.32 GB

# Create new map
server_1    | [Nest] 40  - 05/03/2022, 1:26:57 PM     LOG [WorkerPoolService] Terminating worker 0 due to OoM, worker sizes: 0: 485.59 MB, 1: 485.59 MB, total rss: 2.03 GB
server_1    | [Nest] 40  - 05/03/2022, 1:26:57 PM   DEBUG [WorkerPoolService] Created worker 0
server_1    | [Nest] 40  - 05/03/2022, 1:27:05 PM   DEBUG [worker-0] Downloaded layer s3://global-districtbuilder-dev-us-east-1/regions/US/IL/2021-06-04T15:05:37.089Z/, size: 485.59 MB, rss: 2.45 GB
```

Nothing ever failed, which is great, but if there's an issue with the logic for detecting when to kill workers then that could cause them to thrash unnecessarily. It's also possible that this is a dev-only issue.

The other thing I noticed is that I wasn't able to find the cache folder inside the `server` container; I'm not sure if that's a cause for worry or not:

```
ddohler@lima:~/Development/projects/districtbuilder$ docker-compose exec server bash
root@f86f6d61facc:/home/node/app/server# ls /var/cache/    
apt  debconf  ldconfig
root@f86f6d61facc:/home/node/app/server# 
```

maurizi:
> Nothing ever failed, which is great, but if there's an issue with the logic for detecting when to kill workers then that could cause them to thrash unnecessarily. It's also possible that this is a dev-only issue.

This should be fixed by https://github.com/PublicMapping/districtbuilder/pull/1206/commits/e6b84c67dd43ab2e307826686b3746e11bd0dffd

> The other thing I noticed is that I wasn't able to find the cache folder inside the server container; I'm not sure if that's a cause for worry or not:
Inside the Docker container that folder is [located at `/opt/topology`](https://github.com/PublicMapping/districtbuilder/blob/0ae271612a86c399e7edcc434e0420d29e90176d/docker-compose.yml#L39)

ddohler (approved):
Works great! Everything's a lot snappier now after that fix.

PR #1208, opened 2022-05-05
Hide last page button on Community Maps screen

### LINKED ISSUE #1205: Investigate & fix production database CPU usage
We've periodically seen spikes in database CPU usage.

A cursory look at the slow query logs during one of the incidents showed this query, which is used for the ~last page of the Community maps screen:
```sql
SELECT "project"."id" AS "project_id", "project"."name" AS "project_name", "project"."number_of_districts" AS "project_number_of_districts", "project"."districts" AS "project_districts", "project"."created_dt" AS "project_created_dt", "project"."updated_dt" AS "project_updated_dt", "project"."submitted_dt" AS "project_submitted_dt", "regionConfig"."id" AS "regionConfig_id", "regionConfig"."name" AS "regionConfig_name", "regionConfig"."s3_uri" AS "regionConfig_s3_uri", "regionConfig"."archived" AS "regionConfig_archived", "user"."id" AS "user_id", "user"."name" AS "user_name", "chamber"."name" AS "chamber_name", "chamber"."id" AS "chamber_id", CASE
WHEN districts IS NULL THEN NULL
ELSE JSON_BUILD_OBJECT(
'type', 'FeatureCollection',
'features', ARRAY(
SELECT JSON_BUILD_OBJECT(
'type', 'Feature',
'properties', feature->'properties',
'geometry', ST_AsGeoJSON(ST_Simplify(ST_GeomFromGeoJSON(feature->'geometry'), 0.001))::json
)
FROM jsonb_array_elements(districts->'features') feature
)
)
END AS "project_districts" FROM "project" "project" INNER JOIN "region_config" "regionConfig" ON "regionConfig"."id"="project"."region_config_id" INNER JOIN "user" "user" ON "user"."id"="project"."user_id" LEFT JOIN "chamber" "chamber" ON "chamber"."id"="project"."chamber_id" WHERE "project"."visibility" = $1 AND "project"."archived" = FALSE AND jsonb_array_length(project.districts->'features'->0->'geometry'->'coordinates')::integer = 0 ORDER BY "project"."updated_dt" DESC LIMIT 8 OFFSET 31024;
```

We should:
 - Check the RDS logs to ensure this is the primary cause of the database CPU slowdowns
 - Devise a fix/workaround. Some potential options:
   1. Just make the query faster? maybe we should partition this table by date? Or use a more performant method of pagination than OFFSET / LIMIT?
   1. Maybe adding a similar index the existing one, but in reverse chronological order would make accessing the last several pages fast? Then it would still be slow to access the middle but that will take you forever to actual;ly do
   1. Similar to above, but easier - just don't show buttons for the last page. This might be necessary anyways if we want to stop using OFFSET / LIMIT

aaronxsu:
## Overview

This PR adds a prop to the `PaginationFooter` component so that we configure the community maps UI to not show the last page button.

### Checklist

### Notes

Note that the linked issue's original scope was changed in the first comment.

## Testing Instructions

- Spin up app servers and frontend
- Log in to the home screen. Make sure the pagination footer shows the last page button
- Navigate to the "Community maps" page. Make sure the pagination footer does not show the last page button

Closes #1205

### DIFF
--- a/src/client/components/PaginationFooter.tsx
+++ b/src/client/components/PaginationFooter.tsx
@@ -45,7 +45,10 @@ const PaginationFooter = ({
 
   // We show a set of pages around the current page, as well as links for first & last page
   const startingPage = Math.max(1, currentPage - MAX_PAGES_AROUND_CURRENT);
-  const endingPage = Math.min(totalPages, currentPage + MAX_PAGES_AROUND_CURRENT);
+  const endingPage = Math.min(
+    showLastPageButton ? totalPages : totalPages + 1,
+    currentPage + MAX_PAGES_AROUND_CURRENT
+  );
   const pageNumbers = range(startingPage, endingPage);
 
   const PageLink = ({ number }: { readonly number: number }) => (
--- a/src/client/components/PaginationFooter.tsx
+++ b/src/client/components/PaginationFooter.tsx
@@ -7,6 +7,7 @@ import { range } from "lodash";
 interface StateProps {
   readonly currentPage: number;
   readonly totalPages: number;
+  readonly showLastPageButton?: boolean;
   // eslint-disable-next-line
   readonly setPage: (number: number) => void;
 }
@@ -34,7 +35,12 @@ const style = {
 
 const MAX_PAGES_AROUND_CURRENT = 3;
 
-const PaginationFooter = ({ currentPage, totalPages, setPage }: StateProps) => {
+const PaginationFooter = ({
+  currentPage,
+  totalPages,
+  setPage,
+  showLastPageButton = true
+}: StateProps) => {
   // This component is to be used for server-side pagination with a list of elements rendered separately.
 
   // We show a set of pages around the current page, as well as links for first & last page
@@ -66,14 +72,15 @@ const PaginationFooter = ({ currentPage, totalPages, setPage }: StateProps) => {
           {pageNumbers.map(number => (
             <PageLink number={number} key={number} />
           ))}
-          {pageNumbers[pageNumbers.length - 1] === totalPages || (
-            <React.Fragment>
-              {pageNumbers[pageNumbers.length - 1] < totalPages - 1 && (
-                <span sx={{ px: 2, fontWeight: "800" }}>·</span>
-              )}
-              <PageLink number={totalPages} />
-            </React.Fragment>
-          )}
+          {showLastPageButton &&
+            (pageNumbers[pageNumbers.length - 1] === totalPages || (
+              <React.Fragment>
+                {pageNumbers[pageNumbers.length - 1] < totalPages - 1 && (
+                  <span sx={{ px: 2, fontWeight: "800" }}>·</span>
+                )}
+                <PageLink number={totalPages} />
+              </React.Fragment>
+            ))}
         </ul>
       )}
     </Box>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+ - Hid last page button on Community Maps screen [#1208](https://github.com/PublicMapping/districtbuilder/pull/1208)
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


src/client/components/PaginationFooter.tsx
```diff
@@ -66,14 +72,15 @@ const PaginationFooter = ({ currentPage, totalPages, setPage }: StateProps) => {
...
-          )}
+          {showLastPageButton &&
+            (pageNumbers[pageNumbers.length - 1] === totalPages || (
+              <React.Fragment>
+                {pageNumbers[pageNumbers.length - 1] < totalPages - 1 && (
+                  <span sx={{ px: 2, fontWeight: "800" }}>·</span>
+                )}
+                <PageLink number={totalPages} />
+              </React.Fragment>
+            ))}
```
maurizi, line 83: We don't want to _always and unconditionally_ hide the link to the last page - it should still be possible to click through to the last page by going through the intervening pages first. We just don't want to show the link when you're not close to the last page.

I think this change should take care of that:
```diff
diff --git a/src/client/components/PaginationFooter.tsx b/src/client/components/PaginationFooter.tsx
index 6fbb7450..abc23221 100644
--- a/src/client/components/PaginationFooter.tsx
+++ b/src/client/components/PaginationFooter.tsx
@@ -45,7 +45,10 @@ const PaginationFooter = ({
 
   // We show a set of pages around the current page, as well as links for first & last page
   const startingPage = Math.max(1, currentPage - MAX_PAGES_AROUND_CURRENT);
-  const endingPage = Math.min(totalPages, currentPage + MAX_PAGES_AROUND_CURRENT);
+  const endingPage = Math.min(
+    showLastPageButton ? totalPages : totalPages + 1,
+    currentPage + MAX_PAGES_AROUND_CURRENT
+  );
   const pageNumbers = range(startingPage, endingPage);
 
   const PageLink = ({ number }: { readonly number: number }) => (
```

BTW you can create a bunch of test projects to have a lot of pages on the Community Maps page on your local database by using the [`./scripts/manage create-random-projects` command](https://github.com/PublicMapping/districtbuilder/blob/develop/src/manage/README.md#manage-create-random-projects-number-region) - that was helpful for me when testing this feature.

maurizi (changes requested):

src/client/components/PaginationFooter.tsx
aaronxsu: Good point and agreed. It makes sense to show the last page button when the current page is close to the last one, and we don't show the last page if the current page is far away. I tried applying the suggested diff, but it did not work very well for me. I think the reason is that `endingPage` controls what is in the `pageNumbers`, which is the pages in between.

So I made the changes here: https://github.com/PublicMapping/districtbuilder/pull/1208/commits/55b5219d62a95984c33c4a76dd0f4e96d2928b31 : basically, if we want to conditionally hide the last page button, only when the `currentPage + 3` is no less than the `totalPages` do we show that button, i.e. the current page is approaching the last page. Hope this makes sense. What do you think?

maurizi: With these changes you're now always showing the "·" separator, even when the last page isn't being displayed:
image

That dot separator should only be visible when we're showing the first/last page with a _gap_, e.g.
1 · 3 4 5 6 7 · 8

> I tried applying the suggested diff, but it did not work very well for me. I think the reason is that endingPage controls what is in the pageNumbers, which is the pages in between.

Yes - getting the last page to appear in `pageNumbers` when `showLastPageButton = false` was my intention. We already have code to [not show the last page button when it's in `pageNumbers`](https://github.com/PublicMapping/districtbuilder/blob/8471abd79828c6bf7def780bcd41edaba0486d7b/src/client/components/PaginationFooter.tsx#L69) (which I think was effectively dead code until now?)

Can you explain what scenario you tested that the diff didn't work in? I made sure to test it myself before suggesting the change, and just tried it again in a variety of scenarios and didn't see any issues.

maurizi: @aaronxsu and I discussed this on a call - he thought he needed to change the lines above as well as apply the diff, it's working fine once we sorted that out.

maurizi (approved):

PR #1213, opened 2022-05-09
Reduce max cache and fix max workers / region check

### LINKED ISSUE #1211: Investigate / fix OoM issues
After deploying new CVAP data, there were multiple app servers that reached the maximum rss threshold of 95% w/o any individual worker reaching a maximum cache size.

We should run some load tests & adjust cache settings to prevent this.

maurizi:
## Overview

Reduces the worker max cache size, to prevent OoM issues currently being encountered on production.

Since we seemed to be operating fine as-is on other instances sizes I tried to adjust the way we determine cache size to leave the amounts for development & staging roughly the same while reducing the production server cache size.

### Checklist

### Demo

image

### Notes

The purpose of https://github.com/PublicMapping/districtbuilder/commit/6f0b96abeba39f39bf75d799db75c686202de0c2 was to improve our logging when terminating workers:
https://github.com/PublicMapping/districtbuilder/blob/6f0b96abeba39f39bf75d799db75c686202de0c2/src/server/src/districts/services/worker-pool.service.ts#L172-L176
Without this change the `workerSize` for that worker index will have already been reduced by the time we recreate the worker.

We were also being overly aggressive in loading new regions - only checking regions that were fully loaded and not pending regions when determining if we can add another region. This showed up in my load tests log output as the following (we should max out at 2 workers w/ PA data, but we create 3):
image
This is fixed by https://github.com/PublicMapping/districtbuilder/commit/dcce52eae6fc4a2ff24d511b8df5197cf337c228

## Testing Instructions

Ensure you can start your local server and create/edit maps.

I already load tested staging and the `rss` rose up to a high of 29.5Gb using code on the `develop` branch and maxed out at 27.75Gb on this branch.

Closes #1211

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+ - Reduce max cache size [#1213](https://github.com/PublicMapping/districtbuilder/pull/1213)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


ddohler (approved):
Looks good!

Issue #917, opened 2021-08-11
Upgrade to Terraform 1.0.4


rbreslow:
### General

- [x] Ensure image variants exist in the [docker-terraform](https://github.com/azavea/docker-terraform) build matrix. 

### Terraform 0.12.29 → 0.13.5

<details>

- [x] Run `./scripts/infra plan` and ensure there are no proposed changes.

- [x] Update the container image tag in `docker-compose.ci.yml`.

- [x] Apply the following namespace-related changes.

  **config.tf**

  ```hcl
  provider "aws" {
    region = var.aws_region
  }

  terraform {
    backend "s3" {
      region  = "us-east-1"
      encrypt = "true"
    }
  }
  ```

  **versions.tf**

  ```hcl
  terraform {
    required_version = ">= 1.0.4"
    required_providers {
      aws = {
        source  = "hashicorp/aws"
        version = "~> 3.53.0"
      }
    }
  }
  ```

- [x] Run the automatic upgrade command.

  ```console
  terraform 0.13upgrade
  ```

- [x] Run `./scripts/infra plan`.

- [x] Run `./scripts/infra apply`.

</details>

 ### Terraform 0.13.5 → 0.14.3

<details>

- [ ] Run `./scripts/infra plan` and ensure there are no proposed changes.

- [ ] Update the container image tag in `docker-compose.ci.yml`.

- [ ] Run `./scripts/infra plan`.

- [ ] Run `./scripts/infra apply`.

</details>

 ### Terraform 0.14.3 → 1.0.4

<details>

- [ ] Run `./scripts/infra plan` and ensure there are no proposed changes.

- [ ] Update the container image tag in `docker-compose.ci.yml`.

- [ ] Run `./scripts/infra plan`.

- [ ] Run `./scripts/infra apply`.

</details>

Also, consider setting secret input variables as [sensitive](https://www.terraform.io/docs/language/values/variables.html#suppressing-values-in-cli-output).

---

 **References**

- https://github.com/azavea/fieldscope/pull/932
- https://www.terraform.io/upgrade-guides/1-0.html

PR #1222, opened 2022-05-16
Upgrade terraform to 0.12-> 1.0.0

BryanQuigley:
Bump AWS provider version to 3.53.0 which is the highest we can
go without replacing some of the modules (terraform 1.1 apparently makes this
easier to do).

Also mark some variables sensitive so they don't show up in output which
would help if we move from Jenkins to Github actions.

Closes: #917

## Testing Instructions

- Confirm that everything is still working on both staging and prod
- View the nicely shorter infra plan changes view

### DIFF
--- a/deployment/terraform/config.tf
+++ b/deployment/terraform/config.tf
@@ -1,6 +1,5 @@
 provider "aws" {
   region  = var.aws_region
-  version = "~> 3.10.0"
 }
 
 terraform {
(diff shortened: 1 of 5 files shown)


maurizi (approved):
👍 None of the other terraform variables look like they need to be marked as secret, and all of the other changes seem pretty trivial.

PR #1228, opened 2022-06-07
Fix excess memory usage on projects query

### LINKED ISSUE #1225: Investigate database outage
The production database was shutdown by AWS due to excess memory usage.

We need to look into the cause, and determine what the appropriate fix should be.

rachelekm:
## Overview

The production database was shutdown by AWS due to excess memory usage, with the potential cause being frequent queries that result in large temporary files being created, as described in [the issue comment](https://github.com/PublicMapping/districtbuilder/issues/1225#issuecomment-1137657992). In order to replicate this, I created ~1500 projects of a little over 50 regions and found similar queries resulting in a large amount of temporary files using `pgbadger` logging. While the temporary files in my local instance were noticeably smaller than those related to the production outage, the query is unique to simplifying the geojson districts on read and of similar frequency. It's assumed that multiple calls from users with a lot of projects of more complex geometries would log temporary file sizes closer to those that were created before the outage. In order to fix this, we decided to denormalize the `project` table to include a `simplified_districts` column, which would be set with the simplified geojson on create rather than read. After implementing, there are no more temporary files created after querying for a user's paginated projects.

### Checklist

### Notes

- The `pgbadger` .html "before" and "after" files are included in this PR for viewing with the intent to take them back out before merge. I created the logs for `pgbadger` by selecting the last page of the user maps screen with 1,467 projects of a little over 50 regions. The query durations are different for the "before" and "after" files, since I tested "before" with a couple other actions to compare (creating a new project, duplicating, navigating site, etc) vs just selecting through to the last page where large temporary files were expected. Below are screenshots of the changes between temporary files creation:
Before implementing changes:
Screen Shot 2022-06-07 at 8 57 12 AM
Screen Shot 2022-06-07 at 8 57 02 AM

After implementing changes:
Screen Shot 2022-06-07 at 8 38 24 AM
Screen Shot 2022-06-07 at 8 38 36 AM

- Since we moved the action of simplifying the geojson from the database back into js, we expect more CPU time dedicated to simplification than before. To evaluate wether it made sense to move this into a worker process, we tested creating various regions, including block level districts, and found simplification didn't take longer than ~26ms for those scenarios. For now it seems worth skipping and reevaluating if it becomes an issue.

## Testing Instructions

- `scripts/update`
- `scripts/server`
- Confirm mini maps display as before
- Create a new project via template and import and confirm each functions as expected and mini map displays
- On updating districts, confirm mini map updates to display new districts
- Create a handful of new projects using `scripts/manage create-random-projects` and confirm new projects have `simplified_districts` field and mini maps display

Closes #1225

### DIFF
--- a/src/server/src/projects/controllers/projects.controller.ts
+++ b/src/server/src/projects/controllers/projects.controller.ts
@@ -75,36 +75,6 @@ import { ChambersService } from "../../chambers/services/chambers";
 import { Chamber } from "../../chambers/entities/chamber.entity";
 import { ReferenceLayer } from "../../reference-layers/entities/reference-layer.entity";
 
-import { simplify } from "simplify-geojson";
-import bbox from "@turf/bbox";
-import { BBox } from "@turf/helpers";
-
-// TODO #179: Move to shared/entities
-export function simplifyDistricts(
-  this: any,
-  districts: DistrictsGeoJSON | undefined
-): SimplifiedDistrictsGeoJSON | undefined {
-  function computeBBoxArea(districts: DistrictsGeoJSON): number {
-    const box: BBox = bbox(districts);
-    return (box[2] - box[0]) * (box[3] - box[1]);
-  }
-
-  const boxArea = districts && computeBBoxArea(districts);
-  // We only use the districts column for displaying a mini-map outside of the main Project Screen
-  // so we can simplify the geometries to save on size and improve performance
-  districts &&
-    districts.features.forEach(districtFeature => {
-      const tolerance = boxArea && boxArea > 1 ? 0.005 : 0.001;
-      try {
-        simplify(districtFeature, tolerance);
-      } catch (e) {
-        this.logger.debug(`Could not simplify district ${districtFeature.id}: ${e}`);
-      }
-    });
-
-  return districts;
-}
-
 function validateNumberOfMembers(
   dto: CreateProjectDto | UpdateProjectDto,
   numberOfDistricts: number
@@ -400,7 +370,10 @@ export class ProjectsController implements CrudController<Project> {
     readonly user: User;
     readonly chamber?: Chamber;
     readonly regionConfig: RegionConfig;
-  }): Promise<DistrictsGeoJSON> {
+  }): Promise<{
+    readonly districts: DistrictsGeoJSON;
+    readonly simplifiedDistricts: SimplifiedDistrictsGeoJSON;
+  }> {
     const geoCollection = await this.getGeoUnitTopology(regionConfig);
     const geojson = await geoCollection.merge({
       districtsDefinition,
@@ -439,12 +412,13 @@ export class ProjectsController implements CrudController<Project> {
       !project.districts ||
       project.regionConfigVersion.getTime() !== project.regionConfig.version.getTime()
     ) {
-      const districts = await this.getGeojson(project);
+      const { districts, simplifiedDistricts } = await this.getGeojson(project);
 
       // Note we don't wait for save to return, and we throw away it's result
       void this.service.save({
         ...project,
         districts,
+        simplifiedDistricts,
         regionConfigVersion: project.regionConfig.version
       });
 
@@ -720,21 +694,6 @@ export class ProjectsController implements CrudController<Project> {
       } as Errors<UpdateProjectDto>);
     }
 
-    // Update districts GeoJSON if the definition has changed, the version is out-of-date, or there is no cached value yet
-    const updatedDistrictsGeoJSON =
-      existingProject &&
-      dto.districtsDefinition &&
-      (!existingProject.districts ||
-        existingProject.regionConfigVersion !== existingProject.regionConfig.version ||
-        !_.isEqual(dto.districtsDefinition, existingProject.districtsDefinition))
-        ? await this.getGeojson({
-            ...existingProject,
-            districtsDefinition: dto.districtsDefinition
-          })
-        : undefined;
-
-    const boundSimplifyDistricts = simplifyDistricts.bind(this);
-
     const dataWithDefinitions =
       existingProject &&
       dto.districtsDefinition &&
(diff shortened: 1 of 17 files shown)


src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -75,6 +75,36 @@ import { ChambersService } from "../../chambers/services/chambers";
...
+      const tolerance = boxArea && boxArea > 1 ? 0.005 : 0.001;
+      try {
+        simplify(districtFeature, tolerance);
+      } catch (e) {
+        this.logger.debug(`Could not simplify district ${districtFeature.id}: ${e}`);
+      }
+    });
+
+  return districts;
+}
```
maurizi, line 106: As we discussed, you might want to consider moving the simplification into `merge` in `src/server/worker.ts`.
You could change the `merge` function so it returns `Promise<{ readonly districts: DistrictsGeoJSON, readonly simplifiedDistricts: SimplifiedDistrictsGeoJSON} | null>`, as well as `this.GetGeojson()`, letting you have code that looks like the following:
```ts
const dtoWithDistricts = { ...dto, ...this.getGeojson(project) };
```

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -75,6 +75,36 @@ import { ChambersService } from "../../chambers/services/chambers";
 import { Chamber } from "../../chambers/entities/chamber.entity";
 import { ReferenceLayer } from "../../reference-layers/entities/reference-layer.entity";
 
+import { simplify } from "simplify-geojson";
+import bbox from "@turf/bbox";
+import { BBox } from "@turf/helpers";
+
+// TODO #179: Move to shared/entities
```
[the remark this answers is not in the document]
rachelekm: Whoops, this comment is a leftover from when I was running into types troubles importing `simplifyDistricts` to `manage` and thought it might have been related to 179. I'll take this out.

src/server/src/projects/controllers/projects.controller.ts
```diff
@@ -75,6 +75,36 @@ import { ChambersService } from "../../chambers/services/chambers";
...
+export function simplifyDistricts(
+  this: any,
```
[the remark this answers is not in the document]
rachelekm: I like that better and so far can't see a reason not to pass in `logger` instead! I'll make that change.

src/server/src/worker.ts
```diff
@@ -222,6 +226,28 @@ function getHierarchyDefinition(staticMetadata: IStaticMetadata, topology: Topol
...
+    const tolerance = boxArea && boxArea > 1 ? 0.005 : 0.001;
+    try {
+      simplify(districtFeature, tolerance);
+    } catch (e) {
+      logger.debug(`Could not simplify district ${districtFeature.id}: ${e}`);
+    }
+  });
+
+  return simplifiedDistricts;
+}
```
maurizi, line 249: I had to make some further adjustments to get this to work correctly (I had to look at the library source code to figure out that we should probably use a different exported function https://github.com/maxogden/simplify-geojson/blob/master/index.js)

While I was updating this function I also added more context to the log message in case it occurs in production.

```diff
diff --git a/src/server/global.d.ts b/src/server/global.d.ts
index b76d21fa..ba6e2f74 100644
--- a/src/server/global.d.ts
+++ b/src/server/global.d.ts
@@ -18,5 +18,5 @@ declare module "geojson2shp" {
 declare module "simplify-geojson" {
   import { GeoJSON } from "geojson";
 
-  export function simplify(feature: GeoJSON, tolerance?: number): GeoJSON;
+  export default function simplify<G extends GeoJSON>(feature: G, tolerance?: number): G;
 }
diff --git a/src/server/src/worker.ts b/src/server/src/worker.ts
index e02da140..f92f8555 100644
--- a/src/server/src/worker.ts
+++ b/src/server/src/worker.ts
@@ -36,7 +36,7 @@ import { DistrictsGeoJSON, SimplifiedDistrictsGeoJSON } from "./projects/entitie
 import { RegionConfig } from "./region-configs/entities/region-config.entity";
 import { User } from "./users/entities/user.entity";
 
-import { simplify } from "simplify-geojson";
+import simplify from "simplify-geojson";
 import bbox from "@turf/bbox";
 import { BBox } from "@turf/helpers";
 
@@ -228,22 +228,30 @@ function getHierarchyDefinition(staticMetadata: IStaticMetadata, topology: Topol
 
 // Simplify the geometries to save on size and improve performance
 // when displaying districts outside of the main Project Screen (mini-maps)
-export function simplifyDistricts(districts: DistrictsGeoJSON): SimplifiedDistrictsGeoJSON {
+export function simplifyDistricts(
+  districts: DistrictsGeoJSON,
+  region: RegionConfig
+): SimplifiedDistrictsGeoJSON {
   function computeBBoxArea(districts: DistrictsGeoJSON): number {
     const box: BBox = bbox(districts);
     return (box[2] - box[0]) * (box[3] - box[1]);
   }
 
   const boxArea = districts && computeBBoxArea(districts);
-  const simplifiedDistricts = districts && { ...districts };
-  simplifiedDistricts.features.forEach(districtFeature => {
-    const tolerance = boxArea && boxArea > 1 ? 0.005 : 0.001;
-    try {
-      simplify(districtFeature, tolerance);
-    } catch (e) {
-      logger.debug(`Could not simplify district ${districtFeature.id}: ${e}`);
-    }
-  });
+  const tolerance = boxArea && boxArea > 1 ? 0.005 : 0.001;
+  const simplifiedDistricts = districts && {
+    ...districts,
+    features: districts.features.map(districtFeature => {
+      try {
+        return simplify(districtFeature, tolerance);
+      } catch (e) {
+        logger.debug(
+          `Could not simplify district ${districtFeature.id} for region ${region.id}: ${e}`
+        );
+      }
+      return districtFeature;
+    })
+  };
 
   return simplifiedDistricts;
 }
@@ -368,7 +376,7 @@ async function merge({
   };
   return {
     districts: districts,
-    simplifiedDistricts: simplifyDistricts(districts)
+    simplifiedDistricts: simplifyDistricts(districts, regionConfig)
   };
 }
 ```

maurizi (changes requested):
I tested the code w/ the new changes, but verified via the network tab of the browser dev tools that we were sending full-size districts instead of simplified ones still.

I posted a diff that I think should fix things.

src/server/src/worker.ts
rachelekm: Darn, thanks for catching this. I made some assumptions about the library so taking a look at the source code is really helpful. Just added your changes in!

maurizi (approved):
👍 looks good

Don't forget to delete those pgBadger files before merging

PR #1232, opened 2022-06-17
Fix create random projects

### LINKED ISSUE #1231: Adjust create-random-projects for full state data
Our `create-random-projects` management command works well when creating a small number of projects from a small amount of regions. However, after loading full state data the command exits early with a "No active topology for region" log (even for regions where there should be topology available) before creating any or only a couple projects. Even though it waits for all the promises to resolve to set `layers`, we won't have actually loaded all the regions yet before we start creating projects because not all the items in `layers` are Promises - some are still `null` if we haven't begun to download them yet.

maurizi:
## Overview

Fix the `create-random-projects` command, which was not working if you had a large number of states.
Since this is a less-utilized part of the application, I also added a small test suite to catch this bug as well as the other regression I've seen a few times with this command (failing to update it after adding a new required field to `Project`).

### Checklist

### Notes

This is my first time using `pg-mem` but it's pretty cool!
I thought about just using mocks here, but actually hitting the "database" means we can check for regressions in this code due to changes in the database schema adding a required field - not something we would get if we mocked out the database-level.

## Testing Instructions

- Run `scripts/manage create-random-projects`

Closes #1231

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fixed create-random-projects for large number of states [#1232](https://github.com/PublicMapping/districtbuilder/pull/1232)
 
 
 ## [1.17.3]
--- a/src/manage/package.json
+++ b/src/manage/package.json
@@ -36,13 +36,14 @@
     "topojson-server": "3.0.1",
     "topojson-simplify": "3.0.3",
     "tslib": "1.11.0",
-    "typeorm": "0.2.41"
+    "typeorm": "0.2.41",
+    "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@oclif/dev-cli": "1.22.2",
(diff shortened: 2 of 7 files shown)


rachelekm (commented):
Great job, these tests work great! I'm still bumping into one issue where if I run `scripts/load-full-state-data` before the `create-random-projects` scripts the process exits without a log message:
Screen Shot 2022-06-21 at 10 32 40 AM
I can't tell if the issue is that there's a layer with an unresolved promise that's causing a hang or that the number of layers to resolve just takes too long and the process exits? I've tried it a couple times and it always exits on a different region. Does it make sense to add something like `layersLoaded()` back in but with a dummy timeout just to keep something on the event loop when there's a larger number of regions?

UPDATE: Shortly after writing this comment I realized I still had my `server` container running while trying to run this manage command, which has caused early process exits for me in the past 🤦‍♀️. Stopped the container and was able to successfully run `scripts/manage create-random-projects` with full state data. This is good to go!

rachelekm (approved):

Issue #1229, opened 2022-06-07
Document API endpoints


dmcglone:
Currently, there is no public documentation of DistrictBuilder API endpoints. We should provide some basic instruction on the API endpoints in the form of a markdown file in the repo.

PR #1239, opened 2022-06-27
Add API spec for globalProjects

aaronxsu:
## Overview

This PR adds a `README.md` file under `doc/spec/` that includes API spec for `/api/globalProjects`

### Checklist

### Notes

~It seems this API endpoint is behind auth. I am wondering if there is a way to get a bearer token for authentication before calling the endpoint. If yes, whether it is an endpoint or a set of steps to get this token, I think maybe we should have docs for such a way so that the API spec consumer can follow, since I read from the issue description that this endpoint is designed for public consumption. If this is already documented somewhere, it would be great to guide me to it and have the link included in this doc.~

## Testing Instructions

- Make sure the API spec makes sense and the example response is sufficient.

Closes https://github.com/PublicMapping/districtbuilder/issues/1229

### DIFF
--- a/doc/spec/README.md
+++ b/doc/spec/README.md
@@ -46,7 +46,10 @@ Get a paginated list of global projects.
                             "voting": {
                                 "democrat16": 0,
                                 "republican16": 0,
-                                "other party16": 0
+                                "other party16": 0,
+                                "democrat20": 0,
+                                "republican20": 0,
+                                "other party20": 0
                             },
                             "contiguity": "",
                             "compactness": 0,
--- a/doc/spec/README.md
+++ b/doc/spec/README.md
@@ -8,7 +8,7 @@ Get a paginated list of global projects.
 
 **Method** : `GET`
 
-**Auth required** : YES
+**Auth required** : NO
 
 **Permissions required** : N/A
 
--- a/doc/spec/README.md
+++ b/doc/spec/README.md
@@ -1,0 +1,120 @@
+# API Spec
+
+## List global projects
+
+Get a paginated list of global projects.
+
+**URL** : `/api/globalProjects`
+
+**Method** : `GET`
+
+**Auth required** : YES
+
+**Permissions required** : N/A
+
+**Query Parameters** :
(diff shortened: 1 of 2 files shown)


maurizi:
> It seems this API endpoint is behind auth. 

No it's not? I can definitely access it on production w/o being authenticated: https://app.districtbuilder.org/api/globalProjects?page=1&limit=1&completed=true

aaronxsu:
@maurizi perfect. I crossed out the notes and added a commit.

doc/spec/README.md
```diff
@@ -0,0 +1,120 @@
...
+                        "geometry": {
+                            "type": "MultiPolygon",
+                            "coordinates": [...
+                            ]
+                        },
+                        "properties": {
+                            "voting": {
+                                "democrat16": 0,
+                                "republican16": 0,
+                                "other party16": 0
```
maurizi, line 49: For most states we'll have data for both 2016 & 2020 elections now, so I'd consider including that as well
```suggestion
                                "democrat16": 0,
                                "republican16": 0,
                                "other party16": 0,
                                "democrat20": 0,
                                "republican20": 0,
                                "other party20": 0
```

doc/spec/README.md
```diff
@@ -0,0 +1,120 @@
...
+- `limit`: Integer; Number of global projects per page; Default: `10`; Optional
+- `completed`: Boolean; Filter to return completed projects or not; Default: `false`; Optional
+- `region`: String; Filter to the projects matching specified region code; Default: `undefined`; Optional
+
+## Success Response
+
+**Code** : `200 OK`
+
+**Content examples**
+
```
maurizi, line 26:
```suggestion

**Note** the contents of the `voting` and `demographics` columns are dependent on the region the project is associated with. Not every region will have political data (some may have no data or only data for 2016) or the full range of demographic data available (though every region _will_ have `demographics.population` at the least).

```

maurizi (approved):
Looks good👍 

I suggested a few small additions, but they're fairly small & the overall structure looks good to me.

PR #1244, opened 2022-07-13
Fix deletion of maps w/ archived regions

### LINKED ISSUE #1241: Archived layer loaded after startup causes application crash
## Description
We had a production outage on 7/12/2022 and about 12:45pm.

Looking at the AWS CloudWatch logs, it appears that both application servers attempted to load an topology layer, causing their health checks to fail.
```
[Nest] 29  - 07/12/2022, 4:42:35 PM   ERROR [HealthCheckService] Health Check has failed! 
{
    "topology": {
        "status": "down",
        "total": 66,
        "complete": 65,
        "pending": [],
        "loading": [
            "s3://districtbuilder-production-data-us-east-1/regions/US/PA/2021-10-27T03:18:31.591Z/"
        ]
    }
}
```

Inspecting the production database shows that this `RegionConfig` is archived, so we shouldn't be attempting to load it on the backend, or expose UI controls on the frontend that would cause us to load it.

# AC
 - Determine how to reproduce the issue
 - Patch the backend to return an error instead of trying to load an archived topology
 - Update the frontend to disable/hide whatever controls are triggering this behavior

maurizi:
## Overview

Validation of updates to the `pinned_metric_fields` requires getting the StaticMetadata from the `TopologyService` for the region - but we don't load archived regions in the TopologyService, so when attempting to archive a region for a map w/ an archived region, we caused the `TopologyService` to attempt to load that region at runtime, which then caused the healthchecks to begin to fail.

Archived maps are _mostly_ read-only so this isn't a significant issue, but we still allow deleting of archived maps, which is done by updating the `archived` flag on the Project - which then triggered the above scenario.

Deferring loading from the TopologyService unless the `pinned_metric_fields` are present allows us to 

### Checklist

## Testing Instructions

- Create a map
- Using `scripts/dbshell`, set the `regionConfig` associated with the map to be `archived`
- On the homescreen, attempt to delete your map using the "..." menu option
  - On `develop` this will show a message in the console that it has loaded a topology layer
  - On this PR, it should not show that message

Closes #1241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 ### Fixed
+- Fix server crashed caused by archived maps w/ an archived region [#1244](https://github.com/PublicMapping/districtbuilder/pull/1244)
 
 ## [1.18.0]
 ### Added
--- a/src/server/package.json
+++ b/src/server/package.json
@@ -55,6 +55,7 @@
     "form-data": "4.0.0",
     "geojson2shp": "0.5.0",
     "handlebars": "4.7.7",
+    "jest-mock": "^28.1.3",
     "lodash": "4.17.21",
     "nestjs-typeorm-paginate": "3.1.3",
     "nodemailer": "6.7.2",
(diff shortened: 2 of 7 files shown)


aaronxsu (approved):
Good to go after the changelog entry.

Issue #1237, opened 2022-06-22
Switch from Jenkins to Github Actions


BryanQuigley:
This should remove both:
 * Jenkinsfile
 * Jenkinsfile.release

They could both be replaced by one Github CI file that has conditional releasing depending on branch.

Github actions allows us to not have to depend on physical hardware for this project.

BryanQuigley:
Afaict the issue is due to the number of CPUs expected (Github actions has 2 Cores)

Could reproduce similar failure locally by editing src/server/src/common/constants.ts
 - commenting out import os from "os";
 - changing NUM_WORKERS to 2 - export const NUM_WORKERS = 2;

PR #1245, opened 2022-07-14
Port from Jenkins to Github actions including cibuild and staging deploy

BryanQuigley:
- Use new style cipublish script that can just lookup the DB
   ECR repo so we don't need to have it as a secret.
 - Needed to pass AWS creds into infra docker container
   credentials not found error.
 - Use GITHUB_SHA if exists as it's Github's equivalent for GIT_COMMIT.for the tags. Pass
   into infra container.
 - Add Rollbar credentials (staging) to Github actions and change code to make it clearer when running

TODO: - Port Jenkins release/Production deploy to Github actions

Speed ups (found as part of debugging)
 - Speed up Tippecanoe building (use 2 threads) slightly.
 - Stop running duplicate containers when running tests.  Reduces memory usage and some CPU.

From maurizi
 - worker-pool.service.ts/Fix cache/worker calculation for hosts w/ less than 8GB of memory
   was trying to use negative 600Mb for cache
 - worker-pool.service.ts/Add more logging
 - NODE_ENV fixes

Related: #1237

### Demo

See last Github action run (was running with custom bit to let it build on my branch instead of test)

### Notes

TODO is to replace production builds as well.

## Testing Instructions

- Could change test to branch name in ci.yml and try tests yourself
- Alternatively could look at builds in Github actions previous results.


docker-compose.ci.yml
```diff
@@ -30,6 +30,9 @@ services:
     environment:
       - AWS_PROFILE=${AWS_PROFILE:-district-builder}
       - GIT_COMMIT=${GIT_COMMIT}
```
maurizi, line 32: Should we remove the `GIT_COMMIT` variable here?

maurizi (approved):
Looks good to me👍

PR #1255, opened 2022-07-22
Add Github Release workflow

### LINKED ISSUE #1237: Switch from Jenkins to Github Actions

BryanQuigley:
This was done in the same way to match other Azavea projects (not public)
and to match the way Jenkins works for the project thus far as well.

Entering a sha with a valid image in ECR is therefore required.

This adds a release workflow that can only plan a production change.

### Notes

Was considering having it be automated, but all projects I've come across require manual process for release.

## Testing Instructions

Will be tested as part of this plan:

1.     Review and commit it with it just doing a terraform plan
2.     Test the pipeline to make sure it does what we expect
3.     Do a trivial other PR to add terraform apply
4.     Run release using same hash as before to confirm it works
5.     Deploy bugfix to production

Closes #1237

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      git_commit:
+        description: 'Short Git commit hash to deploy to production'
+        required: true
+
+jobs:
+  release:
+    name: release
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_BUILDKIT: 1
+      GIT_COMMIT: ${{ github.event.inputs.git_commit }}
+      DB_SETTINGS_BUCKET: districtbuilder-production-config-us-east-1
+      DB_DEPLOYMENT_ENVIRONMENT: "production"
+      NODE_ENV: production
+      DB_ROLLBAR_ACCESS_TOKEN: ${{ secrets.DB_ROLLBAR_ACCESS_TOKEN }}
+    steps:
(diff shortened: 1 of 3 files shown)


maurizi:
Reassigning this to @aaronxsu since he's taking over the project maintenance w/ me moving teams

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: "us-east-1"
+
+      - run: |
+          docker-compose -f docker-compose.yml -f docker-compose.ci.yml run --rm terraform -c "
+            unset AWS_PROFILE
+            ./scripts/infra plan"
```
aaronxsu, line 33: This makes sense so far. I am wondering if the `./scripts/infra apply` step was left out on purpose so that this is run manually outside of github actions?

Another note related to Slack message sending, I saw from the Jenkins release pipeline that there is such a notification step so that it sends messages before and after deployment. I am wondering if Github actions has such a workflow we could use.

PR #1257, opened 2022-08-01
Update release process as part of Github actions switchover

BryanQuigley:
* Change name for rollbar from Jenkins to Github_Actions
 * Streamline release workflow, remove manual staging test - and make sure
   we check to see if staging built OK.
 * Remove redundant slack notifications.  Rollbar already provides one.
 * Remove Release variant of Jenkinsfile

### Checklist

These are minor changes related to #1245 and #1255

## Testing Instructions

- Confirm rollbar stops saying Jenkins 
- Confirm new release process makes sense


scripts/infra
```diff
@@ -65,7 +65,7 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
                     -F "access_token=${DB_ROLLBAR_ACCESS_TOKEN}" \
                     -F "environment=${DB_DEPLOYMENT_ENVIRONMENT}" \
                     -F "revision=${GIT_COMMIT}" \
-                    -F "local_username=jenkins"
+                    -F "local_username=GitHub_Actions"
```
aaronxsu, line 68: This is a good change, and I agree that we should just use this rollbar notification to replace the excessive Slack notification after deployment is done, as we saw in the last deployment.

I am wondering if this `local_username=GitHub_Actions` just replaces the `jenkins` string in the notification message, or if it is also a key for something else. Would it be possible to grab the curl command with the env vars filled in and we do a test?

aaronxsu:
Another question about Jenkins server- where does this app's Jenkins server run? Is there any infra change needed to take that down for this app?

scripts/infra
BryanQuigley: It just replaces the message.

"GitHub_Actions deployed revision 11111"

Just tested with curl, see full results in slack.

BryanQuigley:
> Another question about Jenkins server- where does this app's Jenkins server run? Is there any infra change needed to take that down for this app?

It was running in urbanappsci but I took the final bit down when we did the GH actions deploy.

scripts/infra
aaronxsu: Perfect.

aaronxsu (approved):
Looks good and works well. Good to go!

PR #1259, opened 2022-08-23
Enable SSM to ec2 container hosts

BryanQuigley:
This allows us to move away from using the bastion and
to connect to a host by just pressing Connect on the instance in AWS WebUI.

Also installed psql on the host so we can connect to database too.

It should be noted that these hosts can get recycled as part of deployments -
so that's the only process change that needs to happen.

### Checklist

### Notes

I didn't remember that some users will use PGAdmin - that it also [possible](https://manumagalhaes.medium.com/ssm-port-forwarding-connect-to-rds-in-a-private-subnet-with-no-ssh-bonus-with-pgadmin-eefc79d1e8bd) but I haven't tested that yet. 

I have confirmed that I can connect to the database using the psql command.

## Testing Instructions

- Try out how you usually use the bastion but connecting via SSM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+ - Enable SSM to EC2 Hosts [#1259](https://github.com/PublicMapping/districtbuilder/pull/1259)
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


aaronxsu (approved):
Demoed during a pairing. Changes make sense. As mentioned, this does not remove Bastion and the current pgAdmin connection should not be affected. Good to go.

PR #1262, opened 2022-08-30
Updates to data processing script docs

### LINKED ISSUE #1242: Add docs for working with the data processing scripts
Inviting @maurizi to add more details or to-dos here.

lukemckinstry:
## Overview

This PR adds documentation for how a person/group developing, forking or deploying DistrictBuilder in the future can create GeoJSONs that feed into the `process-geojson` script (the output of which can then be fed into the `publish-region` script). `process-region` is documented assuming you have input data in the correct format already - and the purpose of this PR is to document that format.

### Notes

Assuming the user understands GeoJSON, the key point these new docs emphasize is what fields to include as `properties` in the GeoJSON corresponding to each geographic unit. The properties do not need to be exactly as listed, because the `process-geojson` script allows the user to select a subset of fields and rename fields. 

## Testing Instructions

- Review updates in the [`Development Data`](https://github.com/PublicMapping/districtbuilder#development-data) section of the main repo docs ([`README.md`](https://github.com/PublicMapping/districtbuilder#districtbuilder)).

Closes #1242

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,26 @@ $ ./scripts/manage update-region data/output-pa s3://previous/location/of/the/pu
 
 Note: when doing this, you will need to restart your server to see the new data, since it's cached on startup
 
+#### Acquiring custom data to upload to DistrictBuilder
+To create a custom region, you will need a GeoJSON for a state/region at the `county`, `blockgroup`, or `block` level of detail. The file should include the following properties for each geographic unit:
+
+ * geoid (a unique identifier for each geographic unit, eg. census geoid/FIPS)
+ * state
+ * county_identifier
+ * county_name
+ * blockgroup_identifier
+ * block_identifier
+ * population
+ * white
+ * black
+ * asian
+ * hispanic
+ * other
+ 
+We recommend acquiring data from the US Census Bureau, which typically distributes geographic and demographic files separately. Files for a state/region at the desired level of geographic detail (eg. blockgroup, block) should be acquired and then joined on the unique identifier field (eg. the census geoid/FIPS) into a single GeoJSON.
+
+When the GeoJSON is ready, use `process-geojson` to process the data and `publish-geojson` to publish the region to S3. In `process-geojson` the `--demographic` field specifies the columns to upload as demographic data, and multiple groups of demographic data can be uploaded (e.g. CVAP, etc.). The `--levels` field specifies what level of geographic detail (eg. blockgroup, block) contained in the GeoJSON should be processed. 
+
 ### Project Organization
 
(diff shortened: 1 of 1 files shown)


aaronxsu (commented):
The docs added make sense to me. I will leave it to Mike to check with fuller context to see if the fields suffice for the input data.

README.md
```diff
@@ -152,6 +152,26 @@ $ ./scripts/manage update-region data/output-pa s3://previous/location/of/the/pu
...
+ * population
+ * white
+ * black
+ * asian
+ * hispanic
+ * other
+ 
+We recommend acquiring data from the US Census Bureau, which typically distributes geographic and demographic files separately. Files for a state/region at the desired level of geographic detail (eg. blockgroup, block) should be acquired and then joined on the unique identifier field (eg. the census geoid/FIPS) into a single GeoJSON.
+
+When the GeoJSON is ready, use `process-geojson` to process the data and `publish-geojson` to publish the region to S3. In `process-geojson` the `--demographic` field specifies the columns to upload as demographic data, and multiple groups of demographic data can be uploaded (e.g. CVAP, etc.). The `--levels` field specifies what level of geographic detail (eg. blockgroup, block) contained in the GeoJSON should be processed.
```
maurizi, line 173: Expanded on this a bit, but maybe the last section would work better w/ an example?

```suggestion
When the GeoJSON is ready, use `process-geojson` to process the data and `publish-geojson` to publish the region to S3. 

In `process-geojson` the `--demographic` field specifies the columns to upload as demographic data, and multiple groups of demographic data can be uploaded (e.g. CVAP, etc.).

The `--levels` field specifies properties in the GeoJSON containing identifiers for other levels of geographic details. For example `--levels block,blockgroup` should be used when the input GeoJSON is at the `block` level and you wish to generate a `blockgroup` layer in addition to the `block` layer by combining all Blocks w/ the same `blockgroup` ID. 
```

maurizi (changes requested):
Thanks for the documentation updates @lukemckinstry!

I left some feedback on where I'd like to see it expanded - mostly it centers around documenting use cases outside US state-level data (e.g. international data, municipal instead of state-level, custom geography).

lukemckinstry:
Added examples for how to format GeoJSON that a user creates or acquires to be fed to `process-geojson`. The examples include a US state using census boundaries, and a region using precinct boundaries. I looked into the reapportionment/redistribution processes & data for several other countries (UK, Canada, Australia) but was unable to find a relevant example of public redistricting processes like the USA.

maurizi (approved):
Just one small typo but this is looking great otherwise 👍

PR #1263, opened 2022-09-02
Switch PGAdmin setup guide to use aws ssm (Systems Manager)

BryanQuigley:
This just requires AWS credentials - no more keys/VPN.

This is also the last blocker from keeping us moving away from the bastions.

### Checklist

## Testing Instructions

- Confirm you can get PGAdmin connected using the instructions.

### DIFF
--- a/deployment/configure_pgadmin.md
+++ b/deployment/configure_pgadmin.md
@@ -3,43 +3,46 @@
 ### Requirements
 
 - [PG Admin](https://www.pgadmin.org/)
-- VPN Access
 - AWS credentials
 
-### Configuration
+The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production" or Staging with Production.
 
-The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production".
+
+### Connecting
+1. Find an ec2 instance id like i-0e2796c2ebbfbb048 (not the bastion) either via the console or a command like 
+
+```
+aws ec2 describe-instances --query Reservations[*].Instances[*].[InstanceId,Tags] --filters "Name=tag:Environment,Values=Staging"
+```
+
+2. Then enable port 5432 locally to be connect to the database on the same port.
+
+```
+aws ssm start-session     --target i-0e2796c2ebbfbb048     --document-name AWS-StartPortForwardingSessionToRemoteHost     --parameters '{"host":["database.service.districtbuilder.internal"],"portNumber":["5432"], "localPortNumber":["5432"]}'
+```
+
+3. Connect to PGAdmin with the host set as localhost.
+
+### Configure PGAdmin
 
 To begin, right click "Servers" &#8594; "Create" &#8594; "Server". In the dialog that appears, complete the following:
+
 
 **General Tab**
 
 _Name_<br>
 Give your new server a name. The name can be whatever you choose; however, remember that you need to be able to distinguish between production and staging.
 
+
 **Connection Tab**
 
-_Host Name_<br>
-Locate the Host Name in AWS: All Services &#8594; RDS &#8594; Databases (in the left sidebar) &#8594; districtbuilder-staging &#8594; Connectivity & Security &#8594; Copy link under Endpoint.
+_Host Name_ localhost
 
-_Username and Password_<br>
+_Username, Password and Database Name_<br>
 Return to the AWS management console &#8594; All Services &#8594; S3 &#8594; districtbuilder-staging-config-us-east-1 &#8594; Terraform &#8594; download `terraform.tfvars`. Open the `tfvars` file (it can be opened by a text editor) and find `rds_database_username` and `rds_database_password`.
-
-**SSH Tunnel Tab**
-
-Turn on "Use SSH Tunneling".
-
-_Tunnel Host_<br>
-Return to the AWS management console &#8594; All Services &#8594; EC2 &#8594; Instances (left sidebar) &#8594; select the Staging Bastion. Copy and paste the bastion's public IPv4 address as your tunnel host. Note: If you don't see environment name in your instance table, click on the gear icon to the right and add "environment" as a tag column.
-
-_Username_<br>
-Username is `ec2-user`.
-
-_Authentication_<br>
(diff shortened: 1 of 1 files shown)


BryanQuigley:
@dmcglone Would you mind reviewing this to see if these [updated PGAdmin instructions](https://github.com/PublicMapping/districtbuilder/blob/77b884d9b31a5c77b118d57bd128a360e97dff20/deployment/configure_pgadmin.md) would work for you instead of using the current bastion/VPN/ssh bit?   Thanks!

deployment/configure_pgadmin.md
```diff
@@ -3,43 +3,46 @@
...
+```
 
-_Name_<br>
-Give your new server a name. The name can be whatever you choose; however, remember that you need to be able to distinguish between production and staging.
+2. Then enable port 5432 locally to be connect to the database on the same port.
 
-**Connection Tab**
+```
+aws ssm start-session     --target i-0e2796c2ebbfbb048     --document-name AWS-StartPortForwardingSessionToRemoteHost     --parameters '{"host":["database.service.districtbuilder.internal"],"portNumber":["5432"], "localPortNumber":["5432"]}'
+```
```
ktohalloran, line 22: I would suggest reordering this a bit to have users configure PGAdmin but not save it, then run this command, and then click "Save" in PGAdmin; otherwise, this connection is likely to timeout before you can get PGAdmin configured. Similarly, it would be helpful to add a note here stating that there is a timeout on this connection so that users know to time opening PGAdmin with running the command.

deployment/configure_pgadmin.md
```diff
@@ -3,43 +3,46 @@
...
 - AWS credentials
 
-### Configuration
+The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production" or Staging with Production.
 
-The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production".
 
-To begin, right click "Servers" &#8594; "Create" &#8594; "Server". In the dialog that appears, complete the following:
+### Connecting
+1. Find an ec2 instance id like i-0e2796c2ebbfbb048 (not the bastion) either via the console or a command like
```
ktohalloran, line 12: Users need to either configure or export the right AWS profile before running either of these commands, so I would suggest adding that here.

ktohalloran (changes requested):
This is so much easier than the old setup! Just a few notes below and also, I feel that with this setup, it may be worth adding a note about disconnecting, just because there are so many possible ways to do so, e.g. just closing PGAdmin, right clicking on the server name and selecting "Disconnect", or just using Ctl+C in the terminal.

deployment/configure_pgadmin.md
```diff
@@ -3,43 +3,46 @@
...
 
-The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production".
 
-To begin, right click "Servers" &#8594; "Create" &#8594; "Server". In the dialog that appears, complete the following:
+### Connecting
+1. Find an ec2 instance id like i-0e2796c2ebbfbb048 (not the bastion) either via the console or a command like 
 
-**General Tab**
+```
+aws ec2 describe-instances --query Reservations[*].Instances[*].[InstanceId,Tags] --filters "Name=tag:Environment,Values=Staging"
```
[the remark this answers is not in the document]
BryanQuigley: Both worked for me, so going with your change as clearly more compatible with different shells.

deployment/configure_pgadmin.md
BryanQuigley: good point.  I was hoping to have the Connection instructions to be mostly standalone from the config ones.  Will revise.

deployment/configure_pgadmin.md
BryanQuigley: added AWS_PROFILE line.

deployment/configure_pgadmin.md
```diff
@@ -3,43 +3,56 @@
...
+- AWS CLI with [Session Manager Plugin](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-install-plugin.html) installed
 
-### Configuration
+The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production" or Staging with Production.
 
-The configuration file is written referring to the staging database; however, the same process applies to the production database. Just follow the same instructions, replacing the word "staging" with "production".
 
-To begin, right click "Servers" &#8594; "Create" &#8594; "Server". In the dialog that appears, complete the following:
+### Connecting
+1. export AWS_PROFILE=district-builder
```
ktohalloran, line 13:
```suggestion
1. `export AWS_PROFILE=district-builder`
```

deployment/configure_pgadmin.md
```diff
@@ -3,43 +3,56 @@
...
+2. Find an ec2 instance id like i-0e2796c2ebbfbb048 (not the bastion) either via the console or a command like 
 
-_Name_<br>
-Give your new server a name. The name can be whatever you choose; however, remember that you need to be able to distinguish between production and staging.
+```
+aws ec2 describe-instances --query "Reservations[*].Instances[*].[InstanceId,Tags]" --filters "Name=tag:Environment,Values=Staging"
+```
 
-**Connection Tab**
+3. Then enable port 5432 locally to be connected to the database on the same port. Either replace $INSTANCE_ID with the new ID or set the shell variable. Note: [Windows has slightly different syntax](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-sessions-start.html#sessions-remote-port-forwarding).
```
ktohalloran, line 21:
```suggestion
3. Then enable port 5432 locally to be connected to the database on the same port. Either replace `$INSTANCE_ID` with the new ID or set the shell variable. Note: [Windows has slightly different syntax](https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager-working-with-sessions-start.html#sessions-remote-port-forwarding).
```

ktohalloran (approved):
Looks good, thanks for making those changes! Just a few tiny changes below for differentiation between code and context for readability but then it's good to go!

deployment/configure_pgadmin.md
BryanQuigley: added, thanks

deployment/configure_pgadmin.md
BryanQuigley: added, thanks

PR #1264, opened 2022-09-20
[WIP] Support client-side error-logging with Rollbar

### LINKED ISSUE #788: Add client-side rollbar error-logging
We have rollbar error-reporting for unhandled server-side errors, but we don't currently report unhandled client-side errors.

aaronxsu:
## Overview

This PR:
- [X] adds `@react/rollbar and rollbar` as dependencies for the client app
- [X] adjusts related constants so that they can be shared between client and server
- [X] adds support of client-side error-logging and use router-base contexts
- [X] not in the code changes, but created two Rollbar access tokens for client side with `post_client_item` scope
- [ ] supports new env vars for the above two tokens

### Checklist

### Demo

### Notes

## Testing Instructions

- How to test this PR
- Prefer bulleted description

Closes #788


.github/workflows/ci.yml
```diff
@@ -17,17 +17,22 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: CI Build
-        run: ./scripts/cibuild
-        env:
-          NODE_ENV: test
+       # Set the bucket to pull the dotenv file
+      - run: echo "DB_SETTINGS_BUCKET=districtbuilder-staging-config-us-east-1" >> $GITHUB_ENV
```
aaronxsu, line 21: Env vars for CRA apps get read in the build process of the bundle. I uploaded a dotenv file to the above location, which contains the Rollbar access token for error logging for the client app. This token is used for both staging and production.

scripts/cibuild
```diff
@@ -23,18 +28,25 @@ if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
     if [ "${1:-}" = "--help" ]; then
         usage
     else
+        # Pull the dotenv file from S3 before building frontend
+        pull_env
```
aaronxsu, line 32: First step in CI is to pull the dotenv file. This is before the `update` script, because the dotenv file needs to be there before the client app is built as a bundle in the `update` process, which is required by CRA.

src/client/App.tsx
```diff
@@ -42,6 +45,25 @@ const PrivateRoute = ({ children, ...props }: { children: React.ReactNode } & Ro
   );
 };
 
+const environment = window.location.href.includes("staging") ? "staging" : "production";
```
aaronxsu, line 48: I wanted to use `ENVIRONMENT` constant, which is `process.env.NODE_ENV`. But apparently, if the app is built by the `yarn build` command, this variable is always `production`.

src/client/App.tsx
```diff
@@ -53,47 +75,97 @@ const pushReplaceHistory: PushReplaceHistory = {
...
-    <Router>
-      <QueryParamProvider history={pushReplaceHistory}>
-        <Switch>
-          <PrivateRoute path="/" exact={true}>
+const Routes = () => (
+  <Router>
+    <QueryParamProvider history={pushReplaceHistory}>
+      <Switch>
+        <PrivateRoute path="/" exact={true}>
+          <RollbarContext context="home">
```
aaronxsu, line 83: This component is supposed to provide us with better insights based on router in log messages: https://docs.rollbar.com/docs/react#using-with-react-router

src/server/src/common/constants.ts
```diff
@@ -10,4 +9,4 @@ export const DEFAULT_FROM_EMAIL =
 // One worker/CPU works well on AWS using memory-optimized units
 // For dev & CI that's not the case, so just use a fixed amount
 export const NUM_WORKERS =
-  ENVIRONMENT === "Development" || ENVIRONMENT === "test" ? 3 : os.cpus().length;
+  ENVIRONMENT === "development" || ENVIRONMENT === "test" ? 3 : os.cpus().length;
```
aaronxsu, line 12: I believe `NODE_ENV` gives `development` if running `yarn start` or `npm start`

BryanQuigley (changes requested):
Why switching environments to UpperCase?  Does the one app default to lowercase?

I would prefer that cibuild not need access to aws credentials - it is something an external contributor could run with a PR. For Github we could add the client rollbar to secrets like the server one (and for local dev we already have bootstrap). 

Could the commits get squashed a bit and instead of using "Review" for comments, add it to the commit message (or the code)?  Any of these review bits are not going to be there easily for someone reviewing in the future for why a change was made.

aaronxsu:
Hey @BryanQuigley these are good questions. Please see my thoughts below and feel free to lemme know what you think:

>Why switching environments to UpperCase? Does the one app default to lowercase?

I don't think this PR switches it to upper case. It only does that when checking if the env strings match, which is a common practice just in case the strings are the same but only different because of big or small letters.

>I would prefer that cibuild not need access to aws credentials - it is something an external contributor could run with a PR. For Github we could add the client rollbar to secrets like the server one (and for local dev we already have bootstrap).

I would prefer that too. The issue is that it would need to pull dotenv from S3. This dotenv is used in the `update` script to bundle frontend client app code as static asset. As discussed before, the frontend client app was created with "Create React App", which does not read env directly from the environment. It gives us two options: 1. supply a dotenv file 2. supply the env var right before `yarn build`. I went with option 1 in case we will need to manage more env vars in the future.

Regarding the bootstrap script, it pulls a dotenv for server, not the client app. So the dotenv is downloaded to where the server's `package.json` is, I think. The client app's `package.json` is at the root of the repo. Additionally, The server's dotenv is only needed in dev, but not in staging or production, since it is supplied directly from the env, which is what `Nest.js` allows. This is also why `bootstrap` script is not included in any CI-related scripts.

Please lemme know if these make sense or not. I am happy to have a quick call to iron out the rough edges.

>Could the commits get squashed a bit and instead of using "Review" for comments, add it to the commit message (or the code)? Any of these review bits are not going to be there easily for someone reviewing in the future for why a change was made.

Yes, this is on my plan and is usually something I do before merging the PR, in which I choose "Squash and merge". I usually put descriptions and more contexts in the commit message in this final step. The review comments are more for reviewers now. The commit messages in the final squash are for future us.

aaronxsu:
@BryanQuigley and I discussed in a call- scripts and CI related changes are update in this commit: https://github.com/PublicMapping/districtbuilder/pull/1264/commits/e59c24b8c7110d4cf0cb78ffe1d71228cdb34088

BryanQuigley (approved):
LGTM, tested and was able to send Yoda to Rollbar. Although it didn't get to slack because it was associated with the earlier error.

ktohalloran (approved):
Looks good to me 👍 I didn't know much about this, so thanks for asking for my review, I learned a lot from the conversation on this PR!

PR #1266, opened 2022-09-21
Upgrade to VPC that drops bastion and remove other related bits

BryanQuigley:
The bastion has been replaced by using AWS SSM and the use of PGadmin was confirmed to work in an earlier commit.

As part of upgrading to the new VPC we had to upgrade Azavea providers and cert versions so that AWS providers versions didn't conflict.

Note from deploying: had to manually remove the bastion security group from the inbound one

### Checklist

Not an app change... (change from SSM is already in Changelog)

## Testing Instructions

- Double check previous commits enabling SSM and PGAdmin access both work as expected.

### DIFF
--- a/deployment/terraform/network.tf
+++ b/deployment/terraform/network.tf
@@ -1,17 +1,13 @@
 module "vpc" {
-  source = "github.com/azavea/terraform-aws-vpc?ref=6.0.1"
+  source = "git::github.com/azavea/terraform-aws-vpc?ref=7.0.0"
 
   name                       = "vpc${var.project}${var.environment}"
   region                     = var.aws_region
-  key_name                   = var.aws_key_name
   cidr_block                 = var.vpc_cidr_block
   private_subnet_cidr_blocks = var.vpc_private_subnet_cidr_blocks
   public_subnet_cidr_blocks  = var.vpc_public_subnet_cidr_blocks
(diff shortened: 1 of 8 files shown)


deployment/terraform/network.tf
```diff
@@ -1,17 +1,13 @@
 module "vpc" {
-  source = "github.com/azavea/terraform-aws-vpc?ref=6.0.1"
+  source = "git::github.com/azavea/terraform-aws-vpc?ref=7.0.0"
```
aaronxsu, line 2: Wondering what this `git::` does. Could you expand a bit?

aaronxsu (commented):
Changes look good to me. 

Since current staging is overwritten by a recent PR merge, I wonder if we should schedule a call and see the best way to test changes in this PR using staging and see/document if we need any manual steps before the next production deployment.

deployment/terraform/network.tf
BryanQuigley: It didn't work with just bumping the version and I'm still not sure why.

In any case github.com/ is a shortcut do basically do git::github.com/ 

See: https://www.terraform.io/language/modules/sources#github

aaronxsu: Thanks. Makes sense to me now.

PR #1273, opened 2022-10-11
Specify verbosity in parameter group and set to terse.

### LINKED ISSUE #1265: Database Size Increasing faster than expected
## Description
On 9/21 The free space on the production database changed from 60GB free to 43 GB in the course of less then 12 hours. We investigated because of diskqueue warnings that seemed to happen every hour (4 diskqueue spikes at 10:52, 11:46, 12:46, and 13:47 ET).

Trying to determine the size increase in the database of "projects" only showed an increase in 650MB range for 500-700 new projects created.  We expect to be able to see the same disk usage with psql commands to what AWS RDS reports.

Also, the log size was very large (2 above 1 GB, and many in the hundreds of MBs) which is mostly geo data that isn't useful to log.  If that log data counts to our IO usage that might be reducing our performance.

We are looking to do some combination of:
- [ ] Limit how long paramaters that get logged will be via log_parameter_max_length and/or log_parameter_max_length_on_error
- [x] Reduce log verbosity - all logs show up as errors right now
- [ ] Reduce what gets logged - don't log the raw geographic data - not sure how

Logs do count to instance [free/used space](https://aws.amazon.com/premiumsupport/knowledge-center/rds-postgresql-query-logging/) - .  this would imply that they also impact IO usage.

Dropping (but worth considering in a future issue)
- [ ] Just increasing the size of the database again
- [ ] Reduce when we get alerted to DiskQueue
- [ ] Datatype changes (compression?)
- [ ] - [ ] A way to automatic cleanup (or somehow reduce the space) of old unused maps

BryanQuigley:
Verbosity can always be overridden in the live database.

Took two other actions:
 * Stopped copying logs to Cloudwatch - this has cost us a lot in the past (and is redundant to RDS logs)
 * Added AWS session key (how we are transitioning to do ops tasks).

### Checklist

- [ ] Description of PR is in an appropriate section of `CHANGELOG.md` and grouped with similar changes, if possible - these are small changes that may not need an entry.

### Notes

Tried plan on prod and it matches what has been working there.

## Testing Instructions

If no other PRs overwrite what's running in staging right now, both changed values (logs and verbosity) can be confirmed just looking at the DB config and parameter groups.

Closes #1265

### DIFF
--- a/deployment/terraform/database.tf
+++ b/deployment/terraform/database.tf
@@ -58,6 +58,11 @@ resource "aws_db_parameter_group" "default" {
     value = var.rds_log_autovacuum_min_duration
   }
 
+ parameter {
+    name  = "log_error_verbosity"
+    value = "terse"
+  }
+
   tags = {
     Name        = "dbpgDatabaseServer"
     Project     = var.project
@@ -93,6 +98,7 @@ module "database" {
   storage_encrypted          = var.rds_storage_encrypted
   subnet_group               = aws_db_subnet_group.default.name
   parameter_group            = aws_db_parameter_group.default.name
+  cloudwatch_logs_exports    = null
 
   alarm_cpu_threshold                = var.rds_cpu_threshold_percent
   alarm_disk_queue_threshold         = var.rds_disk_queue_threshold
--- a/docker-compose.ci.yml
+++ b/docker-compose.ci.yml
@@ -33,6 +33,7 @@ services:
       - GITHUB_SHA
       - AWS_ACCESS_KEY_ID
       - AWS_SECRET_ACCESS_KEY
+      - AWS_SESSION_TOKEN
       - DB_DEBUG=1
       - DB_SETTINGS_BUCKET=${DB_SETTINGS_BUCKET:-districtbuilder-staging-config-us-east-1}
       - DB_ROLLBAR_ACCESS_TOKEN
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Enabled SSM to EC2 Hosts [#1259](https://github.com/PublicMapping/districtbuilder/pull/1259)
 - Added support for client-side error-logging with Rollbar [#1264](https://github.com/PublicMapping/districtbuilder/pull/1264)
+- Specify verbosity in parameter group and set to terse. [#1273](https://github.com/PublicMapping/districtbuilder/pull/1273)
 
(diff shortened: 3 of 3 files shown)


deployment/terraform/database.tf
```diff
@@ -58,6 +58,11 @@ resource "aws_db_parameter_group" "default" {
     value = var.rds_log_autovacuum_min_duration
   }
 
+ parameter {
+    name  = "log_error_verbosity"
```
aaronxsu, line 62: A note here for future us, from [the official doc](https://www.postgresql.org/docs/13/runtime-config-logging.html#GUC-LOG-ERROR-VERBOSITY):

>Controls the amount of detail written in the server log for each message that is logged. Valid values are `TERSE`, `DEFAULT`, and `VERBOSE`, each adding more fields to displayed messages. `TERSE` excludes the logging of `DETAIL`, `HINT`, `QUERY`, and `CONTEXT` error information. `VERBOSE` output includes the `SQLSTATE` error code and the source code file name, function name, and line number that generated the error. Only superusers can change this setting.

I wonder the last bit here: `Only superusers can change this setting.`- do you anticipate if this could be an issue for Terraform to apply this change?

BryanQuigley: No issue - the postgres permissions do not apply to terraform/AWS changes.

docker-compose.ci.yml
```diff
@@ -33,6 +33,7 @@ services:
       - GITHUB_SHA
       - AWS_ACCESS_KEY_ID
       - AWS_SECRET_ACCESS_KEY
+      - AWS_SESSION_TOKEN
```
aaronxsu, line 36: Could you remind me the reason we are including this env var?

deployment/terraform/database.tf
aaronxsu: So when setting this param in Postgres programmatically, the performing user is a superuser on the Postgres side?

docker-compose.ci.yml
BryanQuigley: That's part of the AWS SSO bit - it will definitely be the default for ops going forward - and ideally everyone for accounts in our org (work to be done there..).

aaronxsu: Got it. Happy to volunteer to test it out when it is rolled out.

aaronxsu (approved):
Good to go after adding the changelog entry.

deployment/terraform/database.tf
BryanQuigley: I don't have visibility into how it's actually done - this is part of AWS RDS magic.

PR #1275, opened 2022-10-14
Upgrade to React 17

### LINKED ISSUE #1177: Update to React Version 17
We'll want to update `eslint` on the `client` to whichever version matches that expected by the newest version of CRA. We should update `eslint` for server & manage at the same time to match versions.

ddohler:
## Overview

Updates to React v17. There should be no functional changes.

### Checklist

### Demo

N/A

### Notes

1. I upgraded Typescript slightly, but I don't think it was necessary; I just bumped it to the latest version that I could without receiving any warnings. 

2. The current version of theme-ui isn't compatible with the new React JSX transform, so we needed to add DISABLE_NEW_JSX_TRANSFORM=true to the client container environment to disable it.

Upgrading to later versions of theme-ui resulted in numerous Typescript errors. We're going to have to upgrade theme-ui to the latest version in order to upgrade to React 18 anyway, and theme-ui 0.15.0 contains breaking changes that I think will incidentally fix the Typescript errors I was seeing (we're on 0.13.1 and the errors started showing up with version 0.14.0), so it didn't seem like the best use of time to look into fixing them at this point.

Anyway, we should include removing `DISABLE_NEW_JSX_TRANSFORM=true` from `docker-compose.yml` to the issue for upgrading to React 18 and include a note about breaking changes in theme-ui v0.15.0.

## Testing Instructions

- Run `./scripts/update` and then give the frontend as thorough of a test as you can -- it should function identically to staging and/or production.

Closes #1177


package.json
```diff
@@ -48,19 +48,19 @@
     "polylabel": "^1.1.0",
     "prop-types": "15.7.2",
     "query-string": "^7.0.1",
-    "react": "16.13.1",
+    "react": "17.0.2",
     "react-aria-menubutton": "^7.0.0",
     "react-aria-modal": "^4.0.0",
-    "react-avatar": "3.9.7",
+    "react-avatar": "3.10.0",
```
ddohler, line 54: There's a note in the release notes about adding React 17 compatibility, so I bumped to that version to be on the safe side. Luckily it was a small version bump.

aaronxsu (approved):
Code changes make sense and look good to me! I tested the app after the React version upgrade locally by following our UAT scripts and ran through it fine without any problem. This is good to good after linking the PR in the changelog.

Regarding the notes about theme-ui and the next upgrade of the React version 18, I think they are very helpful. Would you mind creating a React version upgrade card and copy over your findings to help future us?

Issue #1307, opened 2023-03-31
Investigate performance issue on production


aaronxsu:

KlaasH:
Here's the course of my investigation so far:

### The failures are caused by the database becoming unresponsive

I initially expected to see the old "ECS task has a memory spike and dies" pattern, but in fact there was no memory spike, and the ECS task logs are full of "Health check failed" messages that identify the database as the unhealthy component.  When that goes on long enough, the task gets killed off and another one gets started.  This brings down the app for at least 10 minutes, while the new task initializes, sometimes more.  I didn't look closely into what causes the "more" cases, but I think it might be that the new tasks initialize slower when the database is still in bad shape.  Or possibly we're getting cases where the new task never reaches "healthy" status and it gets cycled again.

### The disk was almost full

The free disk space was hovering around 6 GiB, out of 210 GiB total.  That doesn't seem like enough to provide Postgres with the scratch space it needs for things like large transactions and table maintenance.  It doesn't seem like we ever hit "disk full", but it seems possible that the database engine has been having to work harder due to the lack of space for writing temporary buffers.

I turned on storage auto-scaling, and it scaled up to 231 GiB shortly thereafter.  I set the max to 500 GiB because we don't want to just let it grow forever.  It seems like over the past two months (which is as far back as the storage logs go) we've added about 10 GiB per month, so it should be a while before we hit that limit, and by then it might be time to talk about options for reducing how much we store in the projects table (which accounts for almost all of the space in use).

We had [a conversation on Slack](https://element84.slack.com/archives/C04UJQ8VAV8/p1681239490680409) about brainstorming ways to reduce the size of the projects table, such as by moving projects out and archiving them to S3 after a certain amount of time.  But, as I said there, I think it's premature to put significant effort into saving storage space, since it's still relatively cheap (tens of dollars/month) relative to how much it would cost to design and implement any kind of system to do that. 

### We've been paying a lot for provisioned IOPS

[...]

PR #1312, opened 2023-06-20
Set max_allocated_storage var and switch to updated Terraform db module

KlaasH:
## Overview

In the course of dealing with [issue #1307](https://github.com/PublicMapping/districtbuilder/issues/1307), I switched the database to using storage auto-scaling (and also to the `gp3` storage type).  But [the Terraform module we're using](https://github.com/azavea/terraform-aws-postgresql-rds) for the database config hasn't been updated in a while and doesn't support auto-scaling.  So I wasn't able to update the Terraform config then to match the changes I'd made in the console, but I left it at that.  Turns out, having the deployed state differ from the Terraform config in a way that the config can't even represent is a bad plan, and it bit us, in the form of [an attempt to run the Release](https://github.com/PublicMapping/districtbuilder/actions/runs/5293760757) action crashing and leaving the infrastructure in a somewhat broken in-between state.

So we need to update the Terraform config for the database to match how it is in production, that that's what this does.  It switches to [a fork of the module](https://github.com/KlaasH/terraform-aws-postgresql-rds/tree/max-allocated-storage) that has the `max_allocated_storage` parameter added.

I [opened a PR](https://github.com/azavea/terraform-aws-postgresql-rds/pull/48) on the module repo, but I'm not sure what the maintenance situation is for those modules these days, so I think it makes sense to just use the branch on my fork for now.

### Checklist

### Demo

I made the same manual change to the staging database that I had made to production (changing to gp3 and turning on auto-scaling) then pushed a test branch to run [a test deploy to staging](https://github.com/PublicMapping/districtbuilder/actions/runs/5326258481/jobs/9648051707).  Per [the logs](https://github.com/PublicMapping/districtbuilder/actions/runs/5326258481/jobs/9648051707#step:6:194), it noticed that the database had been changed outside of Terraform, but didn't find any changes that needed to be applied to it.

## Testing Instructions

With this branch checked out, run the Terraform container:
```
DB_SETTINGS_BUCKET=districtbuilder-production-config-us-east-1 \
  DB_DEPLOYMENT_ENVIRONMENT=production \
  GIT_COMMIT=911fdcb \
```
(that's the currently-deployed git commit, to avoid requiring that the app services get updated)

Inside that container, run `./scripts/infra plan`.  It will show some changes—I think that's a combination of the partial apply from the failed deploy and possibly some drift on the AWS side—but it shouldn't show any changes required for `resource "aws_db_instance" "postgresql"` (whereas if you check out `develop` and run it again, it will).

### DIFF
--- a/deployment/terraform/database.tf
+++ b/deployment/terraform/database.tf
@@ -75,14 +75,17 @@ resource "aws_db_parameter_group" "default" {
 }
 
 module "database" {
-  source = "github.com/azavea/terraform-aws-postgresql-rds?ref=3.0.0"
+  # TODO (#1316): go back to github.com/azavea/terraform-aws-postgresql-rds once it's updated
+  # (see https://github.com/azavea/terraform-aws-postgresql-rds/pull/48)
+  source = "github.com/KlaasH/terraform-aws-postgresql-rds?ref=max-allocated-storage"
 
   vpc_id                     = module.vpc.id
   allocated_storage          = var.rds_allocated_storage
+  max_allocated_storage      = var.rds_max_allocated_storage
   engine_version             = var.rds_engine_version
   instance_type              = var.rds_instance_type
   storage_type               = var.rds_storage_type
-  iops                       = var.rds_storage_type == "io1" ? var.rds_iops : null
+  iops                       = var.rds_iops
   database_identifier        = var.rds_database_identifier
   database_name              = var.rds_database_name
   database_username          = var.rds_database_username
--- a/deployment/terraform/variables.tf
+++ b/deployment/terraform/variables.tf
@@ -58,6 +58,11 @@ variable "rds_allocated_storage" {
   type    = string
 }
 
+variable "rds_max_allocated_storage" {
+  default = 0
+  type    = string
+}
+
 variable "rds_engine_version" {
   default = "12.4"
   type    = string
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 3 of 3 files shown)


deployment/terraform/database.tf
```diff
@@ -75,14 +75,17 @@ resource "aws_db_parameter_group" "default" {
 }
 
 module "database" {
-  source = "github.com/azavea/terraform-aws-postgresql-rds?ref=3.0.0"
+  # TODO: go back to github.com/azavea/terraform-aws-postgresql-rds once it's updated
+  # (see https://github.com/azavea/terraform-aws-postgresql-rds/pull/48)
```
aaronxsu, line 79: Thanks for leaving a commend here. I'd suggest we also create an issue in this repo so that we use that one to update the `source` once the PR is merged in the upstream.

deployment/terraform/variables.tf
```diff
@@ -58,6 +58,11 @@ variable "rds_allocated_storage" {
   type    = string
 }
 
+variable "rds_max_allocated_storage" {
```
aaronxsu, line 61: I'd assume the tfvars in S3 is already updated- right?

aaronxsu (approved):
🤞🏼

deployment/terraform/variables.tf
KlaasH: I haven't updated it yet, but I will before I push the button to deploy to production.  I'm going to watch the staging build after I merge and double-check the production vars against the staging ones to make sure they're how we want them.

deployment/terraform/database.tf
KlaasH: Good point. I made https://github.com/PublicMapping/districtbuilder/issues/1316.

aaronxsu: Thanks!

deployment/terraform/variables.tf
aaronxsu: Sounds like a plan to me 👍🏼

PR #1315, opened 2023-06-22
Make TopologyService refuse to load archived RegionConfigs

### LINKED ISSUE #1314: Make TopologyService refuse to load archived RegionConfigs
The server is still doing this:
```
[Nest] 28  - 06/22/2023, 1:04:35 PM   ERROR [HealthCheckService] Health Check has failed! 
{
    "topology": {
        "status": "down",
        "total": 62,
        "complete": 61,
        "pending": [],
        "loading": [
            "s3://districtbuilder-production-data-us-east-1/regions/US/PA/2021-10-27T03:18:31.591Z/"
        ]
    }
}
```
That RegionConfig is known to be broken, and it shouldn't get loaded because it's marked archived.  But something obviously still has a reference to it and is pulling it in.  That can happen because `TopologyService.get()` [allows for adding a RegionConfig to the list](https://github.com/PublicMapping/districtbuilder/blob/911fdcbb0ecbc5b11921c00a9adba13eb02fd49a/src/server/src/districts/services/topology.service.ts#L80-L82).

I'm assuming that's right and useful in some cases, to reload ones that got evicted from the cache, but in this case it's bad.  Adding a broken layer to the list causes the workers to fail health checks and get constantly cycled.

There shouldn't be anything trying to load that RegionConfig, and in fact [we fixed an issue like this before](https://github.com/PublicMapping/districtbuilder/issues/1241), but something is still doing it.  It would be good to find out what, but in the meantime I think it makes sense to try to convert this from a "brings down the site" error by making `TopologyService.get()` refuse to add archived layers to the list.  As a bonus, having it log an error and return `undefined` instead might make it easier to figure out where the bad code is that's still trying to rely on live loading of archived configs.

KlaasH:
## Overview

As described in issue #1314, some endpoint is causing an archived RegionConfig to get added to the layers list, and since it happens to be a broken layer that crashes when you try to load it, it's causing lots of problems.

Ideally we'd know where that request is coming from and fix it (like we did in #1241), but having a check in `TopologyService.get()` will solve the bigger part of the issue, and will be a good thing to have as a failsafe even if we do find the source.

### Checklist

## Testing Instructions

Since I don't know where the bad call is coming from, the best I can come up with for how to test this would be to modify the code locally to trigger it.  E.g. manually archive a `region_config` instance, then hard-code its ID into one of the calls in `projects.controller.ts` (in either the `createOne` or `duplicateOne` methods), then do the corresponding operation and confirm that it produces the error message.

Closes #1314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Updated database Terraform module to use the 'max_allocated_storage' param [#1312](https://github.com/PublicMapping/districtbuilder/pull/1312)
+- Made TopologyService check 'archived' status and refuse to load archived layers [#1315](https://github.com/PublicMapping/districtbuilder/pull/1315)
 
 ### Fixed
 
--- a/src/server/src/districts/services/topology.service.ts
+++ b/src/server/src/districts/services/topology.service.ts
@@ -76,6 +76,10 @@ export class TopologyService {
     if (!this._layers) {
       return;
     }
+    if (regionConfig.archived) {
+      this.logger.error(`Tried to load archived RegionConfig ${regionConfig.id}`);
+      return undefined;
+    }
     if (!(regionConfig.s3URI in this._layers)) {
       // If we encounter a new layer (i.e. one added after the service has started),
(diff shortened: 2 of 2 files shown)


aaronxsu (approved):
🤞🏼

